PR #3, opened 2025-02-19
Remove projects directories from expert

scohen:
We're moving expert from the current setup of an umbrella + projects hybrid to a poncho project with clearer dependencies. This is the first step where we push a bunch of the projects into the current umbrella. 

The next stage will remove the projects from the umbrella and make this project into a poncho.


scohen:
I did each move in a separate commit to aid reviews.

scohen:
@mhanberg @lukaszsamson PTAL. I could only add a single reviewer fro some reason.

PR #2, opened 2025-02-03
Spike runtime

mhanberg:
Pushing this up for posterity

PR #5, opened 2025-03-04
[feat] integrate refactorex

gp-pereira:
After a lot of consideration, I've decided to initially bring Refactorex as a dependency to the project. That way, I was able to quickly integrate all the refactorings I've built last year at once, which will gain us some time to incorporate them one by one in the near future. The code action handler behaviour is pretty similar to the controller I had on Refactorex. 

The downside of this approach is that GenLSP as also brought as a dependency to remote control. I've tried to avoid this by turning it into an optional dependency on Refactorex, but the lib couldn't compile after that. I hope that is not such a big problem though. 

Two things I've noticed and would like to address are that:
1. Expert currently suggests and performs the refactorings in a single step, instead of the two step approach where first the available refactorings are shown to the user and then only the chosen one is performed. With a few changes, Refactorex was also able to work like that but I think that can be a performance hit, but nothing I personally felt while testing.
2. Expert does not differentiate automatic refactorings (suggested by just placing the cursor somewhere) and deliberate refactorings (user asked to refactor), which can also cause performance problems, especially considering the point above. 

I've commented a few formatter options because they were bugging the refactorings (still don't fully understand why, but I believe converting an AST to string uses the project formatter and the injected remote_control doesn't have common, proto, etc as dependencies). I've also commented two tests that were broken in main, but I can fix them if they are important tests :) 

From my testing, all refactorings from Refactorex are indeed working, so I think we can start planning our next steps.


apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
```diff
@@ -0,0 +1,43 @@
...
+  @behaviour CodeAction.Handler
+
+  @impl CodeAction.Handler
+  def actions(%Document{} = doc, %Range{} = range, _diagnostics) do
+    original = Document.to_string(doc)
+
+    range = update_in(range.start.line, &(&1 - 1))
+    range = update_in(range.start.character, &(&1 - 1))
+    range = update_in(range.end.line, &(&1 - 1))
+    range = update_in(range.end.character, &(&1 - 1))
```
scohen: [question] What's the reasoning behind subtracting one here? There's automatic conversion from LSP (zero-indexed, utf16) positions to Elixir compatible positions (one-based, utf8). This change seems to want LSP compatible positions (zero-based). 

Converting a position blindly like this is almost certainly a bug.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
```diff
@@ -0,0 +1,43 @@
...
+
+    case Refactorex.Parser.parse_inputs(original, range) do
```
scohen: [perf] Can refactorex have an API where you don't have to re-parse the document? Lexical gets a lot of its speed from not having to re-parse documents over and over.

This call will happen _all the time_ in some editors, pretty much whenever the cursor moves, performance of this code is pretty critical.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
```diff
@@ -0,0 +1,43 @@
...
+      {:ok, zipper, selection_or_line} ->
+        zipper
```
scohen: ok, this is going to seem pretty trivial, but zippers are much, much slower (Dorgan measured them as 3-4x slower) than using elixir-native AST operations, which is why Lexical doesn't use them that much.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
```diff
@@ -0,0 +1,43 @@
...
+        |> Refactorex.Refactor.available_refactorings(selection_or_line, true)
+        |> Enum.map(fn refactoring ->
+          CodeAction.new(
+            doc.uri,
+            refactoring.title,
+            map_kind(refactoring.kind),
+            Changes.new(doc, Diff.diff(doc, refactoring.refactored))
```
scohen: this is pretty clean.

apps/remote_control/mix.exs
```diff
@@ -49,7 +49,9 @@ defmodule Lexical.RemoteControl.MixProject do
       {:patch, "~> 0.15", only: [:dev, :test], optional: true, runtime: false},
       {:path_glob, "~> 0.2", optional: true},
       {:phoenix_live_view, "~> 1.0", only: [:test], optional: true, runtime: false},
-      {:sourceror, "~> 1.7"}
+      {:sourceror, "~> 1.7"},
+      {:stream_data, "~> 1.1", only: [:test], runtime: false},
+      {:refactorex, "~> 0.1.48"}
```
scohen: do you want to be this specific? 
```suggestion
      {:refactorex, "~> 0.1"}
```

apps/remote_control/test/lexical/remote_control/code_action/handlers/refactorex_test.exs
```diff
@@ -0,0 +1,59 @@
...
+      document
+      |> Refactorex.actions(range, [diagnostic])
+      |> Enum.find(& &1.title == options[:title])
+      |> then(& &1.changes.edits)
+
+    {:ok, changes}
+  end
+
+  test "check if refactorex application is started" do
+    assert Enum.any?(Application.started_applications(), &match?({:refactorex, _, _}, &1))
```
scohen: this test doesn't feel like it's testing anything. Refactorex is inside the `mix.exs` in this application, it should be started, no?

apps/remote_control/test/lexical/remote_control/code_action/handlers/refactorex_test.exs
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  test "underscore variables not used" do
+    {:ok, result} =
+      ~q[
+        def my_func(unused) do
+        end
+      ]
+      |> modify(range: {{1, 1}, {1, 1}}, title: "Underscore variables not used")
```
scohen: We prefer to mark ranges of things inside the initial code. These changes are marked with angle quotes for ranges(`«»`). and a position is marked with a pipe character `|`.  This is done to avoid having to deal with ranges like this, which will be confusing to anyone rather quickly, especially after coming back and reading tests after taking a break. 

There's a lot of helpers for this available in `Lexical.Test.RangeSupport` and `Lexical.Test.CursorSupport`.

apps/server/.formatter.exs
```diff
@@ -10,6 +10,6 @@ locals_without_parens = [with_progress: 3]
 
 [
   locals_without_parens: locals_without_parens,
-  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"],
-  import_deps: imported_deps
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+  # import_deps: imported_deps
```
scohen: [question] What's the reasoning here?

apps/remote_control/test/lexical/remote_control/build_test.exs
```diff
@@ -239,6 +239,7 @@ defmodule Lexical.BuildTest do
       end
     end
 
+    @tag :skip
```
scohen: [question] What's the reasoning here?

apps/protocol/.formatter.exs
```diff
@@ -1,5 +1,5 @@
 # Used by "mix format"
 [
-  import_deps: [:proto],
+  # import_deps: [:proto],
```
scohen: Why was this removed?

apps/remote_control/.formatter.exs
```diff
@@ -20,7 +20,7 @@ locals_without_parens = [with_progress: 2, with_progress: 3, defkey: 2, defkey:
 [
   locals_without_parens: locals_without_parens,
   export: [locals_without_parens: locals_without_parens],
-  import_deps: import_deps,
+  # import_deps: import_deps,
```
scohen: Why was this removed?

apps/remote_control/lib/lexical/remote_control/code_action.ex
```diff
@@ -40,23 +41,13 @@ defmodule Lexical.RemoteControl.CodeAction do
...
-    results
-  end
-
-  defp applies?(:all, _handler_module) do
-    true
+    Enum.flat_map(@handlers, fn handler ->
+      if applies?(kinds, handler),
+        do: handler.actions(doc, range, diagnostics),
+        else: []
+    end)
```
scohen: [nit] This doesn't seem any more readable than the old, full `if`.

scohen:
>  I hope that is not such a big problem though.

It kind of is. There was a pretty strong desire to keep LSP concepts out of the injected app. I'm not going to die on that hill, but having our own way of handling things internally is nice for contributors, though since they come in as a transitive dep, maybe it's not a big deal.

scohen:
> After a lot of consideration, I've decided to initially bring Refactorex as a dependency to the project.

This is great, I should have expressed that the whole design of remote_control is so that we can bring in any dependencies we want.

apps/protocol/.formatter.exs
scohen: Never mind, you explained in your comment, but we need to investigate why this happens.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: My parser also does some conversion, maybe that's why it was needed. I heavily tested this part, since it's probably the most sensitive for my refactorings to properly work and subtracting one everywhere made it. Subtract one for the line was kinda expected, but for the columns was a surprise for me as well. We can def investigate further!

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: Yeah, that's what I tried to address in the two notes I left in the description. First, I think we can split suggesting available refactorings (super quick, just some conditionals) from the actual refactoring. Then, I would like to prevent automatically triggered refactored (context.trigger_kind = 0) from evaluating all refactorings, maybe just quickfixes? 

With the refactorings working, we can now deep dive into improving their performance, although I've never felt it was too slow, always had the available options instantly, even in expert after combining suggesting and performing them.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: I can see that, but I honestly think they are amazing abstractions for traversing the AST. If we can create (maybe you already have) abstractions to do so directly on the tree, I would gladly switch. However, maybe we should consider improving the performance of Sourceror, since it seems like the #1 solution for AST manipulation in Elixir.

apps/remote_control/test/lexical/remote_control/code_action/handlers/refactorex_test.exs
gp-pereira: oh yes, I was testing without having it started, but then I couldn't leverage its task supervision and had to rebuilt, so I decided to started it and just flipped the assert. No more need for it, althought I had to start it manually on test_helper.ex

apps/remote_control/test/lexical/remote_control/code_action/handlers/refactorex_test.exs
gp-pereira: Oh nice! I have some pretty similar on Refactorex, this is what its test look like.

``` 
 test "extracts the selected alias and declare it on top" do
    assert_refactored(
      ExtractAlias,
      """
      defmodule Foo do
        def foo() do
        # v
          Foo.Bar.bar()
        #       ^
        end
      end
      """,
      """
      defmodule Foo do
        alias Foo.Bar

        def foo() do
          Bar.bar()
        end
      end
      """
    )
  end
```

This "framework" I've built allowed me to quickly developed lots of tests, but I can use the already built abstractions as well. Those test are there more to prove the refactorings work than to really test them.

apps/server/.formatter.exs
gp-pereira: Answered it above, but we should investigate further.

apps/remote_control/test/lexical/remote_control/build_test.exs
gp-pereira: Deactivated two tests that were already broken in the main branch so that I could have all tests in remote control working while developing, but I can reactivate them if you want!

apps/remote_control/.formatter.exs
gp-pereira: Answered it above, but we should investigate further.

apps/remote_control/lib/lexical/remote_control/code_action.ex
gp-pereira: removing the assingment followed by the return seemed right to me (although I've noticed this pattern is used a lot in this repo), but the if part is indeed optional

gp-pereira:
> Refactorex adds capability for refactorings, but these need to be reported on boot. Currently, the supported refactorings are listed in `apps/server/lib/lexical/server/state.ex` in a module attribute. I believe this needs to be updated.
> 
> ```elixir
>   @supported_code_actions [
>     :quick_fix,
>     :source_organize_imports
>   ]
> ```

They work without it, I think that is mostly for reference, but I'll add them, no problem!


> > I hope that is not such a big problem though.
> 
> It kind of is. There was a pretty strong desire to keep LSP concepts out of the injected app. I'm not going to die on that hill, but having our own way of handling things internally is nice for contributors, though since they come in as a transitive dep, maybe it's not a big deal.

Absolutely. The GenLSP usage on Refactorex is for it working in its own, since only the refactorings are being used, I also would like to remove this dependency is the near future. Can we move along and refactor this later? 

Besides all that, how do you think we should move forward? keep the refactorings (edits in the ASTs) apart from the LS and use a combination of it and the features from the LS for some specific refactors or bring all of them in and have them only here? I think #2 is more powerful/easier for the future, but #1 could be cool for other projects to also use them.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: I think it's worth mentioning how I find the range in the source code, since your approach is probably different. 
I basically "crop" the selection, parse as its own AST and then just look for it inside the complete AST. So this:

```
#       v
def foo(arg) do
#         ^
end
```

becomes something like this (selectors kept for reference, but are actually removed)

```
#       v
        arg
#         ^

```

This approach made super easy to find what the user is selecting

apps/remote_control/test/lexical/remote_control/code_action/handlers/refactorex_test.exs
gp-pereira: done!

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
scohen: The problems arise when there's a unicode character prior to the range's start. It's horrible naming, but what microsoft terms a `character` is actually a unicode code unit offset. Say we have the string "hi 🎸|" where `|` represents the position. Subtracting 1 will place the position inside the guitar emoji. To place it before the guitar emoji, you'll need to subtract 4.   

> My parser also does some conversion, maybe that's why it was needed.

I'd like to hear more.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
scohen: Expert has the ability to return fragments of the document (as both strings and AST), would this be helpful to you?
Fragments on a single line are insanely efficient. Reparsing the document, markedly less so.

apps/remote_control/lib/lexical/remote_control/code_action.ex
scohen: actually, when you find that pattern, it's likely a remnant of debugging. Feel free to remove it.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
scohen: > and subtracting one everywhere made it.

Subtracting one will work for all text that can be represented by a single code unit in UTF8, which I *think* is all ASCII. Additionally, if the position is before a unicode character

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
scohen: I've talked with Dorgan about this, and we didn't see a lot of opportunities for improvement here. We're running smack into the limits elixir's immutability puts on us.

scohen:
> Can we move along and refactor this later?

Of course. 

> Besides all that, how do you think we should move forward? 

My bias is for power and ease of contributions, which I think is #2, and I suspect that involves deeper integration with things like the AST modules that are defined in Expert's `core` project. We might be able to publish that, as a library, as it's a completely separate mix project.

apps/remote_control/lib/lexical/remote_control/code_action.ex
mhanberg: I think steve meant the

```
if something?,
  do: :ok,
  else: :oh_no
```

rather than

```
if something? do
  :ok
else
  :oh_no
end
```

which i agree with, personally

scohen:
Note: I tried this branch and saw the same errors coming out of the formatter. 
Does refactorex do anything at the project level?

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
lukaszsamson: This `- 1` is in UTF8 graphemes or UTF16 code units? If graphemes then it's OK, if code units then when moving 1 code point left (or 2 bytes) one needs to check if the code point to the left is not a high surrogate pair (0xD800..0xDBFF)

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: I did some tests and it's almost what I need! Unfortunately, it loses the position metadata from the node, because it just returns the string. I use the metadata to find the exact node in the complete AST. Can we adapt it somehow? If so, I could stop changing the range directly as commented above

apps/remote_control/lib/lexical/remote_control/code_action.ex
gp-pereira: Oh, yes, I just answered why I started editing this place (removing the needless assignment) but got carried away haha. Not clear to me what version do you prefer tho @mhanberg. Should I revert this then?

gp-pereira:
> Note: I tried this branch and saw the same errors coming out of the formatter. Does refactorex do anything at the project level?

Not explicitly. I use Sourceror.to_string, which I think uses Elixir.Code that might use the project formatter. 
But did you manage to get the refactorings working for you on emacs?

What do you think is missing for this branch to be merged?

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: What do you guys suggest I use instead? I'd like to use the original range, is there a way to do so for now?

gp-pereira: Solved by using Document.fragment

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: Solved by using Document.fragment

apps/server/.formatter.exs
gp-pereira: Reverted, solved by using the RemoteControl formatter for file

apps/protocol/.formatter.exs
gp-pereira: Reverted, solved by using the RemoteControl formatter for file

apps/remote_control/.formatter.exs
gp-pereira: Reverted, solved by using the RemoteControl formatter for file

apps/remote_control/lib/lexical/remote_control/code_action.ex
gp-pereira: Already changed after introducing context.trigger_kind

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
```diff
@@ -0,0 +1,74 @@
...
+  defp line_or_selection(doc, %{start: start} = range) do
+    [
+      # new lines before the selection
+      String.duplicate("\n", start.line - 1),
+      # same line whitespace before the selection
+      String.duplicate(" ", start.character - 1),
+      # the selection
+      Document.fragment(doc, range.start, range.end)
+    ]
+    |> IO.iodata_to_binary()
```
scohen: is all of this just so that the metadata of lines and characters aligns? Some of the parsing functions take a line and column, I wonder if we can get Dorgan to add this to `Sourceror.parse_string`.

apps/remote_control/lib/lexical/remote_control/code_action.ex
```diff
@@ -30,33 +32,40 @@ defmodule Lexical.RemoteControl.CodeAction do
...
-      end)
 
-    results
-  end
+        not handle_trigger_kind?(handler, trigger_kind) ->
+          []
 
-  defp applies?(:all, _handler_module) do
-    true
+        true ->
```
scohen: i'm not  a huge fan of conds with negative conditions. I think this is just 
```elixir
if handle_kinds?(handler, kinds) or handle_trigger_kind?(handler, trigger_kind) do 
   handler.actions(doc, range, diagnostics)
else 
  []
end
```

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -33,7 +33,9 @@ defmodule Lexical.RemoteControl do
 
   defdelegate list_modules, to: :code, as: :all_available
 
-  defdelegate code_actions(document, range, diagnostics, kinds), to: CodeAction, as: :for_range
+  defdelegate code_actions(document, range, diagnostics, kinds, trigger_kind),
+    to: CodeAction,
+    as: :for_range
```
scohen, line 38: there are a number of formatting changes like this in the PR, why did they happen? Shouldn't we be using `mix fomat`?

apps/remote_control/lib/lexical/remote_control/code_action/handlers/add_alias.ex
```diff
@@ -34,15 +34,15 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.AddAlias do
...
-        []
+      _ -> []
     end
   end
 
   @impl CodeAction.Handler
-  def kinds do
-    [:quick_fix]
-  end
+  def kinds, do: [:quick_fix]
```
scohen: these changes feel unnecesary

apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
```diff
@@ -26,15 +26,15 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.OrganizeAliases do
...
-        []
+      _ -> []
     end
   end
 
   @impl CodeAction.Handler
-  def kinds do
-    [:source, :source_organize_imports]
-  end
+  def kinds, do: [:source, :source_organize_imports]
```
scohen: these changes feel unnecessary

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
scohen: in thee meantime, we can add this function to `AST` to get you what you want:
```elixir
def parse_string(string, opts) do
  
  opts = Keyword.merge(opts, [    [
      literal_encoder: &{:ok, {:__block__, &2, [&1]}},
      token_metadata: true,
      unescape: false,
      columns: true,
      warn_on_unnecessary_quotes: false,
      emit_warnings: false
    ])
    with {:ok, quoted, comments} <- Sourceror.string_to_quoted(source, to_quoted_opts()) do
      {:ok, Sourceror.Comments.merge_comments(quoted, comments)}
    end
   end
```

scohen: I really want to keep the usage of libraries like Sourceror confined to a module or two.

apps/remote_control/test/lexical/remote_control/build_test.exs
scohen: I haven't seen these fail in main. Which versions of elixir / erlang are you running under?

apps/remote_control/test/lexical/remote_control/code_action/handlers/refactorex_test.exs
```diff
@@ -0,0 +1,79 @@
...
+        def my_func() do
+          Enum.map(1..2, &extracted_function(&1))
+        end
+
+        defp extracted_function(i) do
+          i + 20
+        end
+      end]
+    )
+  end
```
scohen, line 85: are  there more?

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
doorgan: This is now supported in v1.8: https://hexdocs.pm/sourceror/Sourceror.html#parse_string/2

apps/remote_control/lib/lexical/remote_control.ex
gp-pereira: It is the mix formatter who is doing that tho. I retried to save this change in a single line and it gets converted to the multiline. We can put the mix format in a precommit hook as well if you want. In this case, i think the line got too long.

apps/remote_control/lib/lexical/remote_control/code_action.ex
gp-pereira: Changed! It's actually an AND tho

apps/remote_control/test/lexical/remote_control/build_test.exs
gp-pereira: Reverted! I was running on Elixir 1.13

apps/remote_control/lib/lexical/remote_control/code_action/handlers/add_alias.ex
gp-pereira: Reverted!

apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
gp-pereira: Reverted!

apps/remote_control/test/lexical/remote_control/code_action/handlers/refactorex_test.exs
gp-pereira: the tests were poorly named. They are testing whether the refactorex integration works with a cursor position, a single line selection and a multiline selection. I've updated their names tom reflect that. The actual testing of the refactorings will come with the refactorings. This should just be testing the integration.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
doorgan: I talked with @gp-pereira about this, while adding those options to sourceror should have been enough, we were seeing some unexpected behavior where the `:column` option would behave as adding indentation to the whole code snippet instead of setting the start column for the source code parsing.

I started working on a solution for this in sourceror(essentially line 46 in the snippet from this comment), but testing against older elixir versions I found that elixir used to behave the way we wanted here but changed in 1.16 onwards. I opened an issue because I think it's a bug: https://github.com/elixir-lang/elixir/issues/14390

gp-pereira: In the mean time, @scohen do you want me to move this to AST and the next function to some CodeMod module? Or can we keep it like this? For the refactorings, I use Sourceror and Zipper in most modules, should I try to find abstractions to all their usages?

scohen: > do you want me to move this to AST and the next function to some CodeMod module?

Yes

>  For the refactorings, I use Sourceror and Zipper in most modules, should I try to find abstractions to all their usages

That would be my preference. I like to keep external libs fairly isolated. For example, we use the `Code` module for parsing in one place, which makes swapping to `Sourceror` or @mhanberg 's parser firefly very easy.  If we had `Code` spread throughout the codebase, it would be pretty tough to do this.

That said, you can land this as-is and address isolating sourceror in a follow-on PR.

doorgan: For visibility: https://github.com/elixir-lang/elixir/issues/14390 is fixed on elixir main and a fix was also backported to sourceror, we should be able to just set the `:column` option here and get the results we expect

apps/remote_control/lib/lexical/remote_control.ex
scohen: I think you're running this on a different version of elixir than the project is compiled under. We usually run under 1.14 to ensure backwards compatibility.

gp-pereira: I believe I've already reverted all formatting changes that weren't necessary. For this one specifically, the line got long so even more recent formatters will put it in multiple lines. I usually run on v1.13, but will use 1.14 from now on.

apps/remote_control/lib/lexical/remote_control/code_action/handlers/refactorex.ex
gp-pereira: It worked perfectly and the formatter option too! Thanks @doorgan

PR #6, opened 2025-03-04
[bugfix] Only index if there's stuff to index

scohen:
A colleague found a fun edge case. He added three new empty files to a project, and Expert wouldn't start. This is because of a divide by zero error when there are no bytes to index, which causes progress updates to fail (indexing completed successfully).

This fix checks for this condition and no-ops if there's nothing to index.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
@@ -101,57 +101,62 @@ defmodule Lexical.RemoteControl.Search.Indexer do
 
     total_bytes = paths_to_sizes |> Enum.map(&elem(&1, 1)) |> Enum.sum()
 
-    {on_update_progess, on_complete} = Progress.begin_percent("Indexing source code", total_bytes)
+    if total_bytes > 0 do
+      {on_update_progess, on_complete} =
+        Progress.begin_percent("Indexing source code", total_bytes)
 
-    initial_state = {0, []}
+      initial_state = {0, []}
 
-    chunk_fn = fn {path, file_size}, {block_size, paths} ->
-      new_block_size = file_size + block_size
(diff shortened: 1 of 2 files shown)


apps/remote_control/lib/lexical/remote_control/search/indexer.ex
```diff
@@ -101,57 +101,62 @@ defmodule Lexical.RemoteControl.Search.Indexer do
...
+      paths_to_sizes
+      |> Stream.chunk_while(initial_state, chunk_fn, after_fn)
+      |> Task.async_stream(
+        fn chunk ->
+          block_bytes = chunk |> Enum.map(&Map.get(path_to_size_map, &1)) |> Enum.sum()
+          result = Enum.map(chunk, processor)
+          on_update_progess.(block_bytes, "Indexing")
+          result
+        end,
+        timeout: timeout
```
lukaszsamson, line 138: What do we want to do on timeout? Fail the chunk or exit the caller process? I do not like the default behavior and always add `on_timeout: :kill_task`

scohen: the timeout is :infinity by default, and I don't think it's changed anywhere.

scohen: FWIW, @lukaszsamson, this isn't new code, I just added an if statement. It might help to view the diff with ignore whitespace turned on.

lukaszsamson: I know

scohen: So, do you think we need to do anything here?

lukaszsamson: No, it’s fine

PR #7, opened 2025-04-01
[chore] Documentation housekeeping

Moosieus:
* Move install instructions to a single page.
* Drop some details that no longer apply (Expert being an umbrella application)
* Centralize architecture docs and expand on them.
* Add script for connecting to project nodes, which I've found useful in the past.

I've written these docs as though they were oriented to novel contributors looking to familiarize themselves with Expert. All of this will of course be heavily subject to change, but I think it's worthwhile to get the ball rolling, even if it's a token effort at this point.

### DIFF
--- a/pages/installation.md
+++ b/pages/installation.md
@@ -27,10 +27,10 @@ Then, clone Expert's git repository:
 git clone git@github.com:elixir-lang/expert.git
 ```
 
-Then change to the lexical directory:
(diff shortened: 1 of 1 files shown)


scohen:
I'm not sure that we should do docs at this point. We're not super sure what's going to be included in the resultant project, and it's almost certain that the docs will need to be changed anyways.

PR #9, opened 2025-04-03
[fix] Determine if completion is displayable based on the project dependencies

### LINKED ISSUE #8: Incorrect dependency filtering for completions
I was working on a project, `telemetry_metrics_statsd`, and completions for its struct modules failed. This is because in `server/code_intelligence/completion.ex`, the function `displayable?` checks to see if the emitted completion is one of lexical's dependencies, and if it is, says that it's not displayable. There are two bugs here. 

The first is that it just does a prefix check, and thinks `TelemetryMetricsStatsd.UDP...` is a module in `Telemetry`. This is wrong, there should be a more rigorous module prefix check rather than a substring check.

The second bug is that we're pulling all of the project's dependencies, and expert might depend on a module that your project depends on, and you definitely want to see that show up in completions.

I think we should get a list of the project under test's dependencies and do a check to see if the module is included in them in order to see if it's displayable.

doorgan:
Fixes #8

This shifts the approach from "is this suggestion part of RemoteControl", which requires quite a lot of special casing, to "is this suggestion part of the project" instead.

The approach here is to:

1. Determine the application that defined the suggested module
2. Determine the dependency applications of the project
3. Check if the application from 1 is part of any dependency of the project, or any of the applications loaded by default by elixir

I think we should generally avoid doing string checks on module names to determine where they beling to, since a project may define modules in the same namespace as some other project. For example, if you are editing a different project/application that defines modules in the `Lexical` namespace: you'd want completions for the modules you define but not for the ones in `Lexical.RemoteControl`.

For `2`, which is the hardest part here, I decided to parse the `.app` files in the build directory. The rationale here is that:

- As soon as the project is compiled, which expert does as-you-type, the app files get updated
- We can trust that it only contains applications related to the project and nothing else, it's not polluted by RemoteControl

What I experienced with other options was:
- `mix.lock` files include dependencies, but not `extra_applications` or applications loaded by elixir by default
- The result from `Mix.Project.deps_apps()` is sometimes `[]` or out of date, which seemed quite unreliable
- The results from `:application_controller.which_applications()`, besides apparently being private API, also include RemoteControl dependencies, applications, and namespaced applications

The downside is that we need to parse the `.app` files a few times whenever we have a bunch of suggestions to check.

### DIFF
--- a/apps/engine/lib/engine/engine.ex
+++ b/apps/engine/lib/engine/engine.ex
@@ -5,12 +5,13 @@ defmodule Engine do
   context of the remote VM.
   """
 
-  alias Forge.Project
-
   alias Engine.Api.Proxy
   alias Engine.CodeAction
   alias Engine.CodeIntelligence
   alias Engine.ProjectNode
+  alias Forge.Project
+
+  alias Mix.Tasks.Namespace
 
   require Logger
 
@@ -63,6 +64,12 @@ defmodule Engine do
   defdelegate document_symbols(document), to: CodeIntelligence.Symbols, as: :for_document
 
   defdelegate workspace_symbols(query), to: CodeIntelligence.Symbols, as: :for_workspace
+
+  def list_apps do
+    for {app, _, _} <- :application.loaded_applications(),
+        not Namespace.Module.prefixed?(app),
+        do: app
+  end
 
   def start_link(%Project{} = project) do
     :ok = ensure_epmd_started()
--- a/apps/engine/lib/engine/engine/api.ex
+++ b/apps/engine/lib/engine/engine/api.ex
@@ -1,12 +1,11 @@
 defmodule Engine.Api do
+  alias Engine.CodeIntelligence
   alias Forge.Ast.Analysis
   alias Forge.Ast.Env
   alias Forge.Document
   alias Forge.Document.Position
   alias Forge.Document.Range
   alias Forge.Project
-
-  alias Engine.CodeIntelligence
 
   require Logger
 
@@ -33,6 +32,10 @@ defmodule Engine.Api do
 
   def list_modules(%Project{} = project) do
     Engine.call(project, Engine, :list_modules)
+  end
+
+  def project_apps(%Project{} = project) do
+    Engine.call(project, Engine, :list_apps)
   end
 
   def format(%Project{} = project, %Document{} = document) do
--- a/apps/engine/lib/mix/tasks/namespace.ex
+++ b/apps/engine/lib/mix/tasks/namespace.ex
@@ -15,6 +15,9 @@ defmodule Mix.Tasks.Namespace do
   use Mix.Task
 
   @dev_deps [:patch]
+  # Unless explicitly added, nimble_parsec won't show up as a loaded app
+  # and will therefore not be namespaced.
+  @no_app_deps [:nimble_parsec]
 
   # These app names and root modules are strings to avoid them being namespaced
   # by this task. Plugin discovery uses this task, which happens after
@@ -27,15 +30,13 @@ defmodule Mix.Tasks.Namespace do
     "forge" => "Forge"
   }
 
-  @deps_apps Engine.MixProject.project()
-             |> Keyword.get(:deps)
-             |> Enum.map(&elem(&1, 0))
-             |> then(fn dep_names -> dep_names -- @dev_deps end)
-             |> Enum.map(&to_string/1)
-
   require Logger
 
   def run([base_directory]) do
+    # Ensure we cache the loaded apps at the time of namespacing
+    # Otherwise only the @extra_apps will be cached
+    init()
+
     Transform.Apps.apply_to_all(base_directory)
     Transform.Beams.apply_to_all(base_directory)
     Transform.Scripts.apply_to_all(base_directory)
@@ -115,10 +116,31 @@ defmodule Mix.Tasks.Namespace do
   end
 
   defp init do
-    @deps_apps
-    |> Enum.map(&String.to_atom/1)
+    discover_deps_apps()
+    |> Enum.concat(@no_app_deps)
+    |> then(&(&1 -- @dev_deps))
     |> root_modules_for_apps()
     |> Map.merge(extra_apps())
     |> register_mappings()
+  end
+
+  defp discover_deps_apps do
+    cwd = File.cwd!()
+
+    :application.loaded_applications()
+    |> Enum.flat_map(fn {app_name, _description, _version} ->
+      try do
+        app_dir = Application.app_dir(app_name)
+
+        if String.starts_with?(app_dir, cwd) do
+          [app_name]
+        else
+          []
+        end
+      rescue
+        _ -> []
+      end
+    end)
+    |> Enum.sort()
   end
 end
--- a/apps/engine/test/fixtures/project/mix.exs
+++ b/apps/engine/test/fixtures/project/mix.exs
@@ -16,7 +16,7 @@ defmodule Project.MixProject do
   # Run "mix help compile.app" to learn about applications.
   def application do
     [
-      extra_applications: [:logger]
+      extra_applications: [:logger, :erts]
     ]
   end
 
--- a/apps/expert/lib/expert/code_intelligence/completion.ex
+++ b/apps/expert/lib/expert/code_intelligence/completion.ex
@@ -17,9 +17,8 @@ defmodule Expert.CodeIntelligence.Completion do
 
   require Logger
 
+  @build_env Mix.env()
   @expert_deps Enum.map([:expert | Mix.Project.deps_apps()], &Atom.to_string/1)
-
-  @expert_dep_modules Enum.map(@expert_deps, &Macro.camelize/1)
 
   def trigger_characters do
     [".", "@", "&", "%", "^", ":", "!", "-", "~"]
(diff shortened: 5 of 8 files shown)


lukaszsamson:
I think this approach misses `extra_applications` from project and deps, beam apps like `stdlib` that elixir auto loads or preloaded `erts` app.

Note 1 about application controller: elixir compiler does not update entries there on recompilation.
Note 2 about application controller: `erts` app is not loaded by default and yet modules from there are available

doorgan:
@lukaszsamson another alternative I see is to read the `.app` files in the project's build directory (`.lexical/build/erl-x/elixir-y/test/lib/<app>/ebin/<app>.app`)
Those files include all the applications and optional applications. The downside of course is again having to read files from disk(many of them this time), but it does have a comprehensive and correct list of applications, and it does get udpated on recompilation.

29d61f0 implements the approach I mention here. I think we might also be able to cache those, but if we have as-you-time recompilation I'd have to measure to know if it's really worth it

apps/remote_control/test/fixtures/project/mix.exs
```diff
@@ -25,6 +25,7 @@ defmodule Project.MixProject do
     [
       # {:dep_from_hexpm, "~> 0.3.0"},
       # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
+      {:stream_data, "~> 1.2"}
```
doorgan: There was a test that checked that a SreamData module was being completed. It worked before because we only checked if the module wasn't a Lexical one, but with the new approach that checks if module is part of the project, we actually need the dependency to be installed.

doorgan:
I've been having quite a lot of bad luck with the CI on a flaky test here

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -223,16 +222,60 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
...
+    project_app = Application.get_application(project.project_module)
+
+    metadata = Map.get(result, :metadata)
+
+    result_app = metadata[:app]
+
+    # we consider it a "project module" if:
+    # 1. the module is in the project apps
+    # 2. there is no module app and there is not a project module
+    #    (this is the case for some test completions)
```
doorgan: Alternatively we can make the tests correctly recognize all project defined modules as part of the project application, still getting familiarized with how this happens

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -223,16 +222,60 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
...
-          end
-        end)
+        project_module?(project, project_apps, suggested_module, result)
     end
   end
 
+  defp get_project_apps(project) do
+    build_path = RemoteControl.Build.path(project)
+
+    for app <- File.ls!(Path.join([build_path, "test", "lib"])),
```
doorgan: I'm assuming here that by the time we try to get completions, RemoteControl is already started and the project is already compiled. I haven't found issues using a build of this PR for a while, but it's still an assumption.

scohen: I think this function should be pushed into `remote_control` since it's in control of the project and what's going on. I think that doing operations inside of remote control will allow you to leverage the mix project. 

Having this logic in the server app was, i think, a mistake.

doorgan: > I think that doing operations inside of remote control will allow you to leverage the mix project.

Yes and no; Mix does not always produce the list of project applications. I'm not sure why exactly, but I would often get an empty list of applications back from Mix. It would also, with current `Mix.Project` apis, miss both the project and the dependencies extra applications.

Loading the `.app` files doesn't require an rpc call to the remote control node, but I see the point of keeping project related logic in the remote control app, I can move this code to that app

apps/common/lib/lexical/ast/module.ex
```diff
@@ -96,4 +100,16 @@ defmodule Lexical.Ast.Module do
 
     {type, split_module}
   end
+
+  @spec to_atom(module() | String.t()) :: module()
+  def to_atom(module) when is_atom(module) do
+    module
+  end
+
+  def to_atom(module_string) when is_binary(module_string) do
```
scohen: is this safe? what happens if we pass in `:some_module` here?

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -243,4 +243,29 @@ defmodule Lexical.RemoteControl do
...
+  Includes:
+  - The project app itself
+  - All dependency applications, both direct and transitive
+  - Optional/extra applications from the project and dependencies
+  """
+  @spec project_apps(Project.t()) :: [atom()]
+  def project_apps(project) do
+    build_path = RemoteControl.Build.path(project)
+
+    for app <- File.ls!(Path.join([build_path, "test", "lib"])),
```
scohen: can we not use `Mix.Project` here to do this work?

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -223,16 +222,45 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
...
+    # 2. there is no module app and there is not a project module
+    #    (this is the case for some test completions)
+    # 3. the module app is the same as the project app
+    # 4. the module is not part of the app, but is the same as the project module
+    # 5. the completion is annotated as part of an app that is in the project apps
+    module_app in project_apps or (is_nil(module_app) and is_nil(project.project_module)) or
+      (not is_nil(module_app) and module_app == project_app) or
+      (is_nil(module_app) and not is_nil(project.project_module) and
+         module == project.project_module) or
+      (is_nil(metadata) or result_app in project_apps)
```
scohen: would pushing this into the remote control app help things? 

Also, i think a cond would be a lot more readable than this very nested statement.

apps/remote_control/lib/lexical/remote_control.ex
scohen: consulting the filesystem on every module completion will cost a lot of performance

doorgan: > can we not use `Mix.Project` here to do this work?

I tried, it doesn't yield good results. As an example, I just tried this on an `igniter` clone and I get these results from the RemoteControl node:

```
Mix.Project.apps_paths: nil
Mix.Project.deps_apps: []
```

Running completions a few times I eventually get this:

```
Mix.Project.apps_paths: nil
Mix.Project.deps_apps: [:sourceror, :earmark_parser, :text_diff, :file_system, :deep_merge, :decimal, :spitfire, :eflame, :mime, :nimble_options, :nimble_parsec, :bunt, :telemetry, :ex_check, :jason, :doctor, :statistex, :mix_test_watch, :ham, :hpax, :mint, :yamerl, :yaml_elixir, :mix_audit, :glob_ex, :credo, :phx_new, :makeup, :makeup_elixir, :makeup_erlang, :ex_doc, :inflex, :erlex, :dialyxir, :nimble_pool, :finch, :req, :rewrite, :benchee, :owl, :mimic]
```

Which is *close*, but is missing `:logger`, `:public_key`, `:ssl`, `:inets`, some of which are loaded by elixir, some of which were defined as `extra_applications` in one of the mix files. As far as I can tell, there is no public Mix api to get those, so we need to resort to reading manifests in the filesystem to get *everything*(which is close to what mix does anyways after compilation).

Also, adding a dependency causes `Mix.deps_apps` to return `[]` again for a while until it catches up.

> consulting the filesystem on every module completion will cost a lot of performance

I was mulling over this too, I ultimately decided to fetch the project applications after compilation, and keep them in memory for completions.

apps/server/lib/lexical/server/code_intelligence/completion.ex
doorgan: > would pushing this into the remote control app help things?

No, there is something in the test setup that is preventing the application controller from recognizing the Project fixture modules as part of the `:project` application, I'm having a bit of a hard time figuring out what is it exactly.

> Also, i think a cond would be a lot more readable than this very nested statement.

Updated in 0ff39e9

apps/remote_control/lib/lexical/remote_control.ex
scohen: the missing apps are all in erlang though, and should be available for completions.

doorgan: That's a good call, I added the builtin apps from both erlang and elixir in acc6565. Short of an api to list them, I had to hardcode them.

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -62,6 +62,8 @@ defmodule Lexical.RemoteControl do
 
   defdelegate workspace_symbols(query), to: CodeIntelligence.Symbols, as: :for_workspace
 
+  defdelegate list_apps(), to: RemoteControl.Build.ApplicationCache, as: :list_apps
```
scohen: is the `as:` necessary here?

doorgan: Trying without it seems to work, so no
This is an old commit though

doorgan:
This has been working as expected in my testing for quite a while, and I need the changes to namespacing in this PR for the burrito packaging work, so I'm merging it

PR #14, opened 2025-04-13
Replace Snowflake IDs with a cut down version

Moosieus:
Replace Snowflake identifiers with a cut down version:
* Machine ID was always `1`, so that is now omitted.
  * The leftover bits are given to the sequence ID instead.
* IDs generated the same millisecond are not sequenced in the order they were generated.
  * This obviates the need to block on incrementing an integer in a GenServer.
* ID generation will no longer error out when the runtime's clock rolls backwards.

Also added a `debug_project.sh` script. This is useful to check the ETS table size (`:ets.info(:lexical_search_v3, :size)`) and estimate time savings. Average benchmarks on my 12th gen i7 laptop are showing 2µs -> 100ns.

Benchee outputs to follow in PR.

### DIFF
--- a/nix/hash
+++ b/nix/hash
@@ -1,1 +1,1 @@
-sha256-i2IJ++lK1C/q+ETq/EYfNgGFKDTC/+boSwxAaT8rozw=
+sha256-eEBjOwQter0SDeJhWvHD7+NaAvbOCmoalYpZU35poKs=


Moosieus:
WSL Windows 10:
```
Operating System: Linux
CPU Information: 12th Gen Intel(R) Core(TM) i7-1270P
Number of Available Cores: 16
Available memory: 15.47 GB
Elixir 1.15.8
Erlang 26.2.5.10
JIT enabled: true
```

Snowflakes:
```
Benchmark suite executing with the following configuration:
warmup: 2 s
time: 5 s
memory time: 0 ns
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 7 s

Benchmarking next_id ...
Calculating statistics...
Formatting results...

Name              ips        average  deviation         median         99th %
next_id      780.34 K        1.28 μs  ±1682.62%        1.14 μs        2.21 μs

Profiling next_id with eprof...

Profile results of #PID<0.244.0>
#                                               CALLS     % TIME µS/CALL
Total                                              13 100.0   25    1.92
:gen.call/4                                         1  0.00    0    0.00
GenServer.whereis/1                                 1  0.00    0    0.00
GenServer.call/2                                    1  0.00    0    0.00
anonymous fn/0 in :elixir_compiler_1.__FILE__/1     1  0.00    0    0.00
:erlang.send/3                                      1  4.00    1    1.00
:erlang.demonitor/2                                 1  4.00    1    1.00
Snowflake.next_id/0                                 1  4.00    1    1.00
Process.whereis/1                                   1  4.00    1    1.00
:erlang.whereis/1                                   1  8.00    2    2.00
GenServer.call/3                                    1  8.00    2    2.00
:erlang.monitor/3                                   1 12.00    3    3.00
:erlang.apply/2                                     1 28.00    7    7.00
:gen.do_call/4                                      1 28.00    7    7.00

Profile done over 13 matching functions
```

PR:
```
Benchmark suite executing with the following configuration:
warmup: 2 s
time: 5 s
memory time: 0 ns
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 7 s

Benchmarking next_id ...
Calculating statistics...
Formatting results...

Name              ips        average  deviation         median         99th %
next_id       11.30 M       88.46 ns ±26844.54%          59 ns         116 ns

Profiling next_id with eprof...

Profile results of #PID<0.244.0>
#                                               CALLS     % TIME µS/CALL
Total                                               7 100.0    5    0.71
:erlang.unique_integer/1                            1  0.00    0    0.00
:os.system_time/1                                   1  0.00    0    0.00
anonymous fn/0 in :elixir_compiler_1.__FILE__/1     1  0.00    0    0.00
System.normalize_time_unit/1                        1  0.00    0    0.00
Lexical.Identifier.next_global!/0                   1 20.00    1    1.00
System.os_time/1                                    1 20.00    1    1.00
:erlang.apply/2                                     1 60.00    3    3.00

Profile done over 7 matching functions
```

apps/common/lib/lexical/identifier.ex
```diff
@@ -1,15 +1,31 @@
...
+  @seq_size 64 - @ts_size
+  @seq_max 2 ** @seq_size - 1
+  # First second of 2024 (milliseconds)
+  @epoch 1_704_070_800_000
+
+  @spec next_global! :: integer
   def next_global! do
-    {:ok, next_id} = Snowflake.next_id()
-    next_id
+    ts = System.os_time(:millisecond) - @epoch
```
lukaszsamson, line 16: System.monotonic_time is better choice here. os_time may be adjusted backwards which can theoretically lead to duplication

Moosieus: The blocker with `System.monotonic_time/1` is that its starting value is arbitrary set on restarts. `System.os_time/1` can rollback, but it's mostly consistent between restarts.

For Expert's purposes, the IDs used in the indexer don't need to be sequenced in the order they were generated, they just need to be unique.

In theory there could be collision here if:
* The language server restarted (resetting `:erlang.unique_integer/1`)
* The OS clock rolled back (rolling back `System.os_time/1`)
* An identifier was generated on the **same millisecond timestamp** with the **same unique integer**

The odds of that happening should be astronomically low.

Presently Expert is intolerant to rollbacks greater than 1ms (causes a MatchError and entering a bad state)

scohen:
I’m r- on this because we’re likely going to change how the index is stored, and won’t need to generate our ids anymore.

Moosieus:
I logged out the rollback times from Snowflake, they're between 12s and 2.5s on this particularly misbehaving laptop, and constantly so.

Closing this for now. If the index storage ends up staying on ETS and others run into time warp issues, this can serve as a point of reference.

PR #15, opened 2025-04-19
[chore] Drop support for Elixir 1.13, 1.14, and OTP 24

Moosieus:
* Set default Elixir version to 1.15.8
* Set default OTP version to 25.0
* Drop Elixir 1.13 and 1.14 from allowed versions and CI tests
* Drop OTP 24 from allowed versions and CI tests

This PR formally drops support for Elixir 1.13, 1.14, and OTP 24. It also raises the patch version of Elixir 1.15 to `>= 1.15.8`.


Moosieus:
Could also raise the patch version of 1.16 to `>= 1.16.3` to capture the same bug fixes applied in 1.15.8

PR #16, opened 2025-04-23
Rename project to Expert

doorgan:
App name changes:

common -> forge
remote_control -> engine
server -> expert
lexical_credo -> expert_credo

Modules in `proto` and `protocol` were prefixed with Expert

The namespacing prefix was changed from `LX` to `XP`
The lsp meta folder was changed from `.lexical` to `.expert`
The lsp logs file was changed from `lexical.log` to `expert.log`
The lsp start script was chaged from `start_lexical.sh` to `start_expert.sh`

The changelog was kept as-is as it reflects the project history(although it will likely be removed in favor of a new expert specific changelog file)

I did not update the documentation because right now might not be the best moment to do that (https://github.com/elixir-lang/expert/pull/7#issuecomment-2795455298)

### DIFF
--- a/apps/engine/lib/engine/engine/build/isolation.ex
+++ b/apps/engine/lib/engine/engine/build/isolation.ex
@@ -3,7 +3,7 @@ defmodule Engine.Build.Isolation do
   Runs functions in an isolated, monitored process
   """
 
-  @spec invoke((() -> term())) :: {:ok, term()} | {:error, term()}
+  @spec invoke((-> term())) :: {:ok, term()} | {:error, term()}
   def invoke(function) when is_function(function, 0) do
     me = self()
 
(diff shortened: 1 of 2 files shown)


apps/expert/test/expert/code_intelligence/completion_test.exs
```diff
@@ -9,7 +9,7 @@ defmodule Expert.CodeIntelligence.CompletionTest do
 
   describe "excluding modules from expert dependencies" do
     test "expert modules are removed", %{project: project} do
-      assert [] = complete(project, "Expert|")
+      assert [] = complete(project, "Expert.CodeIntelligence|")
```
doorgan, line 12: I had to change this because there is a bug on `main` that causes `ExUnitProperties` to showup for `Expert` completions. The bug itself is fixed in #9 but that PR also updates the fixture project to include `:stream_data` as a dependency, which would make this test to return `ExUnitProperties` too.

This still makes sure no expert module is being suggested, so I think it's fine to keep this updated test.

doorgan:
Been using #19 for a while with no issues and both that one and #20 work, so I'm merging this in

PR #18, opened 2025-05-28
feat: workspace folders

### LINKED ISSUE #136: Workspace Folders

doorgan:
This adds support for multiple projects within the server node, and starts a project node for each detected project within the workspace. The detection implemented here is just using the presence of a `mix.exs` file to assume it's a project root.

It also changes the manager node to use the name of the workspace folder instead of the name of the first project that starts distribution. The reason is that if you have projects `a` and `b` in the same root folder `foo`, and you start Expert on the root folder, then the manager node will be named after the first project node that is started. Running `epmd -names` shows this:

- `project-a-entropy`
- `project-b-entropy`
- `manager-a-entropy`

I htink having the manager be named `manager-foo-entropy` sounds more correct.

To test this branch you can clone this repo: https://github.com/doorgan/monorepo_test
Then check on files from both `main` and `secondary` that:
- Completions work
- Go to definition and workspace symbols work
- Code actions (like Refactorex refactors) work
- Instructing the LSP to reindex everything runs the indexer on all projects

Additionally you can also test that Lexical works on `.exs` or `.ex` files in a workspace without a `mix.exs` file.

It is worth noting that this should be tested with VS Code using the Lexical LSP extension and pointing it to an expert release using this branch, or some other editor (like emacs) that doesn't automatically start a new Expert server(ie something other than Neovim)

Closes #136

### DIFF
--- a/apps/expert/lib/expert/active_projects.ex
+++ b/apps/expert/lib/expert/active_projects.ex
@@ -35,6 +35,14 @@ defmodule Expert.ActiveProjects do
     |> Enum.map(fn {_, project} -> project end)
   end
 
+  @spec find_by_root_uri(Forge.uri()) :: Project.t() | nil
+  def find_by_root_uri(root_uri) do
+    case :ets.lookup(__MODULE__, root_uri) do
+      [{_, project}] -> project
+      [] -> nil
+    end
+  end
+
   def add_projects(new_projects) when is_list(new_projects) do
     for new_project <- new_projects do
       # We use `:ets.insert_new/2` to avoid overwriting the cached project's entropy
--- a/apps/expert/lib/expert/state.ex
+++ b/apps/expert/lib/expert/state.ex
@@ -169,8 +169,8 @@ defmodule Expert.State do
     } = did_open.params.text_document
 
     project =
-      with nil <- Enum.find(ActiveProjects.projects(), &Project.within_project?(&1, uri)) do
-        Project.find_project(uri)
+      with %Project{} = closest <- Project.find_project(uri) do
+        ActiveProjects.find_by_root_uri(closest.root_uri) || closest
       end
 
     if project do
--- a/apps/expert/test/expert/expert_test.exs
+++ b/apps/expert/test/expert/expert_test.exs
@@ -28,7 +28,27 @@ defmodule ExpertTest do
       |> Document.Path.to_uri()
       |> Project.new()
 
-    [project_root: project_root, main_project: main_project, secondary_project: secondary_project]
+    nested_root_path = fixtures_path() |> Path.join("nested_projects")
+
+    nested_root_project =
+      nested_root_path
+      |> Document.Path.to_uri()
+      |> Project.new()
+
+    nested_subproject =
+      nested_root_path
+      |> Path.join("subproject")
+      |> Document.Path.to_uri()
+      |> Project.new()
+
+    [
+      project_root: project_root,
+      main_project: main_project,
+      secondary_project: secondary_project,
+      nested_root_path: nested_root_path,
+      nested_root_project: nested_root_project,
+      nested_subproject: nested_subproject
+    ]
   end
 
   setup do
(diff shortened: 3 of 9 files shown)


apps/common/lib/lexical/workspace.ex
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def set_workspace(workspace) do
+    :persistent_term.put({__MODULE__, :workspace}, workspace)
+  end
+
+  def get_workspace do
+    :persistent_term.get({__MODULE__, :workspace}, nil)
+  end
+end
```
doorgan: There were two motivations for me to add this:
- Because we can have multiple projects which won't necessarily be at the workspace root, having a concept for that root is useful to formalize it
- I needed to know the workspace root folder name to be able to name the manager node after it, and storing it in persistent term(I don't see this value changing often, if at all) like we do for the project in the remote control node felt more straightforward than some convoluted refactor to pipe this value all the way into `RemoteControl.start_link`

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -119,7 +119,15 @@ defmodule Lexical.RemoteControl do
   end
 
   def manager_node_name(%Project{} = project) do
-    :"manager-#{Project.name(project)}-#{Project.entropy(project)}@127.0.0.1"
+    workspace = Lexical.Workspace.get_workspace()
+
+    workspace_name =
+      case workspace do
+        nil -> Project.name(project)
```
doorgan: I've seen this code run before the workspace was stored in `:persistent_term`, hence this clause. I want to remove it though.

apps/expert/lib/expert/configuration.ex
```diff
@@ -29,16 +26,11 @@ defmodule Expert.Configuration do
 
   @dialyzer {:nowarn_function, set_dialyzer_enabled: 2}
 
-  @spec new([Structures.WorkspaceFolder.t()], map(), String.t() | nil) :: t
-  def new(workspace_folders, %Structures.ClientCapabilities{} = client_capabilities, client_name) do
+  @spec new(map(), String.t() | nil) :: t
```
mhanberg: They all have typescpecs, so this can be `Structures.ClientCapabilities.t()`

apps/expert/lib/expert/state.ex
```diff
@@ -174,10 +183,26 @@ defmodule Expert.State do
...
+    config = state.configuration
+
+    project =
+      case Enum.find(ActiveProjects.projects(), &Project.within_project?(&1, uri)) do
+        nil ->
+          Project.find_project(uri)
+
+        project ->
+          project
+      end
```
mhanberg: nit: this is a case for a with

```elixir
with nil <- Enum.find(ActiveProjects.projects(), &Project.within_project?(&1, uri)) do
  Project.find_project(uri)
end
```

You could also just use the third argument to Enum.find/3, but that will evaluate it immediately.

apps/expert/lib/expert/state.ex
```diff
@@ -174,10 +183,26 @@ defmodule Expert.State do
...
+
+    if project do
+      ensure_project_node_started(project)
+      ActiveProjects.add_projects([project])
+    end
+
     case Document.Store.open(uri, text, version, language_id) do
       :ok ->
         Logger.info("################### opened #{uri}")
```
mhanberg: this looks leftover from when i was getting genlsp integrated, this can get removed.

apps/expert/lib/expert/state.ex
```diff
@@ -96,10 +116,52 @@ defmodule Expert.State do
...
+      }) do
+    workspace_folders = Enum.uniq((added ++ state.workspace_folders) -- removed)
+
+    removed_projects =
+      for %{uri: uri} <- removed do
+        project = Project.new(uri)
+
+        Expert.Project.Supervisor.stop(project)
+
+        GenLSP.notify(
```
mhanberg: There are convenience functions for logging 

https://hexdocs.pm/gen_lsp/GenLSP.html#log/2

doorgan:
@mhanberg using GenLSP test helpers sometimes exunit will not terminate after running all tests(or sometimes it will run only a subset of the new tests), do you know if this may be related to the GenLSP helpers? Other tests that start project nodes like `apps/expert/test/expert/engine_node_test.exs` don't have this problem, so it seems unlikely the project node is what's causing this

mhanberg:
> @mhanberg using GenLSP test helpers sometimes exunit will not terminate after running all tests(or sometimes it will run only a subset of the new tests), do you know if this may be related to the GenLSP helpers? Other tests that start project nodes like `apps/expert/test/expert/engine_node_test.exs` don't have this problem, so it seems unlikely the project node is what's causing this

Not that I know of, but I'll pull it down and see what happens.

mhanberg:
> > @mhanberg using GenLSP test helpers sometimes exunit will not terminate after running all tests(or sometimes it will run only a subset of the new tests), do you know if this may be related to the GenLSP helpers? Other tests that start project nodes like `apps/expert/test/expert/engine_node_test.exs` don't have this problem, so it seems unlikely the project node is what's causing this
> 
> Not that I know of, but I'll pull it down and see what happens.

Sorry about the wait, but I pulled this down. The tests seem to just run as normal, except for one test failing of a code error.

doorgan:
> except for one test failing of a code error.

Can you share the error? Besides the issue I mentioned, all tests are green on my end

mhanberg:
> > except for one test failing of a code error.
> 
> Can you share the error? Besides the issue I mentioned, all tests are green on my end

The first couple of times i ran it, this was the failure

```
  1) test find references returns locations that the entity returns (Expert.Provider.Handlers.FindReferencesTest)
     test/expert/provider/handlers/find_references_test.exs:55
     ** (FunctionClauseError) no function clause matching in anonymous fn/4 in Expert.Provider.Handlers.FindReferencesTest."test find references returns locations that the entity returns"/1


     code: assert {:ok, [%Location{} = location]} = handle(request, project)
     stacktrace:
       (patch 0.15.0) lib/patch/mock/server.ex:81: Patch.Mock.Server.delegate/3
       (expert 0.7.2) lib/expert/provider/handlers/find_references.ex:25: Expert.Provider.Handlers.FindReferences.handle/2
       test/expert/provider/handlers/find_references_test.exs:72: (test)
```

But I just ran it again and a different test failed

```
  1) test opening files starts a project node when opening a file in a folder not specified as workspace folder (ExpertTest)
     test/expert/expert_test.exs:216
     Assertion with == failed
     code:  assert project.root_uri == secondary_project.root_uri
     left:  "file:///home/mitchell/src/expert/apps/forge/test/fixtures/workspace_folders/main"
     right: "file:///home/mitchell/src/expert/apps/forge/test/fixtures/workspace_folders/secondary"
     stacktrace:
       test/expert/expert_test.exs:263: (test)
```

mhanberg:
I think I just ran into the problem tho

mhanberg:
Something is definitely strange. Sometimes I'll run the test file and It'll only run a single test.
image

doorgan:
I think I fixed both issues. I'm still experiencing the issue where `mix test` won't always terminate after all tests run, but I'm not sure if something is wrong in my setup.

mhanberg:
Are you still experiencing that weird tests don't end problem? CI seems to be passing.

doorgan:
@mhanberg I'm not so I think we're good, I'm holding off a bit because @scohen seemed to be having some issues with this branch; I'm not yet sure if they were introduced by this pr or if they're preexisting

mhanberg:
> @mhanberg I'm not so I think we're good, I'm holding off a bit because @scohen seemed to be having some issues with this branch; I'm not yet sure if they were introduced by this pr or if they're preexisting

It might be worth it to move the "multi root but not workspace folders" work to a different branch and merge in the actual workspace folders support

doorgan:
@mhanberg I'm not sure, Steve reported that the project node would sometimes crash and that would happen regardless of the project node being spawned via workspace folders or some other way
I haven't been able to reproduce that issue though, not even after I installed Emacs on my machine

apps/expert/lib/expert/active_projects.ex
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def start_link do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  def init(_) do
+    __MODULE__ = :ets.new(__MODULE__, [:set, :named_table, :public, read_concurrency: true])
+    {:ok, nil}
+  end
```
heywhy, line 25: I think wrapping initialization of the ets table in a `Task` should be sufficient since it's owned by no process instead of using a `GenServer` whereby the process would be left alive doing nothing. Or the `init/1` method should return `:ignore` so that the Supervisor does not keep it running, but I'm keen on a `Task` added to the supervision tree instead.

heywhy: Please ignore my comment. 😅

doorgan:
I rebased this but with all the changes since August I need to rethink the way we indicate the server is ready to the client. On the fly engine compilation delays how quickly the engine gets ready so if you are in subproject A and open a file in subproject B, you'll see a lot of `:no_connection` errors until the engine gets started, which is quite jarring

mhanberg:
It might help other issues as well (or at least stop red herrings/noise from real issues) but in Next LS, if the runtime (aka engine) wasn't ready yet, it either returned an LSP error with a message or just returned nil, depending on the request. 

I just had the runtime send a ready message to the server and the server kept track of it. 

I can't remember how expert currently does this sort of thing.

doorgan:
Yeah that seems to help, the last commit implements that and testing with the WIP expert vscode extension it seems to work on the `monorepo_test` I linked in the OP

apps/expert/lib/expert/state.ex
```diff
@@ -49,16 +50,31 @@ defmodule Expert.State do
...
-    Task.Supervisor.start_child(:expert_task_queue, fn ->
-      {:ok, _pid} = Project.Supervisor.start(config.project)
-      send(Expert, :engine_initialized)
-    end)
+    projects =
+      for %{uri: uri} <- event.workspace_folders || [],
+          project = Project.new(uri),
+          project.mix_project? do
+        project
+      end
```
Moosieus: Could the `:mix_project?` property be done away with at this rate? Rationale being, we assume every `%Project{}` struct is a mix project. More guarantees downstream and whatnot.

Moosieus:
https://github.com/elixir-lang/expert/blob/1a6a12eb429cc54f5e0b203256ac5bb8987b9309/apps/expert/lib/expert/configuration.ex#L42-L45

This seemed foreboding and worthy of mention

mhanberg:
@doorgan lets get this rebased and then address Cameron's comments, probably time to merge if it seems stable to you.

apps/expert/lib/expert/state.ex
doorgan: I'm not sure we want to remove that entirely, wouldn't it close the door for expert working on bare `.ex` or `.exs` files, even if it doesn't fully support that today?

doorgan:
This isn't stable anymore for me but I can't pinpoint why. I started getting these errors after a computer restart:
```
14:07:49.480 [warning] 'global' at :"expert-project-main-55714@127.0.0.1" failed to connect to :"expert-project-secondary-17485@127.0.0.1"
secondary: 
14:07:49.480 [warning] 'global' at :"expert-project-secondary-17485@127.0.0.1" failed to connect to :"expert-project-main-55714@127.0.0.1"


14:07:49.481 [warning] 'global' at node :"expert-project-secondary-17485@127.0.0.1" requested disconnect from node :"expert-project-main-55714@127.0.0.1" in order to prevent overlapping partitions
```
Then it starts an endless loop of trying to start either project node and nothing works anymore. It still works if I'm on a single root project, the problem happens when there's more than 1 project node.

Also FWIW I don't think we want project nodes connecting to each other.

doorgan:
> https://github.com/elixir-lang/expert/blob/1a6a12eb429cc54f5e0b203256ac5bb8987b9309/apps/expert/lib/expert/configuration.ex#L42-L45
> 
> This seemed foreboding and worthy of mention

This comment was stale, so I removed it

mhanberg:
I'll take a look.

apps/expert/lib/expert/state.ex
Moosieus: This is a bear to broach b/c so much of Expert assumes a Mix project today... Gonna think aloud here a bit.

Bare files include but aren't limited to:
* Compile-able `.ex` files + `Code.require_file` invocations.
* `.exs` scripts.
* `.livemd` Livebooks.
* Code blocks in `*.md` files or otherwise.

AFAIK, the only way to have 'an Elixir project' is:
* Mix (what we support)
* Distillery (succeeded by Mix)
* Roll-your-own (no planned support)

Beyond that, if you 'just have Elixir files', it's down to imperative `Code.require_file` calls or etc... AFAIK at least - I've never been struck by the fey mood to try in earnest. In conclusion, **we support no case today where you're in an Elixir project, but not a Mix project.**

I'd go so far to say in a scenario like this:
```
my_elixir_project
├── foo.ex
└── bar.ex
```

I'd puritanically consider `foo.ex` and `bar.ex` **not to be in a project together**. They have no relation to each other unless `foo.ex` invokes `Code.require_file("bar.ex", __DIR__)`. Even then that's an imperative invocation to unwrap that could reach anywhere in the file system, there's still no practical or useful project scope.

The next question top of mind, is it a practical tradeoff to keep a `%Project{}` struct in scope and handle non-mix requests conditionally, or should we handle non-mix requests from a top-level sans any `%Project{}` struct in scope?

Today the former's less friction with the existing code, but it'd overall be cleaner to disentangle at the top-level, throw back to Rich Hickey's "Maybe Not" talk and all that.

... What was I talking about again? Oh yeah - I don't think removing `:mix_project?` from the Project struct closes the door on supporting bare files, they should just be handled sans Project struct.

That said, I'm fine with leaving it be for now in this PR, talked myself out of it 😛

doorgan: Thinking more about this I think that `mix_exs_path` already covers the `mix_project?` flag anyways, right? And that already drives most of the logic around what to do for a mix project

Another thing is that I think `Project` represents a root path and a set of elixir/otp versions more than a "Project" in the mix sense. It's whatever determines what needs an Engine node to be spawned. It's probably due for a rename.

doorgan:
I think I managed to fix the issue I was having. We were registering the nodes using the longname, but then trying to retrieve their ports using the shortname. Apparently it's not an issue with a single engine node because it's directly connected to the manager, but extra nodes makes `:global` try to fully connect the node mesh, then fails to find the port for the first engine node and kicks the second engine node from the mesh due to that failure to connect.

PR #19, opened 2025-06-04
Use GenLSP data structures

doorgan:
This change replaces the use of `Lexical.Protocol` data structures for the `GenLSP` counterparts, and handles conversion between those and the internal data structures for ranges, positions and locations. Since neither `Proto` nor `Protocol` are really needed after this change, I also removed them altogether and consolidated the few modules we do need for them into the `Forge` application.

This PR needs GenLSP support for cancellation before it can be reviewed/merged

Besides cleanups in general and removal of dead code, currently there are a few issues I'm working on:
- ~~On neovim, completions for module segments after the `.` not always show, but they are recognized once you start typing after the dot~~ fixed
- ~~On VSCode, code actions are not being recognized(but they are and work properly on neovim)~~ fixed
- ~~On neovim, workspace symbols don't show up but they do show up in VSCode~~ happens on main too
- On tests, diagnostics aren't being cleaned up after a document is marked non-dirty

Other than that, expert using this pr is functional

### DIFF
--- a/apps/engine/lib/engine/engine/code_action.ex
+++ b/apps/engine/lib/engine/engine/code_action.ex
@@ -9,18 +9,9 @@ defmodule Engine.CodeAction do
 
   defstruct [:title, :kind, :changes, :uri]
 
-  @type code_action_kind ::
-          :empty
-          | :quick_fix
-          | :refactor
-          | :refactor_extract
-          | :refactor_inline
-          | :refactor_rewrite
-          | :source
-          | :source_organize_imports
-          | :source_fix_all
+  @type code_action_kind :: GenLSP.Enumerations.CodeActionKind.t()
 
-  @type trigger_kind :: :invoked | :automatic
+  @type trigger_kind :: GenLSP.Enumerations.CodeActionTriggerKind.t()
 
   @type t :: %__MODULE__{
           title: String.t(),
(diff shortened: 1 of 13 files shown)


mhanberg:
You're a beast

apps/expert/lib/expert/provider/handlers/document_symbols.ex
```diff
@@ -31,28 +31,28 @@ defmodule Expert.Provider.Handlers.DocumentSymbols do
...
       end
 
-    Symbol.new(
+    %Structures.DocumentSymbol{
       children: children,
       detail: root.detail,
       kind: to_kind(root.type),
       name: root.name,
       range: root.range,
       selection_range: root.detail_range
```
mhanberg, line 40: @doorgan looks like this is still a Forge range, needs to be converted to a GenLSP.Structures.Range

doorgan: It is converted here:
https://github.com/elixir-lang/expert/blob/1b42e49e92640d8d6155627d02d307f3b85f9884/apps/expert/lib/expert/transport/std_io.ex#L25-L107

Conversely, they are converted from genlsp to forge here: https://github.com/elixir-lang/expert/blob/1b42e49e92640d8d6155627d02d307f3b85f9884/apps/expert/lib/expert.ex#L129-L130

I think we'd need to do what I'm doing in `std_io` somewhere in `expert.ex` when using genlsp

mhanberg: I was about to comment back, I realized I deleted the code that did the conversion. 

I'll get it converted again

doorgan:
Been using this branch for a while with no issues, so I'm merging it

PR #20, opened 2025-06-17
refactor: integrate GenLSP

mhanberg:
## Description

Integrate GenLSP server process.

## Notes

- Still working through this, but having trouble with a couple tests and a potential problem with namespacing, hoping @doorgan can help me through these last bits.
- There are some extraneous changes in here that I'll revert before merging. Some credo rules, credo github package, etc
- New GenLSP will get published to hex before merging.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -133,7 +133,7 @@ jobs:
 
       - name: Run dialyzer
         run: |
-          mix compile.protocols
+          make compile.protocols.poncho
           make dialyzer.all
 
   test:


apps/engine/benchmarks/ets_bench.exs
```diff
@@ -33,7 +33,7 @@ defmodule BenchHelper do
 end
 
 cwd = __DIR__
-project = Project.new("file://#{cwd}")
+project = Project.new(%GenLSP.LSP{mod: :foo}, "file://#{cwd}")
```
mhanberg: There are some dummy lsp structs in places that use this function in tests. Putting the LSP struct in that Project struct might not be the right place for it.

apps/expert/lib/expert/state.ex
```diff
@@ -37,10 +34,14 @@ defmodule Expert.State do
     %__MODULE__{}
   end
 
-  def initialize(%__MODULE__{initialized?: false} = state, %Requests.Initialize{
-        id: event_id,
-        params: %Structures.InitializeParams{} = event
-      }) do
+  # TODO: this function has a side effect (starting the project supervisor)
+  # that i think might be better off in the calling function
```
doorgan, line 38: I'm changing that in https://github.com/elixir-lang/expert/pull/18

PR #22, opened 2025-06-27
refactor: use Burrito for packaging

doorgan:
Updates Expert to use Burrito for packaging

This builds both Expert and Engine with the same elixir and otp versions, the idea is that next we will build several Engine versions in CI and Expert will download the right one for the current project, or default to the one included in it.

To build a release for your machine:
```sh
make release.local
```

To build releases for macos, linux and windows:
```sh
make release
```

To only build Engine(this will generate namespaced files at `apps/engine/_build/dev_ns`:
```sh
make build.engine
```

Note that Engine will be built for `dev` and not `prod`, this is because it makes use of `Mix.Project` and related code, which would not be available if it were build with `prod`. The idea behind this command is that we will have a CI step on release that will run `make build.engine` against a matrix of supported elixir+otp versions and publish those artifacts, which Expert will then download so it matches the exact project versions.

---

There are some other relatively minor differences or things to note:
- Due to the way Burrito generates the executable, we lose the ability to include a `debug_shell.sh` script in it, as far as I can see. We need to rethink how to do this, we can still connect an observer to the server node with `:observer.start(:"manager-project-entropy@127.0.0.1")` but I'm not sure how to open a remote session like one would with a regular release script.
- A few tests in Engine were calling code that in production would run in the Expert node(like `Engine.Api`). Because that code was moved to Expert and is no longer accessible by Engine, those tests were moved to Expert to avoid circular dependencies on tests.


mhanberg:
To recap the trouble I have been having testing this locally on x86_64 linux:

- Zig was failing to compile the burrito wrapper, I thought this was due to my Linux kernel being too knew. I had recently updated my desktop to Ubuntu 25.04.
- I re-install my operating system with NixOS, which uses a LTS version of the linux kernel. Error still occurs. I think maybe its Nix being weird (I was using Nix on my Ubuntu install)
- I try a plain jane x86_64 Ubuntu 24.04 virtual machine, still the same error.

I am left with confusion why it passed in CI, but can't seem to pass for me locally. Digit is going to investigate, and I am going to invite him to the repository so it's easier for him.

mhanberg:
> To recap the trouble I have been having testing this locally on x86_64 linux:
> 
>     * Zig was failing to compile the burrito wrapper, I thought this was due to my Linux kernel being too knew. I had recently updated my desktop to Ubuntu 25.04.
> 
>     * I re-install my operating system with NixOS, which uses a LTS version of the linux kernel. Error still occurs. I think maybe its Nix being weird (I was using Nix on my Ubuntu install)
> 
>     * I try a plain jane x86_64 Ubuntu 24.04 virtual machine, still the same error.
> 
> 
> I am left with confusion why it passed in CI, but can't seem to pass for me locally. Digit is going to investigate, and I am going to invite him to the repository so it's easier for him.

Digit was able to get it to wrap in a VM, so I'm just going to classify this as I'm cursed and I'll figure it out later

PR #23, opened 2025-06-29
refactor: isolate Engine from Expert

doorgan:
The burrito packaging will require the Engine app to be fully isolated from Expert, this PR does that

- In order to build the Engine app with one elixir version and have Expert load it for the project node, Expert must not depend on Engine(other than for tests)
- A lot of the code in Engine was actually being setup and used by Expert, so a lot of code had to be moved either to Expert, or to Forge
- Both Expert and Engine used shared data structures like `%Diagnostics{}` that were defined in the Engine application. Those datastructures were moved to `Forge` to remove the dependency of Expert on Engine
- We need to perform two namespacing tasks for two different contexts: once for the Engine app so it can be packaged, and once for Expert so the shared data structures get namespaced. Otherwise Expert would try to use `Forge.Project` and Engine would try to use `XPForge.Project`, causing crashes.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -44,7 +44,7 @@ jobs:
       # Step: Define how to cache deps. Restores existing cache if present.
       - name: Cache deps
         id: cache-deps
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: |
             apps/**/deps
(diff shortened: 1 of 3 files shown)


apps/expert/lib/expert/release.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Expert.Release do
```
mhanberg: this needs to go in #22

PR #24, opened 2025-06-30
feat: on the fly engine builds

doorgan:
Ideally, the Project node would use an Engine built with the same elixir and otp versions as the project.

This PR adds support for that, looking at the user's data folder(`AppData` in windows, `.local/share` in linux, `Library/Application Support` on macos, etc) for an Engine build matching the project versions, and building an engine matching the user's elixir+otp on the fly if none is found

### DIFF
--- a/apps/expert/priv/build_engine.exs
+++ b/apps/expert/priv/build_engine.exs
@@ -13,36 +13,27 @@
 expert_vsn = Keyword.fetch!(args, :vsn)
 engine_source_path = Keyword.fetch!(args, :source_path)
 
-major = :otp_release |> :erlang.system_info() |> List.to_string()
-version_file = Path.join([:code.root_dir(), "releases", major, "OTP_VERSION"])
+expert_data_path = :filename.basedir(:user_data, "Expert", %{version: expert_vsn})
 
-erlang_vsn =
-  try do
-    {:ok, contents} = File.read(version_file)
-    String.split(contents, "\n", trim: true)
-  else
-    [full] -> full
-    _ -> major
-  catch
-    :error ->
-      major
-  end
+System.put_env("MIX_INSTALL_DIR", expert_data_path)
 
-elixir_vsn = System.version()
+Mix.Task.run("local.hex", ["--force"])
+Mix.Task.run("local.rebar", ["--force"])
 
-user_data_path = :filename.basedir(:user_data, "Expert", %{version: expert_vsn})
-out_path = Path.join([user_data_path, "engine-#{elixir_vsn}-otp-#{erlang_vsn}"])
-build_path = Path.join(out_path, "dev")
+Mix.install([{:engine, path: engine_source_path, env: :dev}],
+  start_applications: false,
+  config_path: Path.join(engine_source_path, "config/config.exs"),
+  lockfile: Path.join(engine_source_path, "mix.lock")
+)
 
-if not File.exists?(build_path) do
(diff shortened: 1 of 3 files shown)


apps/expert/priv/build_engine.exs
```diff
@@ -0,0 +1,39 @@
...
+out_path = Path.join([out_path, "engine-#{elixir_vsn}-otp-#{erlang_vsn}"])
+dev_out_path = Path.join(out_path, "dev")
+
+if not File.exists?(dev_out_path) do
+  System.put_env("MIX_INSTALL_DIR", out_path)
+
+  Mix.Task.run("local.hex", ["--force"])
+  Mix.Task.run("local.rebar", ["--force"])
+  Mix.Project.in_project(:engine, engine_source_path, [build_path: out_path], fn _module ->
+    Mix.Task.run("deps.get", [])
```
mhanberg: It might be worth downloading the deps at release time and bundling them along with the source code. 

I'm mostly thinking about once less network call and mix task that could fail. 

Something to consider for the future.

apps/forge/lib/forge/ast.ex
```diff
@@ -292,7 +292,7 @@ defmodule Forge.Ast do
     case do_container_cursor_to_quoted(document_fragment) do
       {:ok, quoted} ->
         quoted
-        |> Future.Macro.path(&match?({:__cursor__, _, _}, &1))
+        |> Forge.Future.Macro.path(&match?({:__cursor__, _, _}, &1))
```
mhanberg: Were none of these working before? 😅

apps/forge/lib/forge/future/code.ex
```diff
@@ -1,7 +1,7 @@
 # Copied from https://github.com/elixir-lang/elixir/blob/bacea2cef6323d0ede4222f36ddcedd82cb514e4/lib/elixir/lib/code.ex
 # I commented print_diagnostic/1, because it is not used in the codebase.
 # And I changed string_to_quoted/2 to make it use :future_elixir
-defmodule Future.Code do
+defmodule Forge.Future.Code do
```
mhanberg: Oh, disregard my other comment.

mhanberg:
Sorry, I didn't realize this was a draft.

apps/forge/lib/forge/ast.ex
doorgan: They did but they complicate namespacing!

doorgan:
@mhanberg yeah, it's functional but one bit I need to figure out first is how to make it viable for development, otherwise we need to rebuild the engine every single time we launch expert in case the engine source changed

doorgan:
@mhanberg this is now ready for review, I moved to `Mix.install` so my previous concern is now resolved, we're still namespacing on every launch but it's not that bad compared to fully recompiling everything every time, and since namespacing isn't incremental I'm not sure if there's a good way to avoid this

mhanberg:
I'll try this out before bed tonight.

mhanberg:
Seems to work, tested with a local release on darwin arm (I still can't get the burrito to work on my linux machine, but I think its from a strange zig compiler memory corruption... the people in the zig discord were extremely confused. still digging into it)

One thing is we should send a lsp progress when the engine is compiling. also in general, should be sending more logs through the logMessage notification so that they show up in the output panel/lsp logs inside the editors. 

i would merge this now tho

doorgan:
I'll add the lsp progress on a followup pr

PR #25, opened 2025-07-03
ci: conventional commits

mhanberg:
Since we are going to use release-please to manage our releases and
changelog, we need to start using conventional commits and the Squash
and Merge style of PR merging.

This workflow will lint the pull request title to ensure that it follows
conventional commits, since the title will be used as the commit subject
when squash and merging


mhanberg:
Since this uses pull_request_target, I believe it won't actually run until its is merged and lives on main.

PR #37, opened 2025-07-29
chore: update release workflow

doorgan:
We're running out of CI minutes in this org so I tested this workflow on a private org with no actions limits
This is largely based on NextLS release workflow: https://github.com/elixir-tools/next-ls/blob/main/.github/workflows/release.yaml

Screenshot 2025-07-29 at 6 17 46 PM


.github/workflows/release.yml
```diff
@@ -1,55 +1,72 @@
...
+    steps:
+      - run: gh release edit ${{ needs.release.outputs.tag_name }} --draft=true --repo='elixir-lang/expert'
+
+  build:
+    needs: [release, draft]
+    runs-on: macos-14
+    if: ${{ needs.release.outputs.release_created }}
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      NEXTLS_RELEASE_MODE: "burrito"
```
mhanberg: Expert?

doorgan: 🤫

VERSION
```diff
@@ -0,0 +1 @@
+0.1.0
```
mhanberg: Good call on this.

I think there might have to be something done to the release please config in order for it to bump this file.

Also, we probably don't need to set the release-type to elixir anymore, maybe we just need to set it to "simple", but I think that also might bump a "version.txt" file.

I use that method in elixir-tools.nvim [here](https://github.com/elixir-tools/elixir-tools.nvim/blob/main/release-please-config.json#L6)

doorgan: Oh I think we can definitely follow that route and use the `version.txt` instead, it's the same idea

doorgan: Noice, that works as expected
Screenshot 2025-07-30 at 11 51 48 PM
Screenshot 2025-07-30 at 11 51 59 PM

PR #40, opened 2025-08-07
chore: silence spammy log messages

scohen:
For a while now, elixir_sense has been emitting lots of log messages that spammed our logs and made it harder to see what was going on. This PR removes the spam.

Now, we remove any log message that has "is already compiled." in it, which handles all of the messages caused by elixir_sense.

### DIFF
--- a/apps/engine/lib/engine/engine/bootstrap.ex
+++ b/apps/engine/lib/engine/engine/bootstrap.ex
@@ -67,7 +67,7 @@ defmodule Engine.Bootstrap do
     }
 
     :logger.add_handler(handler_name, :logger_std_h, config)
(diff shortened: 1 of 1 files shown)


apps/engine/lib/engine/engine/bootstrap.ex
```diff
@@ -66,6 +67,7 @@ defmodule Engine.Bootstrap do
     }
 
     :logger.add_handler(handler_name, :logger_std_h, config)
+    LogFilter.hook_into_logger(handler_name)
```
doorgan:
```suggestion
    LogFilter.hook_into_logger()
```

PR #42, opened 2025-08-09
chore: bump burrito

mhanberg:
Bumping burrito to v1.4 supports a new URL scheme for the beam machine,
which is the CDN that burrito pulls OTP builds from from darwin and
linux

### DIFF
--- a/justfile
+++ b/justfile
@@ -93,5 +93,9 @@ release-plain: (compile "engine")
     cd apps/expert
     MIX_ENV=prod mix release plain --overwrite
 
+[doc('Compiles .github/matrix.json')]
+compile-ci-matrix:
+  elixir matrix.exs
+
 default: release-local
 
--- a/matrix.exs
+++ b/matrix.exs
@@ -7,7 +7,7 @@ projects = [
   "forge"
 ]
 
-# renable these once we make the repo public
+# TODO(#44): renable these once we make the repo public
 versions = [
   # %{elixir: "1.18.1", otp: "27"},
   # %{elixir: "1.18.1", otp: "26"},


mhanberg:
I realized that CI was not correctly linting, etc. So i did a number of things

- fixed some dependency things. we were setting the env of the path depencencies, which was causing some test only deps to not get installed
- after fixing the above, there were some forge test utils that had to get moved into lib
- there were some unused deps in the lockfiles, removed those with mix deps.unlock --unused
- changed the ci workflow to use just, and correctly runs the right commands
- changes the workflow to run each lint/test for each project separately. I think will help narrow down when tests are failing, which project they're coming from.
- this also fixes some deps caching issues i noticed
- it does create a gillion jobs, but not really a big deal.
- moves the matrix to a json file, which is created via an elixir script
- updated the justfile to make it easy to run anything in any project `just mix forge deps.get` `just mix expert credo --strict` etc
    - I am wondering if trying out something like https://github.com/sportradar/elixir-workspace is worth trying out

There is still a dialyzer issue that is now being reported since the test utils are now in lib, will fix those tomorrow.

justfile
```diff
@@ -17,56 +17,51 @@ run project +ARGS:
...
+    #!/usr/bin/env bash
+    set -euxo pipefail
+
+    cd apps/engine
+    MIX_ENV=dev mix compile
+    namespaced_dir=_build/dev_ns
+    rm -rf $namespaced_dir
+    mkdir -p $namespaced_dir
+
+    cp -r _build/dev "$namespaced_dir"
```
doorgan:
```suggestion
    cp -r _build/dev/ "$namespaced_dir"
```
The missing `/` causes the directory itself to be copied, not its content, producing `_build/dev_ns/dev` which breaks namespacing and breaks the engine node

mhanberg: Good catch, I always mess that up.

PR #43, opened 2025-08-14
chore: support dev server using tcp transport

doorgan:
Adds the `just start` command to start expert listening on port `9000` for development

I had to bring back some changes to namespacing from #1 to make this work; we want to namespace the engine and the data structures it shares with Expert, but not anything else as it makes working with iex harder. Also because elixir will try to read non-namespaced config files(ie `runtime.exs`), so if the expert app is namespaced too, Mix will complain about a non-existent `:expert` app being configured(same with `:snowflake` and any other dependency that we configure)

### DIFF
--- a/apps/forge/lib/forge/namespace/transform/beams.ex
+++ b/apps/forge/lib/forge/namespace/transform/beams.ex
@@ -22,7 +22,8 @@ defmodule Forge.Namespace.Transform.Beams do
         fn beam ->
           apply_and_update_progress(beam, me, opts)
         end,
-        timeout: 10_000
+        ordered: false,
+        timeout: :infinity
       )
       |> Stream.run()
     end)


apps/expert/lib/expert/engine_node.ex
```diff
@@ -166,11 +166,19 @@ defmodule Expert.EngineNode do
     # When Engine is built in CI for a version matrix, we'll need to check if
     # we have the right version downloaded, and if not, we should download it.
     defp glob_paths do
-      :expert
-      |> :code.priv_dir()
+      engine_path = System.get_env("EXPERT_ENGINE_PATH", default_engine_path())
+
+      IO.puts("Using Engine path: #{Path.expand(engine_path)}")
```
mhanberg: Probably want to use the logger here

doorgan: Yeah, it's a debug leftover though 🙈

PR #52, opened 2025-08-28
docs: add project versions docs to architecture.md

doorgan:


mhanberg:
it really doesnt matter, but for docs changes you should use the `docs` type instead of the `chore` type

PR #53, opened 2025-08-28
update Lexical -> Expert in install instructions

andyleclair:
Noticed a few places in the install docs where `Lexical` should be `Expert`

### DIFF
--- a/pages/installation.md
+++ b/pages/installation.md
@@ -91,7 +91,7 @@ emacs configuration), insert the following code:
   (lsp-elixir-server-command '("/my/home/projects/expert/apps/expert/burrito_out/expert_linux_amd64")))
 ```
 
-Restart emacs, and Lexical should start when you open a file with a
+Restart emacs, and Expert should start when you open a file with a
 `.ex` extension.
 
 
@@ -111,7 +111,7 @@ You can add Expert support in the following way:
            '("expert_linux_amd64" "start_lexical.sh")))))
 ```
 
-For versions before 30, you can add Eglot support for Lexical in the
+For versions before 30, you can add Eglot support for Expert in the
 following way:
 
 ```emacs-lisp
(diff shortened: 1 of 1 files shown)


pages/installation.md
```diff
@@ -142,9 +142,9 @@ for Eglot:
 
 Click on the extensions button on the sidebar, then search for
 `lexical`, then click `install`.  By default, the extension will automatically
-download the latest version of Lexical.
+download the latest version of Expert.
 
-To change to a local executable, go to `Settings -> Extensions -> Lexical` and
+To change to a local executable, go to `Settings -> Extensions -> Expert` and
```
JohnDoneth: I believe these ones are supposed to be Lexical still since the Lexical extension is being used.

andyleclair: I've noticed that getting it running locally. In neovim, if I set `lspconfig.expert.setup { ...` it wouldn't work, but `lspconfig.lexical.setup` did. I'm happy to help out with any other renaming stuff

andyleclair:
Should we tell people to put the release on their `PATH` (like in `~/.local/bin/expert`)? If we did that, we could set the `cmd` in `nvim-lspconfig` and it should just work?

pages/installation.md
doorgan: For the VSCode extension, we should keep Lexical here since that's the extension that is compatible with Expert at the moment, we need a new Expert vscode extension

On Neovim this is my configuration: https://github.com/doorgan/dotfiles/blob/main/dot_config/nvim/lua/plugins/lsp-config.lua#L26-L55 we should update the docs to be more inline with this

andyleclair: From the `nvim-lspconfigs` github, it appears that `require('lspconfig')` is deprecated now? https://github.com/andyleclair/nvim-lspconfig?tab=readme-ov-file#important-%EF%B8%8F

IMO, if that's what _they're_ saying, it's what _we_ should say?

doorgan:
> Should we tell people to put the release on their `PATH` (like in `~/.local/bin/expert`)? If we did that, we could set the `cmd` in `nvim-lspconfig` and it should just work?

That sounds useful, we can add that!

pages/installation.md
doorgan: > IMO, if that's what they're saying, it's what we should say?

@andyleclair agreed

andyleclair: Ok, I have this configured in my config, confirmed working in neovim `0.11.3`

https://github.com/andyleclair/.dotfiles/blob/main/.config/nvim/lua/custom/plugins/elixir.lua

image

If you like, I can update the docs for that? I've got a PR cooking for `nvim-lspconfig` that essentially has the same stuff, I just wanted to make sure it worked first!

doorgan: Awesome, let's do it, thank you! 💜

doorgan (approved):
Thank you! 💜

PR #65, opened 2025-08-29
chore: add install task to justfile

erickgnavar:
A unify task to build and install the binary was missing, also add to
installation documentation how to call it


erickgnavar:
Also it will be nice to reference to `expert` in the `installation.md`(for editor/IDE configuration) file instead of being aware of the target, this way we just use one filename.

doorgan (approved):
Awesome, thank you! 💜

PR #71, opened 2025-08-29
Add Zed to installation.md

PJUllrich:

PR #76, opened 2025-08-29
docs: Add Zed instructions to Editor Installation docs

phinnaeus:
Maybe redundant since it's documented by Zed already, but I didn't look there initially so I figured it might be helpful to have here as well.


st3fan (commented):
Says `Export` and not `Expert`.

doorgan:
This is a duplicate of #71, but thanks for the pr still!

PR #82, opened 2025-08-29
fix: trim any quotes wrapping PATH when elixir is managed by mise

rktjmp:
When Expert.Port detects the elixir executable, it extracts the current env from mise/rtx if present, by calling `mise env -s bash`. This returns

    ...
    export PATH='<path>'
    ...

If we don't strip the surrounding single quotes, later when we re-inject `PATH` into the environmetn, `sh` will fail to intepret it correctly, meaning future calls to any `mise` managed executables such as `erl` will fail as a command not found. This will then cause servers to fail to load and bubble an `{:erpc, :noconnection}` error up to the user.

https://github.com/elixir-lang/expert/issues/59#issuecomment-3236766517

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -87,6 +87,10 @@ defmodule Expert.Port do
           key_and_value
           |> String.split("=", parts: 2)
           |> Enum.map(&String.trim/1)
+          |> then(fn
+            ["PATH", path] -> ["PATH", String.trim(path, "'")]
+            other -> other
+          end)
 
         {key, value}
 
@@ -108,6 +112,10 @@ defmodule Expert.Port do
           key_and_value
           |> String.split("=", parts: 2)
           |> Enum.map(&String.trim/1)
+          |> then(fn
+            ["PATH", path] -> ["PATH", String.trim(path, "'")]
+            other -> other
+          end)
 
         {key, value}
 


rktjmp:
One user in the linked thread said they had issues with `asdf`, but I think its because they were running elixir 1.14, which I don't think the nightly build is generated to support (looking at the CI workflow). This is why I did not apply the patch to the asdf branch. I could not find any examples or mention outside of lexical that referenced `PATH_SAVE` + asdf, so I'm not sure if that would also need adjusting.

You can see the described failure case by doing something like this:

```sh
# in /bin/sh, which is what elixir's shell launcher scripts run under, not bash!
# and assuming you have something to run not in a default path such as /usr/bin etc
# eg: you have ~/.local/bin/expert
$ expert
<ctrl-c> # it works
$ PATH="'$PATH'"
$ expert
sh: expert: command not found
```

doorgan:
> I could not find any examples or mention outside of lexical that referenced PATH_SAVE + asdf

That's a leftover from the previous packaging method, we should remove the `PATH_SAVE` workaround

> One user in the linked thread said they had issues with asdf, but I think its because they were running elixir 1.14

There is one specific project of mine that is giving a similar error to the one described in the comment(it is not finding `erl`) so I think this may be affecting asdf as well

rktjmp:
I did test asdf 0.18, elixir 1.17, erlang 27, expert nightly in a container but it just worked for me.

You could try the same logging I did in that comment, see what `$PATH` is inside the `elixir` script.

doorgan (approved):
Yeah, there is something else that is preventing my asdf from finding `erl` that is unrelated to the bug you found, so let's merge this one

PR #106, opened 2025-09-02
chore(nix): nix package

### LINKED ISSUE #95: Nix build is broken
```
┃        Definition values:
┃        - In `/nix/store/6gczld1j7dw7pgq1ik0mr06nk9apzmma-source/flake.nix':
┃            {
┃              mkExpert = <function, args: {erlang}>;
┃            }
┃        - In `/nix/store/gdp7sdcydj516sjvp81v2nmvxqq11f1r-source/parts/lib.nix':
┃            {
┃              compatibleVersionPackages = <function>;
┃              compatibleVersions = {
┃                "18.3.4.11" = {
┃                  elixirs = {
┃            ...
┃        Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
┣━━━

```

Currently trying to build the package with nix gives this error

adamcstephens:
I opted to use [deps_nix](https://hexdocs.pm/deps_nix/readme.html) as it simplifies hash management, and allows re-use of dependencies as they're individual fixed output derivations (FOD). Unfortunately, it makes this first diff quite large.

I can go back to straight hashes if this feels too invasive for the non-nix users, but it shouldn't affect them outside an extra dep. I also added to all apps even though it seems it may not be necessary for forge and expert_credo?

I've seen issues with builds failing during the release stage, but it's likely a timing issue as re-build works.

cc @mhanberg 

This is an alternative to #96
fixes #95


apps/expert/mix.exs
adamcstephens, line 1: This allows for disabling burrito during release and is only used in the nix package.

mhanberg:
Do you know why the deps.nix files has a bunch of dependencies we don't have like rustler and explorer?

apps/expert/mix.exs
```diff
@@ -49,7 +49,7 @@ defmodule Expert.MixProject do
       expert: [
         strip_beams: false,
         cookie: "expert",
-        steps: release_steps() ++ [&Burrito.wrap/1],
```
mhanberg: Can we not just use the plain release rather than adjusting the burrito release?

adamcstephens:
> Do you know why the deps.nix files has a bunch of dependencies we don't have like rustler and explorer?

There's some nix code in deps.nix for handling rustler and some overrides. These aren't actually being used in this case but are indeed in this file. If you focus solely on the `packages` attribute you can see what packages are actually being included.

Or here's a view from the repl:

```
nix-repl> packages.aarch64-darwin.expert.passthru.mixNixDeps
{
  burrito = «derivation /nix/store/9i02pscamc3r97cdfdm5p51321xcw1im-burrito-1.4.0.drv»;
  finch = «derivation /nix/store/c21mgb9d151r8kx3f9b63mny8knbcpb0-finch-0.20.0.drv»;
  gen_lsp = «derivation /nix/store/h99k1n7szga5x0ypq9w4smjaphayzclb-gen_lsp-0.11.0.drv»;
  hpax = «derivation /nix/store/zb6xmaa8bqq2b564v5rmfg25dvf79x8j-hpax-1.0.3.drv»;
  jason = «derivation /nix/store/8m0qh1afkca668bh4i14a8hyc0jx0ai0-jason-1.4.4.drv»;
  logger_file_backend = «derivation /nix/store/q4lxq4jnbdz8gxgp8451pcrqcb3np2kh-logger_file_backend-0.0.14.drv»;
  mime = «derivation /nix/store/wb2sj5v7ky2czzd8iinrgqv233np12j2-mime-2.0.7.drv»;
  mint = «derivation /nix/store/dhldz73fpvkibr4rpdfqxfziawyy6lxf-mint-1.7.1.drv»;
  nimble_options = «derivation /nix/store/q4cvd8aqisp2qwbkswa5saaf99nizz1m-nimble_options-1.1.1.drv»;
  nimble_parsec = «derivation /nix/store/m9j677z65iw48iyv66hszfg7dqcfxgaf-nimble_parsec-1.2.3.drv»;
  nimble_pool = «derivation /nix/store/swffphbiwaynfygk7xzvlxr9yc39japz-nimble_pool-1.1.0.drv»;
  path_glob = «derivation /nix/store/55hwn8bbns0fifslh2zg4nf4h4vnjnjh-path_glob-0.2.0.drv»;
  req = «derivation /nix/store/rxa63xh4hfr7frn07x344skddhzg8w1d-req-0.5.15.drv»;
  schematic = «derivation /nix/store/aq39whqirp182lq387vyi76g522sjc9v-schematic-0.2.1.drv»;
  snowflake = «derivation /nix/store/kckhxlmxg8kak8319s938xqs0h6gwmar-snowflake-1.0.4.drv»;
  sourceror = «derivation /nix/store/04alxga3s8wg0y59xcliy1ffynw86s94-sourceror-1.10.0.drv»;
  telemetry = «derivation /nix/store/4vh71gahwjpirm9qasndqddixhf810nx-telemetry-1.3.0.drv»;
  typed_struct = «derivation /nix/store/kkysmw7p2clr7fhnnp5lmmk68g9c6kfz-typed_struct-0.3.0.drv»;
}

nix-repl> packages.aarch64-darwin.expert.passthru.engineDeps 
{
  elixir_sense = «derivation /nix/store/mxm9kjz2s9fs9b4s6yrjjb3jl0qd18gq-elixir_sense-e3ddc403554050221a2fd19a10a896fa7525bc02.drv»;
  gen_lsp = «derivation /nix/store/ycvdr58p40nw8cj6rq020pnwszkmpxa1-gen_lsp-0.11.0.drv»;
  jason = «derivation /nix/store/q3pqxbkxdrblx53qr8sy5x5m7jl1vn2v-jason-1.4.4.drv»;
  nimble_options = «derivation /nix/store/6bl0pdjifi8wvc0r2wafnq9k7lv3bayx-nimble_options-1.1.1.drv»;
  nimble_parsec = «derivation /nix/store/yd7b099v9jwn1fksj9b0413mnakzhn1c-nimble_parsec-1.2.3.drv»;
  path_glob = «derivation /nix/store/rgq3cxil1nj2gp6wvm4s3ks3vmx6z61a-path_glob-0.2.0.drv»;
  refactorex = «derivation /nix/store/1a569a8plylzbk3pz8lzsmxr2yrjz2f9-refactorex-0.1.52.drv»;
  schematic = «derivation /nix/store/fax6q4mbvxgbvmsdj3083nfnjm4x2p8k-schematic-0.2.1.drv»;
  snowflake = «derivation /nix/store/4siibhny4630wibb8rfs9fshsgk9ahjp-snowflake-1.0.4.drv»;
  sourceror = «derivation /nix/store/2p443qfcp2v6qsp16yzqg1v4ap38655v-sourceror-1.10.0.drv»;
  telemetry = «derivation /nix/store/4vh71gahwjpirm9qasndqddixhf810nx-telemetry-1.3.0.drv»;
  typed_struct = «derivation /nix/store/03xiv0cajfhg4wrvd4h8j3hg8x6lmr48-typed_struct-0.3.0.drv»;
}
```

apps/expert/mix.exs
adamcstephens: `mix release` will run these steps, which currently includes burrito. I moved burrito down into `release_steps()` and made it conditional on an env var. By default burrito will still be built, but this gives us a way to easily skip the burrito step.

mhanberg:
Gotcha, okay. 

I tested locally on NixOS, seemed to start up. Need to actually test it tho still (I'm about to board an airplane). 

Is this implementation able to be upstreamed to nixpkgs?

When it is put into nixpkgs, do you recommend removing our local package here?

adamcstephens:
This is pretty close to what would be in nixpkgs, but likely we'd switch back to mixFodDeps there with hashes. A nixpkgs package can't import deps.nix from the repo, since that would be IFD (import from derivation). (The source is a derivation, and then we'd be trying to import from within the source). I think deps_nix is great for in-project packages though which is why I suggested it here.

There are benefits for keeping a package in repo here, in that users can easily access and build from any commit in the repo. But there's the downside of keeping deps.nix (or hashes) up to date with the mix.exs/mix.lock. If they drift, the nix package will fail to build, so ideally they're updated together. It's up to you and the other maintainers whether it stays though. :)

I'd be interested in if this package works for you. In my testing with neovim it only generates errors

```

[ERROR][2025-09-02 09:12:38] ...lsp/handlers.lua:562	"** (ErlangError) Erlang error: {:erpc, :noconnection}\n    (kernel 10.2.7.2) erpc.erl:1376: :erpc.call/5\n    (xp_expert 0.1.0) lib/expert/provider/handlers/go_to_definition.ex:17: XPExpert.Provider.Handlers.GoToDefinition.handle/2\n    (xp_expert 0.1.0) lib/expert.ex:94: XPExpert.handle_request/2\n    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:372: anonymous fn/2 in XPGenLSP.loop/3\n    (xp_telemetry 1.3.0) telemetry.erl:324: :xp_telemetry.span/3\n    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:371: anonymous fn/5 in XPGenLSP.loop/3\n    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:601: anonymous fn/4 in XPGenLSP.attempt/4\n    (elixir 1.18.4) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2\n"
[ERROR][2025-09-02 09:12:40] ...lsp/handlers.lua:562	"** (ErlangError) Erlang error: {:erpc, :noconnection}\n    (kernel 10.2.7.2) erpc.erl:1376: :erpc.call/5\n    (xp_expert 0.1.0) lib/expert/state.ex:122: XPExpert.State.apply/2\n    (xp_expert 0.1.0) lib/expert.ex:200: XPExpert.apply_to_state/2\n    (xp_expert 0.1.0) lib/expert.ex:139: XPExpert.handle_notification/2\n    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:528: anonymous fn/2 in XPGenLSP.loop/3\n    (xp_telemetry 1.3.0) telemetry.erl:324: :xp_telemetry.span/3\n    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:524: anonymous fn/4 in XPGenLSP.loop/3\n    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:601: anonymous fn/4 in XPGenLSP.attempt/4\n"
```

adamcstephens:
I will port this to nixpkgs as an unstable release once I see it working fully. :) Then move to stable versioning once y'all do.

apps/expert/mix.exs
mhanberg: What I am saying is that there is a named release called "plain", where we can say "mix release plain" and it doesn't include anything burrito related.

adamcstephens: Ahh yeah, we can do that if you prefer.

adamcstephens: Done.

nix/expert.nix
mhanberg, line 1: do you recommend renaming this file from `expert.nix` to `package.nix`? Or does that not really matter. I think we named it `package.nix` in Next LS

mhanberg:
I'm getting the same error as #59, but it can start and load.

mhanberg:
@adamcstephens in another PR, if you have time could you add a CI check for the nix build?

nix/expert.nix
adamcstephens: It doesn't really matter.

Issue #108, opened 2025-09-02
Don't use EPMD.


mhanberg:
We shouldn't need to use EPMD for clustering as we are always aware of the nodes. 

@josevalim has had similar problems with livebook related to EPMD and has fixed them with a epmdless style approach. 

We should be able to leverage that approach in expert. 

Should help with problems like TODO

josevalim:
Perhaps the steps here will be enough. You start the main one with `-start_epmd false -erl_epmd_port 6789` and each individual instance you start with `-start_epmd false -erl_epmd_port 6789 -dist_listen false`:

https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-epmd-less-deployment

mhanberg:
@josevalim do you know if this should work when many instances of the application? My intuition is telling me that the erl_epmd_port might need to be unique for each instance

josevalim:
Yes, it should. The parent node is listening on port 6789 and each child node connects to port 6789 but does not listen to any (-dist_listen false). This requires, however, for the child node to be the one connecting to the parent.

josevalim:
Oh, sorry, I believe I have misunderstood your question. You want to run multiple Expert servers, not multiple clients, then you are right!

In order to do that, you can write this code:

```elixir
defmodule Expert.EPMD do
  def dist_port do
    :persistent_term.get(:expert_dist_port, nil)
  end

  # EPMD callbacks

  def register_node(name, port), do: register_node(name, port, :inet)

  def register_node(name, port, family) do
    :persistent_term.put(:expert_dist_port, port)

    # We don't care if EPMD is not running
    case :erl_epmd.register_node(name, port, family) do
      {:error, _} -> {:ok, -1}
      {:ok, _} = ok -> ok
    end
  end

  defdelegate start_link(), to: :erl_epmd
  defdelegate port_please(name, host), to: :erl_epmd
  defdelegate port_please(name, host, timeout), to: :erl_epmd
  defdelegate listen_port_please(name, host), to: :erl_epmd
  defdelegate address_please(name, host, family), to: :erl_epmd
  defdelegate names(host_name), to: :erl_epmd
end
```

This is basically a wrapper around EPMD so you get to know which port it is running on. Now, in your vm.args, you will set this: `-epmd_module Elixir.Expert.EPMD`.

Finally, when starting the clients, you will do this:

```elixir
if port = Expert.EPMD.dist_port() do
[...]

PR #115, opened 2025-09-04
Do not start EPMD

jswanner:
Configure release not to start epmd automatically. See #108.


jswanner:
Note: I have no idea what I'm doing, but it is working for me with a local release. Also, using a hardcoded port as suggested in #108 (`-erl_epmd_port 6789`) caused my local release of Expert to only work for one project at a time (leading to the same `Erlang error: {:erpc, :noconnection}` error this was meant to fix), using a dynamic port (`-erl_epmd_port 0`) did not have that problem.

mhanberg:
Yeah the main thing we need is to make sure the engine node (the thing we start in a port with your project code in it) can connect to the same port that the release is advertising.

jswanner:
```elixir
$ ./apps/expert/bin/debug_shell.sh expert
Erlang/OTP 27 [erts-15.2.7.1] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]

Interactive Elixir (1.18.4) - press Ctrl+C to exit (type h() ENTER for help)
Error while evaluating: /Users/jacob/Code/expert/.iex.exs
** (CompileError) cannot compile code (errors have been logged)

iex(manager-expert-38131@127.0.0.1)1> Node.list()
[:"project-expert-38131@127.0.0.1", :"shell@127.0.0.1"]
```

mhanberg:
@jswanner José added some more details that i think we were missing here: https://github.com/elixir-lang/expert/issues/108#issuecomment-3257153811

jswanner:
I don't want this PR to be a distraction, which is why I'm closing it.

I was unable to get José's suggestion of using a custom EPMD module to work. As I said before: I don't really know what I'm doing, but I have learned more about Erlang distribution in this process.

In case it helps anyone else that goes down this path, from what I can when things fail the port for the project node exits immediately after opening:

```
Port #Port<0.14> has exited due to: :normal
```

And, as for the vm.args, the custom EPMD module is used with:

```
-epmd_module Elixir.XPExpert.EPMD
```

PR #162, opened 2025-10-09
fix: let the system figure out the elixir version for the project

doorgan:
This is an alternative to #93 that should fix the issue for any version manager, if any

Previously we were trying to identify the version manager used by the user(asdf or mise/rtx in particular) and try to get them to provide an environment we could use to get the elixir+erlang versions for the project

The issue we're facing is that we can't reliably find an environment that way, so we end up with PATHs that don't include elixir, include the wrong version, or include *only* `elixir` and not `erl`.

Similarly for mise, we had issues where the env parsing stopped working(compared to lexical, from which this code is based on).

This PR uses the approach used by [Radon IDE](https://github.com/software-mansion/radon-ide/blob/feaac732dd6033498913ca76042b27a83b8f1e4f/packages/vscode-extension/src/utilities/subprocess.ts#L8C16-L17): launch an interactive shell on the project path so the OS populates the PATH, grab that PATH and use it to then find `elixir`

By doing this it doesn't matter which version manager you use or by what means elixir is installed, we can reliably get the same version the project expects.

### DIFF
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -171,10 +171,11 @@ defmodule Expert.EngineNode do
     # Expert release, and we build it on the fly for the project elixir+opt
     # versions if it was not built yet.
     defp glob_paths(%Project{} = project) do
+      lsp = Expert.get_lsp()
       project_name = Project.name(project)
       {:ok, elixir, env} = Expert.Port.elixir_executable(project)
 
-      Logger.info("Found elixir for #{project_name} at #{elixir}")
+      GenLSP.info(lsp, "Found elixir for #{project_name} at #{elixir}")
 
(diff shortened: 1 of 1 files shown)


apps/expert/lib/expert/engine_node.ex
```diff
@@ -171,8 +171,11 @@ defmodule Expert.EngineNode do
     # Expert release, and we build it on the fly for the project elixir+opt
     # versions if it was not built yet.
     defp glob_paths(%Project{} = project) do
+      project_name = Project.name(project)
       {:ok, elixir, env} = Expert.Port.elixir_executable(project)
 
+      Logger.info("Found elixir for #{project_name} at #{elixir}")
```
mhanberg: Maybe log this with the LSP logger so it shows up in the output panel.

PR #167, opened 2025-10-15
feat: epmdless deployments

### LINKED ISSUE #59: Error using LSP: `Erlang error: {:erpc, :noconnection}`
Environment:

OS: Mac OS Tahoe & Sequoia (another person here next to me has the same issue)
Elixir: 1.18.4
Erlang: 27.1.2
Editor: Zed & Neovim (same error)
Installation Method: Both the `gh` CLI & building from source
Affected components: any call to the LSP it appears.

```elixir
Erlang error: {:erpc, :noconnection}
    (kernel 10.2.7.1) erpc.erl:1376: :erpc.call/5
    (xp_expert 0.1.0-04cd32c) lib/expert/code_intelligence/completion.ex:84: XPExpert.CodeIntelligence.Completion.completions/3
    (xp_expert 0.1.0-04cd32c) lib/expert/code_intelligence/completion.ex:36: XPExpert.CodeIntelligence.Completion.complete/4
    (xp_expert 0.1.0-04cd32c) lib/expert/provider/handlers/completion.ex:21: XPExpert.Provider.Handlers.Completion.handle/2
    (xp_expert 0.1.0-04cd32c) lib/expert.ex:90: XPExpert.handle_request/2
    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:372: anonymous fn/2 in XPGenLSP.loop/3
    (xp_telemetry 1.3.0) /home/runner/work/expert/expert/apps/expert/deps/telemetry/src/telemetry.erl:324: :xp_telemetry.span/3
    (xp_gen_lsp 0.11.0) lib/gen_lsp.ex:371: anonymous fn/5 in XPGenLSP.loop/3
```

I'm aware this is a nightly build and hiccups are to be expected, so thanks once again for all your hard work, hopefully this issue is helpful! I'm at your disposal if you need any other information, and apologies if I didn't install it right, it is very possible 😆

### LINKED ISSUE #108: Don't use EPMD.

doorgan:
Closes #108 
Should fix #59 

Removes the usage of EPMD, and instead uses a custom EPMD module that provides an epmd port that project nodes use to connect to the manager node directly, instead of relying on epmd to resolve the names to ports.

There is one caveat to this implementation: while it works fairly well, we're still using the ETS search store, which implements a WAL that only supports writes from a single node. We're currently relying on EPMD to let other project nodes discover who is the current wal leader so we can ensure there is always a single writer for the WAL. With this change, we can end up with multiple writers if you open the same project with multiple editors which can lead to issues. I didn't observe problems so far in my tests with multiple editors but it's still something that needs to be changed.

There are two ways to solve that problem: switch to sqlite(we've been wanting to do this for a while), or do something similar to what livebook does to discover the runtime nodes.

### DIFF
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -29,17 +29,21 @@ defmodule Expert.EngineNode do
 
     def start(%__MODULE__{} = state, paths, from) do
       this_node = inspect(Node.self())
+      dist_port = Expert.EPMD.dist_port()
 
-      args = [
-        "--name",
-        Project.node_name(state.project),
-        "--cookie",
-        state.cookie,
-        "--no-halt",
-        "-e",
-        "Node.connect(#{this_node})"
-        | path_append_arguments(paths)
-      ]
+      args =
+        [
+          "--erl",
+          "-start_epmd false -erl_epmd_port #{dist_port} -dist_listen false",
+          "--name",
+          Project.node_name(state.project),
+          "--cookie",
+          state.cookie,
+          "--no-halt",
+          "-e",
+          "Node.connect(#{this_node}); IO.puts(\"ok\")"
+          | path_append_arguments(paths)
+        ]
 
       port = Expert.Port.open_elixir(state.project, args: args)
 
@@ -110,7 +114,6 @@ defmodule Expert.EngineNode do
   use GenServer
 
   def start(project) do
-    :ok = ensure_epmd_started()
     start_net_kernel(project)
 
     node_name = Project.node_name(project)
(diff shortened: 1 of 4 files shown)


apps/expert/lib/expert/engine_node.ex
```diff
@@ -278,7 +273,7 @@ defmodule Expert.EngineNode do
 
   @impl true
   def handle_call({:start, paths}, from, %State{} = state) do
-    :ok = :net_kernel.monitor_nodes(true, node_type: :visible)
+    :ok = :net_kernel.monitor_nodes(true, node_type: :all)
```
doorgan, line 276: In this setup project nodes are `:hidden`

mhanberg:
I works on my work computer where I got the no_connection error

PR #169, opened 2025-10-16
fix: add lsp logging when failing to find an elixir executable

### LINKED ISSUE #128: Current check for tooling manager errors without much information
- Asdf version: 0.18.0-arch (rev 028097)
- Expert version: Nightly on https://github.com/elixir-lang/expert/commit/e9d69dd34c49eef4061c6cc010930a99c616e908

I had both asdf and mise installed, but I only had mise activated. Expert fails to parse an asdf error output.

Here's the rundown:

1. [Expert prioritizes asdf as first manager to check](https://github.com/elixir-lang/expert/blob/e9d69dd34c49eef4061c6cc010930a99c616e908/apps/expert/lib/expert/port.ex#L38)
2. [The check involves checking the manager binary and finding `elixir` in the toolings env](https://github.com/elixir-lang/expert/blob/e9d69dd34c49eef4061c6cc010930a99c616e908/apps/expert/lib/expert/port.ex#L54-L55)
3. I get the following output from asdf (the command to inspect the current manager environment):
```
asdf env elixir && echo Good || echo Bad     
Bad
```
4. [This match fails without fallback for returning `nil`](https://github.com/elixir-lang/expert/blob/e9d69dd34c49eef4061c6cc010930a99c616e908/apps/expert/lib/expert/port.ex#L68-L82), blocking the check from proceeding to try other managers.

#### Suggestions

- Check whether asdf returned a success exit code before attempting to parse the output.
- Even when exit code is successful, if any match fails when parsing the output, invalidate the manager instead of crashing out (or be lenient about it providing warning messages).
- If anything is out of the ordinary when checking tooling managers, provide warnings.

#### An additional related case

Mise fetching deps from `.tool-versions` instead of `mise.config.toml` (no asdf installed), described in more detail here https://github.com/elixir-lang/expert/issues/128#issuecomment-3300152289.

In this case `mise env elixir && echo good || echo bad` outputs below. Same error when matching output lines as the one described in the issue and suggestions should address it the same way.

```
export PATH='/Users/REDACTED_USERNAME/.local/share/mise/installs/erlang/28.0.2/bin:/Users/REDACTED_USERNAME/.local/share/mise/installs/elixir/1.18.4-otp-28/bin:/Users/REDACTED_USERNAME/.local/share/mise/installs/elixir/1.18.4-otp-28/.mix/escripts:/Users/REDACTED_USERNAME/.local/share/mise/installs/node/24.1.0/bin:/opt/homebrew/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Library/TeX/texbin:/opt/homebrew/bin:/Users/REDACTED_USERNAME/.orbstack/bin:/Users/REDACTED_USERNAME/.orbstack/bin'
good
```

doorgan:
Provides additional information about what Expert tried to use to find Elixir
Closes #128

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -42,15 +42,25 @@ defmodule Expert.Port do
     shell = System.get_env("SHELL")
 
     {path, 0} = System.cmd(shell, ["-i", "-l", "-c", "cd #{root_path} && echo $PATH"])
-    elixir = :os.find_executable(~c"elixir", to_charlist(path))
 
-    env =
-      Enum.map(System.get_env(), fn
-        {"PATH", _path} -> {"PATH", path}
-        other -> other
-      end)
+    case :os.find_executable(~c"elixir", to_charlist(path)) do
+      false ->
+        GenLSP.error(
+          Expert.get_lsp(),
+          "Couldn't find an elixir executable for project at #{root_path}. Using shell at #{shell} with PATH=#{path}"
+        )
 
-    {:ok, elixir, env}
+        {:error, :no_elixir}
+
+      elixir ->
+        env =
+          Enum.map(System.get_env(), fn
+            {"PATH", _path} -> {"PATH", path}
+            other -> other
+          end)
+
+        {:ok, elixir, env}
+    end
   end
 
   @doc """


mhanberg:
You can ignore that nix ci failure.

PR #185, opened 2025-10-28
feat!: add CLI flag handling

### LINKED ISSUE #81: Implement CLI flags
## Description

The cli executable should implement the cli flags that Next LS supported.

- `--version`
- `--help`
- `--stdio`
- `--port <port>`

andyleclair:
Closes #81

### DIFF
--- a/apps/expert/lib/expert/application.ex
+++ b/apps/expert/lib/expert/application.ex
@@ -12,13 +12,65 @@ defmodule Expert.Application do
 
   @impl true
   def start(_type, _args) do
+    {opts, _argv, _invalid} =
+      OptionParser.parse(Burrito.Util.Args.argv(),
+        strict: [version: :boolean, help: :boolean, stdio: :boolean, port: :integer]
+      )
+
+    help_text = """
+    Expert v#{Expert.vsn()}
+
+    The official language server for Elixir
+
+      Home page: https://expert-lsp.org
+    Source code: https://github.com/elixir-lang/expert
+
+    expert [flags]
+
+    #{IO.ANSI.bright()}FLAGS#{IO.ANSI.reset()}
+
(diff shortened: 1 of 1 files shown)


doorgan:
Looks good and works as advertised!
One important detail is that every expert integration so far assumes expert uses stdio by default. This changes so that you need to explicitly tell it which transport to use. If none is provided, it shows the help message.
We should update the installation docs too, or default to launch expert on stdio and require the `--help` flag for the help message

doorgan:
cc @mhanberg, if we want to change the default behavior I think this is the best time to do so since we're not even 0.1

mhanberg:
Yes, I agree. 

I'm going to pull this down and make sure it works before I approve.

andyleclair:
> Looks good and works as advertised! One important detail is that every expert integration so far assumes expert uses stdio by default. This changes so that you need to explicitly tell it which transport to use. If none is provided, it shows the help message. We should update the installation docs too, or default to launch expert on stdio and require the `--help` flag for the help message

So, I copied that code basically verbatim from NextLS. I'm totally happy to make it default to stdio!

andyleclair:
@mhanberg @doorgan I think it makes the most sense to leave the default as stdio (no need to break any clients imo), does it make sense to even have the flag? I could go either way

mhanberg:
I want the `--stdio` flag, definitely. IDK if i care as much if it defaults to stdio without any flags, but seems weird. Most typescript based servers require the flag I think.

doorgan:
FWIW we're at a point where we have the luxury of making breaking changes, so we can ship this as-is if we want the flag

mhanberg:
I was testing this and have a patch to push up, will hopefully finish it tonight.

mhanberg:
Never mind, this is good to go 👍

PR #188, opened 2025-11-01
fix(engine): handle failing build script

niklas:
When the build script fails for some reason, the EngineNode hung, because the :EXIT would not be received, as the Port was started in some other process, but not in the EngineNode's GenServer. Only the latter explicitely traps exits.

To fix this without causing any side effects (handle_info(:EXIT) in Node), we run the Port in a Task subprocess that traps exits.

Additionally, we log the last line received from the build script to help the user to debug the issue, instead of failing silently with exit reason :normal.

### DIFF
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -215,15 +215,14 @@ defmodule Expert.EngineNode do
           GenLSP.info(lsp, "Finding or building engine for project #{project_name}")
 
           with_progress(project, "Building engine for #{project_name}", fn ->
-            Task.async(fn ->
+            fn ->
               Process.flag(:trap_exit, true)
 
-              Port.open(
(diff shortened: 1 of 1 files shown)


apps/expert/lib/expert/engine_node.ex
```diff
@@ -230,20 +232,20 @@ defmodule Expert.EngineNode do
...
       receive do
         {^port, {:data, ~c"engine_path:" ++ engine_path}} ->
           engine_path = engine_path |> to_string() |> String.trim()
           Logger.info("Engine build available at: #{engine_path}")
 
           {:ok, ebin_paths(engine_path)}
 
-        {^port, _data} ->
-          wait_for_engine(port)
+        {^port, {:data, data}} ->
```
mhanberg, line 243: It might also help if we log the build output here.

PR #205, opened 2025-11-10
feat: epmdless clustering

doorgan:
Alternative to #191 without introducing dependencies.

TODO:
- [x] Figure out why the engine node thinks `Forge.EPMD` does not exist.

### DIFF
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -28,23 +28,6 @@ defmodule Expert.EngineNode do
     @dialyzer {:nowarn_function, start: 3}
 
     def start(%__MODULE__{} = state, paths, from) do
-      epmd_module = to_charlist(Forge.EPMD)
-
-      case :init.get_argument(:epmd_module) do
-        {:ok, [[^epmd_module]]} ->
-          :ok
-
-        _ ->
-          Application.put_env(:kernel, :epmd_module, Forge.EPMD, persistent: true)
-
-          # Note: this is a private API
-          if :net_kernel.epmd_module() != Forge.EPMD do
-            raise("""
-            you must set the environment variable ELIXIR_ERL_OPTIONS="-epmd_module #{Forge.EPMD}"
-            """)
-          end
-      end
-
       this_node = to_string(Node.self())
       dist_port = Forge.EPMD.dist_port()
 
--- a/apps/expert/lib/expert/application.ex
+++ b/apps/expert/lib/expert/application.ex
@@ -64,6 +64,8 @@ defmodule Expert.Application do
           System.halt(1)
       end
 
+    ensure_epmd_module!()
+
     children = [
       {Forge.NodePortMapper, []},
       document_store_child_spec(),
@@ -89,5 +91,24 @@ defmodule Expert.Application do
   @doc false
   def document_store_child_spec do
     {Document.Store, derive: [analysis: &Forge.Ast.analyze/1]}
+  end
+
+  def ensure_epmd_module! do
+    epmd_module = to_charlist(Forge.EPMD)
+
+    case :init.get_argument(:epmd_module) do
+      {:ok, [[^epmd_module]]} ->
+        :ok
+
+      _ ->
+        Application.put_env(:kernel, :epmd_module, Forge.EPMD, persistent: true)
+
+        # Note: this is a private API
+        if :net_kernel.epmd_module() != Forge.EPMD do
+          raise("""
+          you must set the environment variable ELIXIR_ERL_OPTIONS="-epmd_module #{Forge.EPMD}"
+          """)
+        end
+    end
   end
 end


apps/expert/rel/remote.vm.args.eex
```diff
@@ -0,0 +1,13 @@
...
+## Increase number of concurrent ports/sockets
+##+Q 65536
+
+## Tweak GC to run more often
+##-env ERL_FULLSWEEP_AFTER 10
+
+## Enable deployment without epmd
+## (requires changing both vm.args and remote.vm.args)
+-epmd_module Elixir.XPForge.EPMD
+-start_epmd false -dist_listen false
```
mhanberg: I don't think we need the -dist_listen false anymore

mhanberg:
>  Figure out why the engine node thinks Forge.EPMD does not exist.

Are you seeing this in tests or when you actually put it up?

doorgan:
@mhanberg both in tests and when building a release, I'm getting this error:
```
Kernel pid terminated (application_controller) ("{application_start_failure,kernel,{{shutdown,{failed_to_start_child,net_sup,{shutdown,{failed_to_start_child,'Elixir.Forge.EPMD',{'EXIT',{undef,[{'Elixir.Forge.EPMD',start_link,[],[]},{supervisor,do_start_child_i,3,[{file,\"supervisor.erl\"},{line,959}]},{supervisor,do_start_child,3,[{file,\"supervisor.erl\"},{line,945}]},{supervisor,'-start_children/2-fun-0-',3,[{file,\"supervisor.erl\"},{line,929}]},{supervisor,children_map,4,[{file,\"supervisor.erl\"},{line,1820}]},{supervisor,init_children,2,[{file,\"supervisor.erl\"},{line,889}]},{gen_server,init_it,2,[{file,\"gen_server.erl\"},{line,2229}]},{gen_server,init_it,6,[{file,\"gen_server.erl\"},{line,2184}]}]}}}}}},{kernel,start,[normal,[]]}}}")
```

apps/expert/rel/remote.vm.args.eex
doorgan: Yeah, this is a leftover of me splitting hairs

mhanberg:
> @mhanberg both in tests and when building a release, I'm getting this error:
> 
> ```
> Kernel pid terminated (application_controller) ("{application_start_failure,kernel,{{shutdown,{failed_to_start_child,net_sup,{shutdown,{failed_to_start_child,'Elixir.Forge.EPMD',{'EXIT',{undef,[{'Elixir.Forge.EPMD',start_link,[],[]},{supervisor,do_start_child_i,3,[{file,\"supervisor.erl\"},{line,959}]},{supervisor,do_start_child,3,[{file,\"supervisor.erl\"},{line,945}]},{supervisor,'-start_children/2-fun-0-',3,[{file,\"supervisor.erl\"},{line,929}]},{supervisor,children_map,4,[{file,\"supervisor.erl\"},{line,1820}]},{supervisor,init_children,2,[{file,\"supervisor.erl\"},{line,889}]},{gen_server,init_it,2,[{file,\"gen_server.erl\"},{line,2229}]},{gen_server,init_it,6,[{file,\"gen_server.erl\"},{line,2184}]}]}}}}}},{kernel,start,[normal,[]]}}}")
> ```

What stands out to me is that the module name is not namespaced.

apps/engine/lib/engine/bootstrap.ex
```diff
@@ -12,6 +12,8 @@ defmodule Engine.Bootstrap do
   require Logger
 
   def init(%Project{} = project, document_store_entropy, app_configs) do
+    Application.put_env(:kernel, :epmd_module, Forge.EPMD, persistent: true)
```
mhanberg: Do you know why this is necessary?

apps/expert/lib/expert/engine_node.ex
```diff
@@ -29,19 +29,29 @@ defmodule Expert.EngineNode do
...
-        | path_append_arguments(paths)
-      ]
-
-      case Expert.Port.open_elixir(state.project, args: args) do
+      dist_port = Forge.EPMD.dist_port()
+
+      args =
+        [
+          "--erl",
+          "-start_epmd false -epmd_module #{Forge.EPMD}",
```
mhanberg, line 37: I have a feeling this is not being namspaced correctly. I would try just making it a string `-epmd_module Elixir.XPForge.EPMD`

doorgan: It is namespacing correctly, it does get converted to `Elixir.XPForge.EPMD`

apps/engine/lib/engine/bootstrap.ex
doorgan: I't not or it should not be, I'm trying out things from the Livebook codebase
It seems like the vm tries to find the EPMD module *before* loading the code paths provided via `-pa`
Here in particular I was trying to not specify the epmd module with erl flags and instead try to change the epmd module after the engine node starts. This line wasn't reached because the rpc command to bootstrap didn't get through, but this in the `-e` flag doesn't help either

doorgan:
> What stands out to me is that the module name is not namespaced.

Ah, that log is from running tests, where namespacing doesn't run
If I build a release I get this in the lsp logs:
```
[ERROR][2025-11-10 19:03:47] ...p/_transport.lua:36     "rpc"   "/Users/dorgan/dev/expert/apps/expert/burrito_out/expert_darwin_arm64"  "stderr"        "plication_controller) (\"{application_start_failure,kernel,{{shutdown,{failed_to_start_child,net_sup,{shutdown,{failed_to_start_child,'Elixir.XPForge.EPMD',{'EXIT',{undef,[{'Elixir.XPForge.EPMD',start_link,[],[]},{supervisor,do_start_child_i,3,[{file,\\\"supervisor.erl\\\"},{line,420}]},{supervisor,do_start_child,2,[{file,\\\"supervisor.erl\\\"},{line,406}]},{supervis"
```

apps/expert/lib/expert/engine_node.ex
mhanberg: Ahh, gotcha

doorgan:
I can reproduce the same issues on a barebones mix project: https://github.com/doorgan/epmdless

It does partially work if I follow @josevalim's example and run `elixirc` on a file containing both modules(although I get the `requested disconnect from node in order to prevent overlapping partitions` warning and children can't connect to each other), but it breaks inside a mix project

mhanberg:
I haven't researched further to validate and find a solution, but I believe the reason this is happening in the engine but not in expert is because expert is started as a release, which i believe by default loads all modules before the app starts.

Normal `mix run` or `mix phx.server` or `iex -S mix` invocations will lazily load modules as they are started, which is why it isn't present when it tries to utilize the `-epmd_module`. Which it seems to not load it when it tries to invoke it i guess.

Relevant section from the release docs at https://hexdocs.pm/elixir/releases.html#operating-system-scripts

```
# # Set the release to load code on demand (interactive) instead of preloading (embedded).
# export RELEASE_MODE=interactive
```

josevalim:
I can take a look later but in Livebook we don't start the node in the release, we explicitly call `:net_kernel.start` or similar to start them. It may be for the reasons above. i will investigate later.

doorgan:
@josevalim thanks for all the help! This PR is now ready for testing and review

justfile
```diff
@@ -32,11 +32,27 @@ mix project="all" *args="":
     case {{ project }} in
       all)
         for proj in {{ apps }}; do
-          (cd "apps/$proj" && mix {{args}})
+          case $proj in
+            expert)
+              (cd "apps/$proj" && elixir --erl "-start_epmd false -epmd_module Elixir.Forge.EPMD" -S mix {{args}})
+            ;;
+            engine)
+              (cd "apps/$proj" && elixir --erl "-start_epmd false -epmd_module Elixir.Forge.EPMD" -S mix {{args}})
```
doorgan, line 40: I think it's still useful to keep them separate even if they're the same right now but I don't feel strongly about it

apps/expert/lib/expert/engine_node.ex
```diff
@@ -28,20 +28,56 @@ defmodule Expert.EngineNode do
...
+        _ ->
+          Application.put_env(:kernel, :epmd_module, Forge.EPMD, persistent: true)
+
+          # Note: this is a private API
+          if :net_kernel.epmd_module() != Forge.EPMD do
+            raise("""
+            you must set the environment variable ELIXIR_ERL_OPTIONS="-epmd_module #{Forge.EPMD}"
+            """)
+          end
+      end
```
josevalim: Should we do this when expert boots rather than when we try to start a child node?

apps/expert/rel/remote.vm.args.eex
```diff
@@ -0,0 +1,13 @@
...
+
+## Increase number of concurrent ports/sockets
+##+Q 65536
+
+## Tweak GC to run more often
+##-env ERL_FULLSWEEP_AFTER 10
+
+## Enable deployment without epmd
+## (requires changing both vm.args and remote.vm.args)
+-epmd_module Elixir.XPForge.EPMD
```
josevalim, line 12: Please double check this is indeed necessary. You may be fine using epmd for remote debugging (or maybe remote debugging is pointless).

josevalim (approved):
Looks good to me! I have added some comments but none of them are stoppers (and they might be wrong :D)

mhanberg (approved):
Works on my corporate macbook and on my personal linux desktop

:shipit:

PR #207, opened 2025-11-12
docs(installation): update zed installation instructions to include setting the --stdio flag

luhagel:
Until the zed extension handles this natively, Zed users need to add an extra `arguments` setting within the `lsp` section of their config.

### DIFF
--- a/pages/installation.md
+++ b/pages/installation.md
@@ -291,6 +291,13 @@ So, first install the extension and then update your `settings.json` to use Expe
 
 ```json
 {
+  "lsp": {
+    "expert": {
+      "binary": {
+        "arguments": ["--stdio"]
+      }
+    }
+  },
   "languages": {
     "Elixir": {
       "language_servers": [


luhagel:
Talking about the Zed Extension handling it natively: https://github.com/zed-extensions/elixir/pull/45

doorgan:
It seems that PR was already merged, so maybe we could update the docs here to indicate how to pass the `--port <port>` flag if you want to use tcp mode instead?

luhagel:
Works for me - I'd maybe consider merging this for now, as there can be a couple days to a couple weeks delay between merging and publishing, had that with HEEX support for the emmet extension a few month ago.
So maybe get the info out there now (optionally amended with the tcp docs) and then counter-commit once it's actually available to the public?

Just let me know your preferred path of action and I'm happy to oblige

(Or just scrap it, it's not like I poured hours of work into this ^^)

doorgan:
> there can be a couple days to a couple weeks delay between merging and publishing

Ah got it, makes sense then, let's merge this

doorgan (approved):

PR #238, opened 2025-11-27
refactor: update client support function to match the configuration state

### LINKED ISSUE #113: Unexpected return values for `Expert.Configuration.client_supports?/1`
While debugging something I stumbled upon this:

```elixir
iex> Expert.Configuration.client_supports?(:tags)
%{value_set: [1]}

iex> Expert.Configuration.client_supports?(:signature_help)
%GenLSP.Structures.SignatureHelpClientCapabilities{
  context_support: nil,
  signature_information: %{
    documentation_format: ["markdown", "plaintext"],
    parameter_information: %{label_offset_support: true},
    active_parameter_support: true
  },
  dynamic_registration: false
}
```

The function spec claims the function only returns a boolean:

https://github.com/elixir-lang/expert/blob/e510182b90fc931cc0fe9e65d4a241521d990312/apps/expert/lib/expert/configuration.ex#L56-L57

I wanted to prepare a PR, but then I was wondering whether we should simply check for `value !== false` to see if that feature is supported or if there is more logic needed to figure out whether those features are supported or not 🤷

doorgan:
The previous function claimed to return a boolean, but given the nested nature of the configuration struct, it's more useful to make it return whatever is stored instead

Closes #113

### DIFF
--- a/apps/expert/lib/expert/configuration.ex
+++ b/apps/expert/lib/expert/configuration.ex
@@ -53,12 +53,12 @@ defmodule Expert.Configuration do
     :persistent_term.get(__MODULE__, false) || new()
   end
 
-  @spec client_supports?(atom()) :: boolean()
-  def client_supports?(key) when is_atom(key) do
-    client_supports?(get().support, key)
+  @spec client_support(atom()) :: term()
+  def client_support(key) when is_atom(key) do
+    client_support(get().support, key)
   end
 
-  defp client_supports?(%Support{} = client_support, key) do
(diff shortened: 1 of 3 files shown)


joaothallis (approved):

mhanberg:
@joaothallis please don't approve pull requests if you aren't on the team

mhanberg:
I think this is actually a `refactor` rather than a fix.

doorgan:
I was a bit on the fence because technically this changes the code behavior, but yeah for all intents and purposes it's a refactor!

Issue #241, opened 2025-12-03
Failure spawning project using a neovim frontend and mise


badosu:
I have been using expert with neovide (a neovim frontend) and mason for quite a while.

After the last update I noticed a peculiar failure, the "engine" builds but fails to spawn the project child:

<details>
<summary>expert.log excerpt</summary>
<pre>
21:32:21.960 [info] Engine build available at: /home/badosu/.local/share/Expert/0.1.0-b237fd5/elixir-1.18.4-erts-14.2.5.2/337a120c4a0f3608086a2f2a3a040525/_build/dev_ns
21:32:21.992 [debug] sent notification server -> client $/progress
21:32:22.174 [debug] Node port message: /home/badosu/.local/share/mise/installs/elixir/1.18.4-otp-26/bin/elixir: line 244: exec: erl: not found

21:32:22.176 [info] Port #Port<0.11> has exited due to: :normal
21:32:24.993 [debug] sent notification server -> client $/progress
21:32:24.993 [error] GenServer :"teiserver::node_process" terminating
** (stop) :start_timeout
Last message: :maybe_start_timeout
State: nil
21:32:24.994 [error] Process #PID<0.174.0> terminating
** (exit) {:error, :start_timeout}
    (stdlib 6.2.2.1) gen_server.erl:2210: :gen_server.init_it/6
    (stdlib 6.2.2.1) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Initial Call: XPExpert.Project.Node.init/1
Ancestors: [:"teiserver::supervisor", XPExpert.ProjectSupervisor, XPExpert.Supervisor, #PID<0.158.0>]
Message Queue Length: 0
Messages: []
Links: [#PID<0.171.0>]
Dictionary: []
Trapping Exits: false
Status: :running
Heap Size: 28690
Stack Size: 29
Reductions: 21266
21:32:24.994 [error] Process :"teiserver::node_process" (#PID<0.182.0>) terminating
** (exit) :start_timeout
    (stdlib 6.2.2.1) gen_server.erl:2476: :gen_server.handle_common_reply/8
    (stdlib 6.2.2.1) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
</pre>
</details>

Of particular note is the following line:

```
21:32:22.174 [debug] Node port message: /home/badosu/.local/share/mise/installs/elixir/1.18.4-otp-26/bin/elixir: line 244: exec: erl: not found
```

[...]

PR #244, opened 2025-12-05
fix: properly log when engine fails to initialize

badosu:
Previously whenever there was some error initializing the project node a few things would happen:

- The node supervisor start code would fail with match error
- The server would shutdown via init_timeout
- There would be no indication as to the reason, except reading the node message debug message in `expert.log`

Now:

- We properly respond to an initialize request with a server not initialized error
- We gracefully shut down the node processes
- We send an error message to the client with the last message sent by the node:
```
Node exited with status 127, last message: /home/badosu/.local/share/mise/installs/elixir/1.18.4-otp-26/bin/elixir: line 244: exec: erl: not found
```

This PR is motivated as an attempt to mitigate one of the many situations where expert fails silently, e.g.: https://github.com/elixir-lang/expert/issues/241

### Remarks

- We additionally remove code related to waiting for `didChangeConfiguration` since it was noop

### DIFF
--- a/apps/expert/lib/expert.ex
+++ b/apps/expert/lib/expert.ex
@@ -59,7 +59,7 @@ defmodule Expert do
       Task.Supervisor.start_child(:expert_task_queue, fn ->
         config = state.configuration
 
-        log_info(lsp, "Starting project at uri #{config.project.root_uri}")
+        log_info(lsp, "Starting project")
 
         start_result = Project.Supervisor.start(config.project)
 
@@ -209,22 +209,22 @@ defmodule Expert do
   end
 
   def log_info(lsp \\ get_lsp(), message) do
-    Logger.info(message)
+    message = log_prepend_project_root(message, assigns(lsp).state)
 
+    Logger.info(message)
     GenLSP.info(lsp, message)
   end
 
   # When logging errors we also notify the client to display the message
   def log_error(lsp \\ get_lsp(), message) do
+    message = log_prepend_project_root(message, assigns(lsp).state)
+
     Logger.error(message)
-
-    log_level = Enumerations.MessageType.error()
-
     GenLSP.error(lsp, message)
 
     GenLSP.notify(lsp, %GenLSP.Notifications.WindowShowMessage{
       params: %GenLSP.Structures.ShowMessageParams{
-        type: log_level,
+        type: Enumerations.MessageType.error(),
         message: message
       }
     })
@@ -331,4 +331,12 @@ defmodule Expert do
         "Failed to start engine #{name}: #{inspect(reason)}"
     end
   end
+
+  defp log_prepend_project_root(message, %State{
+         configuration: %Expert.Configuration{project: %Forge.Project{} = project}
+       }) do
+    "[Project #{project.root_uri}] #{message}"
+  end
+
+  defp log_prepend_project_root(message, _state), do: message
 end
--- a/apps/expert/test/expert_test.exs
+++ b/apps/expert/test/expert_test.exs
@@ -1,4 +1,7 @@
 defmodule Expert.ExpertTest do
+  alias Expert.State
+  alias Forge.Test.Fixtures
+
   use ExUnit.Case, async: true
   use Patch
 
@@ -7,16 +10,14 @@ defmodule Expert.ExpertTest do
   test "sends an error message on engine initialization error" do
     with_patched_transport()
 
-    assigns = start_supervised!(GenLSP.Assigns, id: make_ref())
-    GenLSP.Assigns.merge(assigns, %{state: %{}})
-
-    lsp = %GenLSP.LSP{mod: Expert, assigns: assigns}
+    project = Fixtures.project()
+    lsp = initialize_lsp(project)
 
     reason = :something_bad
 
     assert {:noreply, ^lsp} = Expert.handle_info({:engine_initialized, {:error, reason}}, lsp)
 
-    error_message = "Failed to initialize: #{inspect(reason)}"
+    error_message = "[Project #{project.root_uri}] Failed to initialize: #{inspect(reason)}"
     error_message_type = GenLSP.Enumerations.MessageType.error()
 
     assert_receive {:transport,
@@ -34,5 +35,37 @@ defmodule Expert.ExpertTest do
                         message: ^error_message
                       }
                     }}
+  end
+
+  defp initialize_lsp(project) do
+    assigns = start_supervised!(GenLSP.Assigns, id: make_ref())
+
+    {:ok, _response, state} = State.initialize(State.new(), initialize_request(project))
+    GenLSP.Assigns.merge(assigns, %{state: state})
+
+    %GenLSP.LSP{mod: Expert, assigns: assigns}
+  end
+
+  defp initialize_request(project) do
+    root_uri = project.root_uri
+    root_path = Forge.Project.root_path(project)
+
+    %GenLSP.Requests.Initialize{
+      id: 1,
+      jsonrpc: "2.0",
+      method: "initialize",
+      params: %GenLSP.Structures.InitializeParams{
+        capabilities: %GenLSP.Structures.ClientCapabilities{},
+        process_id: "",
+        root_uri: root_uri,
+        root_path: root_path,
+        workspace_folders: [
+          %GenLSP.Structures.WorkspaceFolder{
+            name: root_path,
+            uri: root_uri
+          }
+        ]
+      }
+    }
   end
 end


apps/expert/lib/expert/engine_node.ex
```diff
@@ -372,10 +388,17 @@ defmodule Expert.EngineNode do
     {:stop, :shutdown, state}
   end
 
+  @impl true
+  def handle_info({_port, {:exit_status, exit_status}}, %State{} = state) do
+    state = State.on_exit_status(state, exit_status)
+
+    {:stop, :shutdown, state}
```
badosu: Now that we shutdown here, it won't fail with init_timeout and will provide a helpful reason.

apps/expert/lib/expert/port.ex
```diff
@@ -112,7 +112,7 @@ defmodule Expert.Port do
         opts
       end
 
-    Port.open({:spawn_executable, launcher}, [:stderr_to_stdout | opts])
+    Port.open({:spawn_executable, launcher}, [:stderr_to_stdout, :exit_status] ++ opts)
```
badosu, line 115: Whenever the process exits with a code, send a `{:exit_status, code}` message to the port.

apps/expert/lib/expert/state.ex
```diff
@@ -50,17 +50,15 @@ defmodule Expert.State do
       end
 
     config = Configuration.new(event.root_uri, event.capabilities, client_name)
-    new_state = %__MODULE__{state | configuration: config, initialized?: true}
     Logger.info("Starting project at uri #{config.project.root_uri}")
 
-    response = initialize_result()
-
-    Task.Supervisor.start_child(:expert_task_queue, fn ->
-      {:ok, _pid} = Project.Supervisor.start(config.project)
```
badosu, line 59: This match would expect the project to always start successfully, which is not the case.

apps/expert/lib/expert/state.ex
```diff
@@ -50,17 +50,15 @@ defmodule Expert.State do
       end
 
     config = Configuration.new(event.root_uri, event.capabilities, client_name)
-    new_state = %__MODULE__{state | configuration: config, initialized?: true}
     Logger.info("Starting project at uri #{config.project.root_uri}")
 
-    response = initialize_result()
-
-    Task.Supervisor.start_child(:expert_task_queue, fn ->
```
badosu, line 58: Is there a reason to keep this asynchronous? Now we wait for the status of the child start.

mhanberg: We purposely want the engine to start asynchronously so we can initialize the server as soon as possible. Starting the engine can be slow, and an editor clients timeout if it takes too long.

badosu: So should we respond as initialized and send a generic error message on initialization failure instead?

doorgan: We should respond as initialized and then reply with the equivalent of an empty response to subsequent requests that require the engine to be started. After #18 this would be "send an empty request if it requires that specific engine to be started"

badosu: I changed it to still perform same initialization logic as before, with the only difference that we now at least send an error message to the client if an initialization error occurred.

apps/expert/lib/expert.ex
```diff
@@ -51,19 +52,33 @@ defmodule Expert do
 
   def handle_request(%GenLSP.Requests.Initialize{} = request, lsp) do
     state = assigns(lsp).state
-    Process.send_after(self(), :default_config, :timer.seconds(5))
```
badosu, line 54: Note (same as before): the logic behind `:default_config` is noop since we always initialize the `state.configuration`. I removed this and related lines.

apps/expert/lib/expert.ex
```diff
@@ -273,4 +278,33 @@ defmodule Expert do
...
+      child ->
+        "Failed to start child #{inspect(child)}: #{inspect(reason)}"
+    end
+  end
+
+  defp initialization_error_message(reason) do
+    "Failed to initialize: #{inspect(reason)}"
+  end
+
+  defp node_initialization_message(name, reason) do
```
badosu, line 320: Much more helpful message available to the client via log inspection.

apps/expert/lib/expert/state.ex
```diff
@@ -35,8 +35,6 @@ defmodule Expert.State do
     %__MODULE__{}
   end
 
-  # TODO: this function has a side effect (starting the project supervisor)
-  # that i think might be better off in the calling function
```
badosu, line 39: Done

apps/expert/test/expert_test.exs
```diff
@@ -0,0 +1,96 @@
...
+        port: port
+      }
+
+      client = GenLSP.Test.client(expert)
+
+      [expert: expert, client: client]
+    end
+
+    test "replies to initialize with expert info and capabilities", %{client: client} do
+      id = System.unique_integer([:positive])
```
badosu: I felt the need to actually do a full test of the expert server, but in the end it was just much easier to do unit testing of `handle_info`, I don't mind discarding this if it is considered to not be valuable.

apps/expert/lib/expert.ex
```diff
@@ -185,20 +200,10 @@ defmodule Expert do
     {:noreply, lsp}
   end
 
-  def handle_info(:default_config, lsp) do
-    state = assigns(lsp).state
+  def handle_info({:engine_initialized, {:error, reason}}, lsp) do
+    GenLSP.error(get_lsp(), initialization_error_message(reason))
```
badosu: Perhaps we should terminate if there was an initialization error?

I've found the dangling LSP process to be very annoying (especially before https://github.com/elixir-lang/expert/pull/250 was merged).

apps/expert/lib/expert/engine_node.ex
```diff
@@ -47,9 +50,14 @@ defmodule Expert.EngineNode do
           # If we start distribution manually after all the code is loaded,
           # everything works fine.
           """
-          {:ok, _} = Node.start(:"#{Project.node_name(state.project)}", :longnames)
+          node_start = Node.start(:"#{Project.node_name(state.project)}", :longnames)
           #{Forge.NodePortMapper}.register()
```
mhanberg: I think it makes sense to put this bit inside the {:ok, _} clause of the case expression, because if we don't cluster, we can't register with the node port mapper

apps/expert/lib/expert/engine_node.ex
```diff
@@ -103,6 +111,25 @@ defmodule Expert.EngineNode do
       end
     end
 
+    def on_exit_status(%__MODULE__{} = state, exit_status) do
+      Logger.debug(
```
mhanberg: instead, lets put an info log in the 0 clause of the case, saying the engine shutdown

then in the error clause, put an error log saying it shutdown unexpectedly

apps/expert/lib/expert.ex
```diff
@@ -51,19 +52,33 @@ defmodule Expert do
...
-      {:ok, response, state} ->
-        lsp = assign(lsp, state: state)
-        {:ok, response} = Expert.Protocol.Convert.to_lsp(response)
+    with {:ok, response, state} <- State.initialize(state, request),
+         {:ok, response} <- Expert.Protocol.Convert.to_lsp(response) do
+      config = state.configuration
 
-        {:reply, response, lsp}
+      Task.Supervisor.start_child(:expert_task_queue, fn ->
+        Logger.info("Starting project at uri #{config.project.root_uri}")
```
mhanberg: Lets also add a GenLSP.info log here with the same message.

in nextls i had a "logger" process/module that would log to both, probably need something similar here (probably just a module). (no need to look into that as part of this PR)

apps/expert/lib/expert.ex
mhanberg: No, because we might have other workspace folders that are fine. We eventually probably want a workspace command to restart the engine.

Here, in addition to an error log, lets send a [window/showMessage](https://hexdocs.pm/gen_lsp/0.11.1/GenLSP.Notifications.WindowShowMessage.html) notification as well. This will actively show up as a toast in their editor, where it won't be missed.

apps/expert/lib/expert.ex
```diff
@@ -273,4 +278,33 @@ defmodule Expert do
...
+  defp initialization_error_message(reason) do
+    "Failed to initialize: #{inspect(reason)}"
+  end
+
+  defp node_initialization_message(name, reason) do
+    case reason do
+      # NOTE:
+      # ** (Mix.Error) httpc request failed with: ... Could not install Hex because Mix could not download metadata ...
+      {:shutdown, {:error, :normal, message}} ->
+        "Node #{name} shutdown with error:\n\n#{message}"
```
mhanberg: Let's change all this verbiage from "node" to "engine". I believe it'll help bug reports in the future, as a non contributor will probably be confused by the usage of "node"

apps/expert/test/expert_test.exs
```diff
@@ -0,0 +1,96 @@
...
+  require GenLSP.Test
+
+  import Expert.Test.Protocol.TransportSupport
+
+  defp start_expert do
+    patch(System, :argv, fn -> ["--port", "0"] end)
+
+    assert {:ok, _} = Application.ensure_all_started(:expert)
+
+    on_exit(fn -> Application.stop(:expert) end)
```
mhanberg: there are helpers to in GenLSP already for starting the server with the tcp adapter.

badosu: Yeah, but they start the server individually whereas Expert requires its own constellation of other servers available to operate (hence why we started as an application).

badosu: For now I just removed the full expert server test as I'm not fully confident it will work reliably (the application might already have been started on another test). I might send as a separate PR, let me know your thoughts.

As for using `GenLSP.Test.server` we'd require allowing it to receive options suitable for our server, for example see [buffer initialization hardcoded](https://github.com/elixir-tools/gen_lsp/blob/main/lib/gen_lsp/test.ex#L49).

badosu: ~~I reintroduced full server testing with a different approach.~~

And removed it again, it seems to not work as well regardless of the approach, might be better suited on a separate PR.

mhanberg: > As for using `GenLSP.Test.server` we'd require allowing it to receive options suitable for our server, for example see [buffer initialization hardcoded](https://github.com/elixir-tools/gen_lsp/blob/main/lib/gen_lsp/test.ex#L49).

I don't really understand what you are meaning here.

With the GenLSP test sdk, we should be able to start individual expert server instances and test them in isolation. I did it for basically every test in Next LS.

I think there are some small changes to make to make that possible as it stands right now, but can be accomplished later on.

apps/expert/lib/expert/engine_node.ex
```diff
@@ -103,6 +111,27 @@ defmodule Expert.EngineNode do
       end
     end
 
+    def on_exit_status(%__MODULE__{} = state, exit_status) do
+      stop_reason =
+        case exit_status do
+          0 ->
+            Logger.info("Engine shutdown")
```
mhanberg: Actually, sorry one more thing, we should also be logging the workspace (root_uri/directory whatever) that the engine belonged to. `Engine for #{root_dir_or_whatever} shut down.`

apps/expert/lib/expert.ex
```diff
@@ -185,20 +201,34 @@ defmodule Expert do
...
 
-      {:ok, config} = State.default_configuration(state)
-      {:noreply, assign(lsp, state: %{state | configuration: config})}
-    else
-      {:noreply, lsp}
-    end
+  def log_info(message) do
+    Logger.info(message)
+
+    GenLSP.log(get_lsp(), message)
```
mhanberg: I believe this should be GenLSP.info. GenLSP.log is the equivalent of Logger.debug

apps/expert/test/expert_test.exs
doorgan: @mhanberg I don't know if it's the same issue discussed here, but there's something very weird when we try to use `GenLSP.Test` in expert, ExUnit starts behaving very erratically, randomly skips test, sometimes doesn't even terminate after the test results are reported, etc. I left a comment about this in `engine_node_test.exs`.
The test ergonomics are seriously better with those helpers, but I don't know why it behaves like that in expert but not nextls.

mhanberg: I'll try and take another look at that when I have time. Probably during the winter break coming up.

badosu: > > As for using `GenLSP.Test.server` we'd require allowing it to receive options suitable for our server, for example see [buffer initialization hardcoded](https://github.com/elixir-tools/gen_lsp/blob/main/lib/gen_lsp/test.ex#L49).
> 
> I don't really understand what you are meaning here.

For example, our application starts our genlsp buffer with a name and specific opts: `{GenLSP.Buffer, [name: Expert.Buffer] ++ buffer_opts}`. Maybe I'm missing something, but this is moot until I can make the full server test reliable anyway.

> I think there are some small changes to make to make that possible as it stands right now, but can be accomplished later on.

Happy to hear if they seem straightforward enough.

mhanberg: > For example, our application starts our genlsp buffer with a name and specific opts: {GenLSP.Buffer, [name: Expert.Buffer] ++ buffer_opts}. Maybe I'm missing something, but this is moot until I can make the full server test reliable anyway.

The idea is you don't start/utilize the application, you start an instance of the server per test, each has their own buffer and tcp connection to ensure isolation.

When you start the server with the `server/2` function in the GenLSP test sdk, it starts a buffer and passes it to the options of your lsp process. The one started in the application supervision tree doesn't really matter.

You can see more [https://github.com/elixir-tools/next-ls/blob/main/test/support/utils.ex#L67]

we can look into this later, tho.

apps/expert/lib/expert/engine_node.ex
```diff
@@ -103,6 +111,28 @@ defmodule Expert.EngineNode do
...
+        case exit_status do
+          0 ->
+            project = state.project
+            Logger.info("Engine for #{project.root_uri} shut down")
+
+            :shutdown
+
+          _error_status ->
+            Logger.error(
+              "Engine shut down unexpectedly, node exited with status #{exit_status}). Last message: #{state.last_message}"
```
mhanberg, line 125: Add the root_uri here as well and any other log about the engine

apps/expert/test/expert_test.exs
badosu: > The idea is you don't start/utilize the application, you start an instance of the server per test, each has their own buffer and tcp connection to ensure isolation.

Oh, I completely understand that. In fact I had [this iteration](https://github.com/elixir-lang/expert/pull/244/commits/e381804dbd99fe0cbd0fbd685f13a0e5ad8c924d#diff-ffb240185cbe519a46509d31348498eafc361f28213b31cc7fc8b3bbe618b476) and one based on that that didn't start particularly Expert instead using `GenLSP.Test.server`.

The problem is that for now it seems to me that _the implementation of GenLSP.Test.server itself_ does a few things I'd rather have it not, or at least accept options for those. You can see for example the kind of `child_spec` I'd want for it to be able to initialize instead [here](https://github.com/elixir-lang/expert/blob/main/apps/expert/lib/expert/application.ex#L78-L82).

We can also vendor it ourselves anyway so this is not a big deal, or maybe even having the special names don't cause any issues.

Again this is a bit moot right now anyway so I won't digress further into it.

apps/expert/lib/expert/engine_node.ex
badosu: I added the project as a prefix but I'm not sure that's desired in the current incarnation. Let me know if [this message format](https://github.com/elixir-lang/expert/pull/244/commits/5ae550e0b548b32adffab7a78c399d0683f3549e#diff-ffb240185cbe519a46509d31348498eafc361f28213b31cc7fc8b3bbe618b476R20) looks good to you.

apps/expert/test/expert_test.exs
mhanberg: > Again this is a bit moot right now anyway so I won't digress further into it.

Discussion is fine, its great to get feedback
 
> We can also vendor it ourselves anyway so this is not a big deal,

I'm the author of GenLSP. We can change it however we want.

> The problem is that for now it seems to me that the implementation of GenLSP.Test.server itself does a few things I'd rather have it not, or at least accept options for those. You can see for example the kind of child_spec I'd want for it to be able to initialize instead [here](https://github.com/elixir-lang/expert/blob/main/apps/expert/lib/expert/application.ex#L78-L82).

I'm not really following this. It overrides the assigns, buffer, and task queue but passes through any other options.

Those 3 processes should be opaque to the consumer, so you really shouldn't be reading into them from a test anyway, but they are also available from in the return value from the server function.

Why might you want to start those yourself?

badosu: > Why might you want to start those yourself?

There might be a circumstance in which the name is explicitly used, for example at least once we explicitly use the `:expert_task_supervisor` name.

If you assess all is good then all is good :+1:

badosu:
Let me know if there are any pending issues to be addressed.

mhanberg (approved):
Looks good to me.

@doorgan let me know if you have any comments, if not lets merge.

doorgan:
Looks good to me too! I'll merge

PR #248, opened 2025-12-08
fix: elixir path discovery

### LINKED ISSUE #247: path_env_at_directory captures ANSI escape codes from shell logout scripts (zsh)
Expert fails to resolve the `elixir` executable on systems where the shell environment emits output during shell logout(maybe only zsh?) (e.g., standard `/etc/zlogout` scripts that run `clear`).

---
### The Problem, I think:

In `Expert.Port.path_env_at_directory/2`, the code attempts to resolve the project's `PATH` by spawning an interactive login shell:
[https://github.com/elixir-lang/expert/blob/main/apps/expert/lib/expert/port.ex#L82](https://github.com/elixir-lang/expert/blob/main/apps/expert/lib/expert/port.ex#L82)
```elixir
path
|> String.trim()
|> String.split("\n")
|> List.last() # <--- THE PROBLEM
```

---
### The Failure Mode:

1.  Expert runs `zsh -l -i -c "echo $PATH"`.
2.  The shell prints the **valid PATH** to stdout.
3.  Because `-l` (login) was passed, the shell sources **logout scripts** (e.g., `/etc/zlogout`) upon exiting.
4.  If the logout script contains `clear` (common in many distros/OS configurations), it emits **ANSI escape codes** (e.g., `\e[H\e[J`) to stdout **after** the `PATH`.
5.  `List.last/1` captures the escape codes instead of the `PATH` string.
6.  `os.find_executable` fails because the path is garbage.

---
### Steps to Reproduce:

1.  Use **zsh**.
2.  Add `clear` to `~/.zlogout` (simulating the system-level `/etc/zlogout` found on macOS/Arch/etc).
3.  Attempt to start Expert.
4.  Observe crash or failure to find `elixir`.

---
### Suggested Fixes:

1.  Do not rely on `List.last|first/1`. Filter the output to find the line containing the expected delimiters (e.g., `:`), as the `PATH` is printed before the shell teardown hooks run e.g. `Enum.find(&String.contains?(&1, "/"))` or something more robust
2.  Use a marker string and a regex match e.g.:
```elixir
# could likely do with just one marker... :)
marker_start = "__EXPERT_PATH_START__"
marker_end = "__EXPERT_PATH_END__"

# example command
"cd #{directory} && printf \"#{marker_start}:%s:#{marker_end}\" \"$PATH\""

# regex match on the output of the command
regex = ~r/#{marker_start}:(.*):#{marker_end}/s

case Regex.run(regex, output) do
  [_, clean_path] -> 
    clean_path
```

---
### A user-side fix
You can either comment out the `clear` lines from your `/etc/zlogout` or add a `clear` override function to your shell e.g.:
```bash
# Shadow the 'clear' command. 
# This persists until the shell closes, so /etc/zlogout will use THIS 
# instead of /usr/bin/clear.
clear() {
  # If we are in a pipe (Expert LSP), do nothing.
  if [[ ! -t 1 ]]; then
    return
  fi
  
  # If we are a real human, run the real clear binary.
  command clear "$@"
}
```

Not very... nice, but a workaround if anyone else has this problem.

soisyourface:
Closes #247

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -98,31 +98,6 @@ defmodule Expert.Port do
     end
   end
 
-  defp _path_env_at_directory(directory, shell) do
-    env = [{"SHELL_SESSIONS_DISABLE", "1"}]
-
-    path =
-      case Path.basename(shell) do
-        "fish" ->
-          {path, 0} =
-            System.cmd(shell, ["-l", "-c", "cd #{directory} && string join ':' $PATH"], env: env)
-
-          path
-
-        _ ->
-          {path, 0} =
-            System.cmd(shell, ["-i", "-l", "-c", "cd #{directory} && echo $PATH"], env: env)
-
-          Logger.error("this is the PATH: #{inspect(path)}")
-          path
-      end
-
-    path
-    |> String.trim()
-    |> String.split("\n")
-    |> List.last()
-  end
-
   @doc """
   Launches an executable in the project context via a port.
   """


apps/expert/lib/expert/port.ex
```diff
@@ -62,32 +64,38 @@ defmodule Expert.Port do
...
       end
 
-    path
-    |> String.trim()
-    |> String.split("\n")
-    |> List.last()
+    {output, _} = System.cmd(shell, args, env: env)
+
+    # This ignores banners (start) and logout garbage (end)
+    case Regex.run(~r/#{@path_marker}:(.*?):#{@path_marker}/s, output) do
```
doorgan, line 92: Should we expect the path string to start at the beginning of the line and end at the same line?
```suggestion
    case Regex.run(~r/^#{@path_marker}:(.*?):#{@path_marker}$/s, output) do
```

soisyourface: with the current implementation we aren't splitting on `\n` any more we are just looking for everything in between the markers. so, adding the boundaries `^`/`$` would cause this to not function.

we can, of course, split on `\n` and then iterate over each line looking for a pattern, but the marker solution seemed less error prone.

PR #254, opened 2025-12-11
feat: add engine subcommands to expert

### LINKED ISSUE #209: Clean local engine build
## Description

Expert uses Mix.install to build and install the engine. During development, if you make dependency changes to the engine, Mix.install won't know and it might run into dependency errors like in #200.

## Solution

I think that a reasonable solution for this is to include a subcommand in the expert binary for deleting/cleaning this builds

```sh
$ expert engine ls
/Users/mitchell/Library/Application Support/Expert/0.1.0
/Users/mitchell/Library/Application Support/Expert/0.1.1
/Users/mitchell/Library/Application Support/Expert/0.2.0

$ expert engine clean
Delete /Users/mitchell/Library/Application Support/Expert/0.1.0? [Yn]
Delete /Users/mitchell/Library/Application Support/Expert/0.1.1? [Yn]
Delete /Users/mitchell/Library/Application Support/Expert/0.2.0? [Yn]

$ expert engine clean --force
Deleted /Users/mitchell/Library/Application Support/Expert/0.1.0
Deleted /Users/mitchell/Library/Application Support/Expert/0.1.1
Deleted /Users/mitchell/Library/Application Support/Expert/0.2.0
```

rmand97:
closes #209 

Taking a stab at the issue. Hoping I understood it correctly.

### DIFF
--- a/apps/expert/lib/expert/application.ex
+++ b/apps/expert/lib/expert/application.ex
@@ -15,10 +15,22 @@ defmodule Expert.Application do
     argv = Burrito.Util.Args.argv()
 
     # Handle engine subcommand first (before starting the LSP server)
-    with ["engine" | engine_args] <- argv do
-      engine_args
-      |> Expert.Engine.run()
-      |> System.halt()
+    case argv do
+      ["engine" | engine_args] ->
+        engine_args
+        |> Expert.Engine.run()
+        |> System.halt()
+
+      [subcommand | _] ->
+        if not String.starts_with?(subcommand, "-") do
+          IO.puts(:stderr, """
+          Error: Unknown subcommand '#{subcommand}'
+
+          Run 'expert --help' for usage information.
+          """)
+
+          System.halt(1)
+        end
     end
 
     {opts, _argv, _invalid} =


apps/expert/lib/expert/application.ex
```diff
@@ -12,8 +12,19 @@ defmodule Expert.Application do
 
   @impl true
   def start(_type, _args) do
+    argv = Burrito.Util.Args.argv()
+
+    # Handle engine subcommand first (before starting the LSP server)
+    case argv do
+      ["engine" | engine_args] ->
+        Expert.Engine.run(engine_args)
```
mhanberg: Can you make this return the exit code, and put the `System.halt(exit_code)` part here?

That'll make the unit tests easier to write, as well as pull the halting part up to the top level. 

This way it's very clear that the program stops after this code runs.

rmand97: Will do. 
Just a quick question. Would you like it to always return success (`0`), event though it fails deleting a file?

like if its fails the middle one of 3 files in `expert clean --force` it will continue after failing, logging something went wrong, and delete the last one and return `0`.  Or would like it to stop after failing and return `1`?

mhanberg: If it fails any, lets return a non zero exit code (1).

apps/expert/lib/expert/engine.ex
```diff
@@ -0,0 +1,163 @@
...
+      _ -> false
+    end
+  end
+
+  defp print_location_info do
+    IO.puts("\nEngine builds are stored in: #{base_dir()}")
+  end
+
+  @spec print_help() :: no_return()
+  defp print_help do
```
mhanberg, line 205: Let's also add help text to each subcommand, and the options can be presented for the relevant subcommand.

So in the existing help text here, we won't have any options, and the  `--force` option will show up in the help text for the `clean` subcommand

mhanberg:
@rmand97 looks like there are some failing tests.

apps/expert/lib/expert/application.ex
```diff
@@ -12,8 +12,21 @@ defmodule Expert.Application do
...
+    # Handle engine subcommand first (before starting the LSP server)
+    case argv do
+      ["engine" | engine_args] ->
+        engine_args
+        |> Expert.Engine.run()
+        |> System.halt()
+
+      _ ->
+        :noop
+    end
```
mhanberg, line 34:
```suggestion
    with ["engine" | engine_args] <- argv do
      engine_args
      |> Expert.Engine.run()
      |> System.halt()
    end
```

apps/expert/lib/expert/engine.ex
```diff
@@ -0,0 +1,224 @@
...
+
+  @help_options ["-h", "--help"]
+
+  @spec run([String.t()]) :: non_neg_integer()
+  def run(args) do
+    {opts, subcommand, _invalid} =
+      OptionParser.parse(args,
+        strict: [force: :boolean],
+        aliases: [f: :force]
+      )
```
mhanberg: I think at this level, we would only want a help option.

apps/expert/lib/expert/engine.ex
```diff
@@ -0,0 +1,224 @@
...
+
+    case subcommand do
+      ["ls"] -> list_engines()
+      ["ls", options] when options in @help_options -> print_ls_help()
```
mhanberg: I believe the way to accomplish the "subcommand" cli we are wanting is to parse the options, get the subcommand, then parse the options again.

Also not we use parse_head instead of parse.

So for example

```elixir
{opts, rest, _invalid} =
      OptionParser.parse_head(args,
        strict: [help: :boolean],
        aliases: [h: :help]
      )

case rest do
  ["clean" | clean_opts] -> clean_engines(clean_opts) 
  # the rest
end

defp clean_engines(argv) do
  {opts, _rest, _invalid} =
      OptionParser.parse_head(argv,
        strict: [force: :boolean, help: :boolean],
        aliases: [f: :force, h: :help]
      )

  # do the rest

end
```

apps/expert/lib/expert/engine.ex
```diff
@@ -0,0 +1,250 @@
...
+        strict: [help: :boolean],
+        aliases: [h: :help]
+      )
+
+    cond do
+      opts[:help] ->
+        print_help()
+
+      true ->
+        case subcommand do
```
mhanberg: I think we want a fallback case to print an "unknown sub command" message.

mhanberg:
❤️

rmand97:
Glad to see you could use it! Keep up the awesome work!

PR #257, opened 2025-12-13
chore: improve plain releases for local builds

doorgan:
Updates the `just release-plain` command to include a `start_expert`script that takes care of wrapping the expert executable and pass the CLI flags correctly.

Also updates the docs nudging people to use plain releases when building expert from source.

The main idea here is that we make it easier and faster for users building expert from source (eg when developing) and reserve the cross platform burrito executables for the official releases that regular users download.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,7 +172,7 @@ jobs:
             ${{ runner.os }}-mix-${{ matrix.elixir }}-${{ matrix.otp }}-
 
       - name: Build and release
-        run: just release-local
+        run: just burrito-local
 
   prep-matrix:
     runs-on: ubuntu-latest
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -26,7 +26,7 @@ jobs:
       - run: just deps engine
       - name: Set release version to latest rev
         run: sed -i "$ s/$/-$(git rev-parse --short HEAD)/" version.txt
-      - run: just release-all
+      - run: just burrito
         env:
           MIX_ENV: prod
       - name: Create Checksum
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ jobs:
         with:
           version: "0.15.2"
       - uses: actions/checkout@v6
-      - run: just release-all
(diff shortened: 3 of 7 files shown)


apps/expert/lib/expert/release.ex
```diff
@@ -59,4 +59,43 @@ defmodule Expert.Release do
...
+      You can also run Expert in TCP mode by passing the `--port PORT` argument:
+
+          #{cmd} --port 9000
+
+      To get a list of all available command line options, run:
+
+          #{cmd} --help
+      """)
+
+      # Silence the release "announce" message
```
doorgan, line 90: I didn't like the default announcement message because it's misleading, you can't run expert with them *and* you can't pass cli flags to `start` or `start_iex`, so figured this way to replace it with a custom one.

apps/expert/lib/expert/release.ex
```diff
@@ -59,4 +59,43 @@ defmodule Expert.Release do
...
+    else
+      cmd = Path.relative_to(executable_path, File.cwd!())
+
+      Mix.shell().info("""
+
+      #{IO.ANSI.bright()}✨ Expert build created at:#{IO.ANSI.reset()} #{release.path}
+
+      To use it, point your editor LSP configuration to:
+
+          #{executable_path} --stdio
```
mhanberg, line 79: I think we want all of these to show the full executable_path, for consistency.

apps/expert/rel/expert.bat
```diff
@@ -0,0 +1,5 @@
+@echo off
```
mhanberg: why do we have this file and the one in overlays?

mhanberg:
This would be a chore, not a feat

I think maybe it should be `just release` which does the plain release, `just burrito` does a burrito build for all platforms, and `just burrito-local` does a burrito of your local arch/os.

Will require changing the CD workflows

apps/expert/rel/expert.bat
doorgan: Agh, there's a bug in the filetree plugin I use on neovim that would refuse to delete this file every time I moved it or renamed it. I forgot to remove this one *again* before committing

mhanberg (approved):

PR #260, opened 2025-12-16
refactor: make progress tracking better follow LSP semantics

Moosieus:
This pull request overhauls the way progress reporting is handled in Expert. The top-level API remains largely the same for callers.

The prior architecture started a per-project stateful GenServer that handled progress tracking logic internally:
* It was difficult to reason about due to the depth and statefulness of the call stack.
    * Or at least, I found it difficult. YMMV, but my brain works better with shallow call stacks.
* The Progress GenServers were initiated per-project, even though workDoneTokens aren’t scoped as such in LSP.
* Had no means to handle client-initiated workDoneTokens
    * I had a `with_client_progress` function incorporated here, but I removed it for the sake of review. There’s already a lot here.
* Had no means to deallocate workDoneTokens should they be cancel-able.
    * Requests for long-running tasks can be canceled by the client, in which case it cancels the request by ID, but not any workDoneTokens associated.
        * Hence why the work-done-tokens now live in the caller’s context.

The new progress reporting modules are organized as follows:
* `Forge.Progress` - a behaviour that defines the `with_progress` and `with_tracked_progress` helper functions, permitting the implementing module provides the following functions:
    * `begin/2`
    * `report/2`
    * `complete/2`
* `Expert.Progress` - a process-less module that directly handles progress reporting, implements the aforementioned
* `Engine.Progress` - a thin wrapper that calls `Expert.Progress` via erpc.
* `Forge.Progress.Tracker` - An ephemeral GenServer for reporting progress between concurrently running tasks.
    * Supports straight percentage reporting by default, or custom callbacks like `turbo-encabulating files: 3/25` should that be preferable.

All in all: Less processes, more sequential Elixir, more flexibility, better alignment with LSP semantics.

Some additional changes:
* Make tests that wait the `project_compiled` message less flaky by increasing the timeout from 100ms -> 5s.
* Debounce namespacing logs
* When the engine starts, store its manager name in a persistent term and use that, instead of building it.

### DIFF
--- a/apps/engine/lib/engine/build/project.ex
+++ b/apps/engine/lib/engine/build/project.ex
@@ -1,9 +1,9 @@
 defmodule Engine.Build.Project do
-  alias Forge.Project
   alias Engine.Build
+  alias Engine.Build.Isolation
   alias Engine.Plugin
   alias Engine.Progress
-  alias Engine.Build.Isolation
+  alias Forge.Project
   alias Mix.Task.Compiler.Diagnostic
 
   require Logger


apps/expert/lib/expert.ex
```diff
@@ -57,6 +57,8 @@ defmodule Expert do
     with {:ok, response, state} <- State.initialize(state, request),
          {:ok, response} <- Expert.Protocol.Convert.to_lsp(response) do
       Task.Supervisor.start_child(:expert_task_queue, fn ->
+        # dirty sleep to allow initialize response to return before progress reports
+        Process.sleep(50)
```
badosu, line 61: Hmm.. Perhaps it would be an interesting addition for gen_lsp to support something like:

```ex
{:reply, reply, {:continue, term, state}} |
{:noreply, {:continue, term, state}}
```

Then support a `handle_continue` callback?

apps/expert/lib/expert/engine_node.ex
```diff
@@ -250,18 +252,22 @@ defmodule Expert.EngineNode do
 
           launcher = Expert.Port.path()
 
+          Logger.info("Finding or building engine for project #{project_name}")
```
badosu: You can replace both this and line below with (something like):

```elixir
Expert.log_info(lsp, "Finding or building engine")
```

The project "name" is prepended to the message already.

apps/expert/lib/expert.ex
Moosieus: Matching `GenServer` semantics? Yeah, that'd be sweet.

The key to resolving the timing issue here is to make sure the language-client gets the initialize response, before the language-server initializes anymore workDoneTokens.

badosu: Yeah, that would be the idea. So that we can support a situation like this where we want to perform work on a continuation after the current work is done, with _some_ guarantee the reply was sent - although there might be other valid reasons to want to do a continuation I guess. E.g.:

```elixir
def handle_request(req, lsp) do
  reply = get_reply(req)
  {:reply, reply, {:continue, :initialize_engine, lsp}}
end

def handle_continue(:initialize_engine, lsp) do
  # maybe something related but we want to offload that particular loop iteration
  # maybe something unrelated and we want to separate scope
  # maybe something that should run in some continuous "work" manner
  # maybe we wanted some guarantee the reply was sent
  {:noreply, lsp}
end
```

Anyway, not an issue for this particular PR, I wondered if you'd tell me if this sounds like a reasonable "solution" for the sleep issue.

apps/expert/lib/expert/engine_node.ex
```diff
@@ -221,7 +223,7 @@ defmodule Expert.EngineNode do
 
       case Expert.Port.elixir_executable(project) do
         {:ok, elixir, env} ->
-          GenLSP.info(lsp, "Found elixir for #{project_name} at #{elixir}")
+          Expert.log_info(lsp, "Finding or building engine at #{elixir}")
```
badosu: `elixir` here is the path of the executable. So maybe something like `"Found elixir executable at #{elixir}"`?

Moosieus: Good looking out. Gaffe on my behalf fixed.

apps/expert/lib/expert.ex
Moosieus: I took a crack at this:  https://github.com/Moosieus/gen_lsp/tree/moo/handle-continue

Not tested in-anger against Expert at time of writing, though.

badosu: Looks great!

apps/engine/lib/engine/compilation/tracer.ex
```diff
@@ -56,10 +57,9 @@ defmodule Engine.Compilation.Tracer do
   end
 
   defp maybe_report_progress(file) do
-    if Path.extname(file) == ".ex" do
-      file
-      |> progress_message()
-      |> Engine.broadcast()
+    with ".ex" <- Path.extname(file),
```
badosu, line 60: Don't mind my ignorance, are `.ex` files the only ones the engine might compile?

Moosieus: If I'm following the semantics right, yeah. For example, you wouldn't compile a `.heex` file directly, you'd compile an `.ex` source file that invokes [HEEx's tag engine](https://github.com/phoenixframework/phoenix_live_view/blob/v1.1.19/lib/phoenix_live_view/html_engine.ex#L1). Macros all the way down and whatnot.

apps/engine/lib/engine/build/project.ex
```diff
@@ -1,51 +1,60 @@
 defmodule Engine.Build.Project do
   alias Forge.Project
-
-  alias Engine.Build
+  alias Engine.{Build, Plugin, Progress}
```
doorgan: Nit
```suggestion
  alias Engine.Build
  alias Engine.Plugin
  alias Engine.Progress
```

apps/engine/lib/engine/build.ex
```diff
@@ -1,9 +1,7 @@
 defmodule Engine.Build do
-  alias Forge.Document
-  alias Forge.Project
-
   alias Engine.Build.Document.Compilers.HEEx
   alias Engine.Build.State
+  alias Forge.{Document, Project}
```
doorgan:
```suggestion
  alias Forge.Document
  alias Forge.Project
```

apps/expert/lib/expert/progress.ex
```diff
@@ -0,0 +1,140 @@
+defmodule Expert.Progress do
+  @moduledoc """
+  LSP progress reporting for the Expert language server.
+  """
+
+  use Forge.Progress
+
+  alias Expert.Configuration
+  alias Expert.Protocol.Id
+  alias GenLSP.{Notifications, Requests, Structures}
```
doorgan:
```suggestion
  alias GenLSP.Notifications
  alias GenLSP.Requests
  alias GenLSP.Structures
```

apps/expert/test/engine/build_test.exs
```diff
@@ -76,9 +77,7 @@ defmodule Engine.BuildTest do
       {:ok, project} = with_project(:project_metadata)
       EngineApi.schedule_compile(project, true)
 
-      assert_receive project_compiled(status: :success)
-      assert_receive project_progress(label: "Building " <> project_name)
```
doorgan, line 80: Is this event no longer fired here?

apps/forge/lib/forge/progress/tracker.ex
```diff
@@ -0,0 +1,67 @@
...
+
+  @doc """
+  Starts a tracker process.
+
+  ## Options
+
+  - `:token` - The progress token (required)
+  - `:total` - The total value representing 100% (required)
+  - `:report_fn` - Callback invoked on each update (required)
+    Signature: `(message, current, total, token) -> any()`
```
doorgan, line 26: We can also document this with typespecs but I'm not sure it really matters.

apps/forge/lib/forge/progress.ex
```diff
@@ -0,0 +1,130 @@
...
+      Uses a default callback that reports percentage-based progress.
+      """
+      def with_tracked_progress(title, total, work_fn) when is_function(work_fn, 1) do
+        with_tracked_progress(title, total, work_fn, &default_report/4)
+      end
+
+      def with_tracked_progress(title, total, work_fn, report_fn)
+          when is_function(work_fn, 1) and is_function(report_fn, 4) do
+        run_with_progress(title, [percentage: 0], fn token ->
+          {:ok, tracker} = Tracker.start_link(token: token, total: total, report_fn: report_fn)
```
doorgan, line 77: Should we use a DynamicSupervisor for this?

apps/forge/lib/forge/progress.ex
```diff
@@ -0,0 +1,130 @@
...
+
+      defp execute_work(token, work_fn) do
+        try do
+          token |> work_fn.() |> complete_with(token)
+        rescue
+          e ->
+            complete(token, message: "Error: #{Exception.message(e)}")
+            reraise e, __STACKTRACE__
+        end
+      end
```
doorgan, line 100: `def`/`defp` use an implicit `try`, so I think this works too:
```suggestion
      defp execute_work(token, work_fn) do
        token |> work_fn.() |> complete_with(token)
      rescue
        e ->
          complete(token, message: "Error: #{Exception.message(e)}")
          reraise e, __STACKTRACE__
      end
```

apps/forge/lib/forge/progress/tracker.ex
Moosieus: I'm leaning away from typespecs and more into the type system even if aspirationally, gradual-gradual type system and all that. I prefer comments like the above b/c they tell the semantic meaning directly, instead of needing to be inferred by the type's structure or the type's name.

apps/expert/test/engine/build_test.exs
Moosieus: Check the diff - assertion for `project_compiled` was kept, just with a more lenient timer to fix flaky tests.

doorgan: I mean `project_progress`, but now I realize that message was replaced by a different mechanism, so disregard my comment

apps/forge/lib/forge/progress.ex
Moosieus: I don't think it'd add much.

Moosieus:
> For my own understanding: we are moving from the engine node sending progress messages to the server to instead unifying everything under the new Progress API(so, we report progress directly to the LSP via erpc calls to the server, instead of sending progress to the server and having it create the messages). Is this correct?

That's correct. Tried to make it as simple as can be w/o being simplistic.

doorgan (approved):
@Moosieus got it, I'm on board with these changes then!

PR #264, opened 2025-12-17
docs: add --stdio flag on emacs configuration documentation

rockneurotiko:
Now that `expert` requires `--stdio` (or `--port`), this PR changes the documentation for emacs to include that option.

### DIFF
--- a/pages/installation.md
+++ b/pages/installation.md
@@ -112,7 +112,7 @@ emacs configuration), insert the following code:
 (use-package elixir-mode
   :ensure t
   :custom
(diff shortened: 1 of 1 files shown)


mhanberg:
If you're interested you could also add the emacs docs to the website https://github.com/expert-lsp/expert-lsp.org/blob/main/_pages/docs/editors.md

rockneurotiko:
Oh, I didn't know about that, I'll add it there too :)

mhanberg:
> Oh, I didn't know about that, I'll add it there too :)

It's fairly new, no worries

PR #268, opened 2025-12-19
fix(release): don't cd into rel directory before starting app

mhanberg:
Changing the directory before changes the cwd as seen by expert, so the
log file is placed in the wrong spot.

I'm not actually sure if the windows script works, but we can have
someone test it.

### DIFF
--- a/apps/expert/rel/overlays/bin/start_expert.bat
+++ b/apps/expert/rel/overlays/bin/start_expert.bat
@@ -1,4 +1,4 @@
 @echo off
 
-"%~dp0plain.bat" eval "System.no_halt(true); Application.ensure_all_started(:xp_expert)" --stdio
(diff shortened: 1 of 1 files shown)


apps/expert/rel/overlays/bin/start_expert.bat
```diff
@@ -1,5 +1,4 @@
 @echo off
 
-cd /d "%~dp0"
-.\plain.bat eval "System.no_halt(true); Application.ensure_all_started(:xp_expert)" --stdio
+"%~dp0plain.bat" eval "System.no_halt(true); Application.ensure_all_started(:xp_expert)" --stdio
```
doorgan: I forgot to pass the args in the windows file instead of hardcoding `--stdio`
```suggestion
"%~dp0plain.bat" eval "System.no_halt(true); Application.ensure_all_started(:xp_expert)" %*
```

PR #272, opened 2025-12-20
fix: support Nushell for PATH detection

leonqadirie:
Adds Nushell (`nu`) support to `path_env_at_directory/2` in `port.ex`.
Specifically: Adds a `"nu"` case in `path_env_at_directory/2` that uses `;` for command chaining and `$env.PATH | str join ":"` to convert the PATH list to a colon-separated string.

---
Nushell differs from POSIX shells in two ways:
- The `&&` operator is not supported (requires `;` or `and`)
- PATH is stored as a list in `$env.PATH`, not a colon-separated string

Currently, Expert crashes immediately when `SHELL` is set to Nushell:

```
Error: nu::parser::shell_andand

  x The '&&' operator is not supported in Nushell
   ,-[source:1:39]
 1 | cd /path/to/project && printf "EXPERT_PATH:%s:EXPERT_PATH" "$PATH"
   :                     ^|
   :                      `-- instead of '&&', use ';' or 'and'
```

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -107,12 +107,12 @@ defmodule Expert.Port do
           ["-l", "-c", cmd]
 
         "nu" ->
-            # Nushell stores PATH as a list in $env.PATH, so we join with colons.
-            # Nushell doesn't support && operator, use ; instead.
-            cmd =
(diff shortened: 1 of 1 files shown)


mhanberg:
Looks good, just fix up the formatter errors

PR #278, opened 2025-12-23
feat(engine): support shorthand notation inside ~H sigil

### LINKED ISSUE #269: Jump to definition in HEEx templates
At @Jump-App one of the most longed for features in Expert was ability to jump to definition in HEEx templates, when clicking on components. This works in cases like this:

```elixir
~H"""
<MyComponents.my_button>Submit</MyComponents.my_button>
"""
```

... but not when using a "shorthand notation", i.e.:

```elixir
import MyComponents

~H"""
<.my_button>Submit</.my_button>
"""
```

The reason why the former works is IIRC because ElixirSense interprets `MyComponenets.my_button` correctly as a function call, but obviously `.my_button` cannot be interpreted like a correct Elixir code.

I made a change to Expert that adds a "normalization" step for contents of `~H` sigil, which detects the shorthand notation and replaces it as `my_button(assigns)` before sending to ElixirSense. You can see the change [here](https://github.com/katafrakt/expert/pull/4) (this is a PR against my fork). We released Expert binaries with this change internally and several people reported go to definition working just fine with it.

Now, we would like to upstream it. But I do know that this is a bit of a hacky change, so I want to discuss it first. Perhaps there is a better way to explore? Perhaps compiling contents of the sigil `Phoenix.LiveView.TagEngine` would be a better approach, but then it would require some kind of plugin system, to only activate it when LiveView is actually present.

Do you have any thoughts on this?

katafrakt:
Closes #269 

The full-module notation (`<MyMod.button>`) was already supported in "go to definition", because when passed to ElixirSense, it correctly recognized the form and found the appropriate module and the function. However, with a shorthand notation it was not so simple, because `.button` is not valid Elixir and ElixirSense was not able to make anything of it.

This implements the support for shorthand notation by doing a preliminary step before sending the code to ElixirSense. It modifies the AST and the Document of ForgeAnalysis and replaces all the calls to `<.button>` with `< button(assigns)` so that ElixirSense can correctly interpret it as a local function call with arity 1. This only happens when `phoenix_live_view` is added as a dependency[^1].

This works for functions defined in the same module, but also for imported functions. Support for ending tag is also included.

While this might seem as a hacky solution, it makes handling shorthand notation as close as possible to handling the full-module notation, making sure these two stay conceptually close.

[^1]: I looked into making a more detailed check here, basically checking if `Phoenix.Component` is imported into current module, but there's just too many ways to do so. I don't think the code analysis can do that, we would need to actually compile the code (using ElixirSense?). But I settled for a much simpler condition in this case.

### DIFF
--- a/apps/forge/test/fixtures/navigations/lib/component.ex
+++ b/apps/forge/test/fixtures/navigations/lib/component.ex
@@ -1,0 +1,9 @@
+defmodule Component do
+  defmacro __using__(_) do
+    quote do
+      import(unquote(__MODULE__))
+    end
+  end
+
+  def sigil_H(_, _), do: nil
+end
--- a/apps/forge/test/fixtures/navigations/lib/my_components.ex
+++ b/apps/forge/test/fixtures/navigations/lib/my_components.ex
@@ -4,7 +4,7 @@ defmodule MyComponents do
   This module provides components that can be used with shorthand notation.
   """
 
-  use Phoenix.Component
+  use Component
 
   @doc """
   A simple button component.
--- a/apps/forge/test/fixtures/navigations/lib/my_live_view.ex
+++ b/apps/forge/test/fixtures/navigations/lib/my_live_view.ex
@@ -1,5 +1,5 @@
 defmodule Navigations.MyLiveView do
-  use Phoenix.Component
+  use Component
   import MyComponents
 
   def render(assigns) do


apps/engine/lib/engine/code_intelligence/heex_normalizer.ex
```diff
@@ -0,0 +1,172 @@
...
+      new_ast = normalize_ast(analysis, position)
+      new_document = normalize_document(analysis, position)
+      %{analysis | ast: new_ast, document: new_document}
+    else
+      analysis
+    end
+  end
+
+  defp phoenix_component_available? do
+    Code.ensure_loaded?(Phoenix.Component)
```
doorgan: We probably want `Engine.Module.Loader.ensure_loaded?` here. `Code.ensure_loaded?/1` can be quite slow, so we have a cache for it
```suggestion
    Engine.Module.Loader.ensure_loaded?(Phoenix.Component)
```

apps/engine/lib/engine/code_intelligence/heex_normalizer.ex
```diff
@@ -0,0 +1,172 @@
...
+  # It's done in both the AST and document text.
+  #
+  # This allows ElixirSense to understand the shorthand HEEX notation as a local function
+  # (be it imported or not) and return correct location for go-to-definition and hover.
+  #
+  # This normalization is only performed when Phoenix.Component is available in the project
+  # (i.e., phoenix_live_view is in the dependencies).
+  @spec maybe_normalize(Analysis.t(), Position.t()) :: Analysis.t()
+  def maybe_normalize(analysis, position) do
+    if phoenix_component_available?() do
```
doorgan, line 30: If possible, I think we also need to check in the `analysis` scopes if `sigil_H` from `Phoenix.Component` is imported

apps/engine/lib/engine/code_intelligence/heex_normalizer.ex
```diff
@@ -0,0 +1,172 @@
...
+    with {:ok, path} <- Ast.path_at(analysis, position),
+         {:sigil_H, _, _} = sigil <- Enum.find(path, &match?({:sigil_H, _, _}, &1)) do
+      new_sigil = normalize_heex_node(sigil)
+
+      analysis.ast
+      |> Zipper.zip()
+      |> Zipper.find(&(&1 == sigil))
+      |> case do
+        nil -> analysis.ast
+        zipper -> zipper |> Zipper.replace(new_sigil) |> Zipper.root()
```
doorgan: In case we need to speed this up for large files, there's also `Sourceror.FastZipper`

apps/engine/lib/engine/code_intelligence/heex_normalizer.ex
katafrakt: done, thanks

apps/engine/lib/engine/code_intelligence/heex_normalizer.ex
katafrakt: changed to `FastZipper`

katafrakt:
@doorgan thank you for the review!

Admittedly, I didn't really focus on Hover while working on this. I can have a look, if you think this is in the scope here.

As for error logs, I haven't really seen these errors yet, but I would like to look into this. Where do they appear for you? In the `project.log`?

doorgan:
@katafrakt
> Admittedly, I didn't really focus on Hover while working on this. I can have a look, if you think this is in the scope here.

I think it's in scope for this feature, but also feel free to do it in a follow up. If this gets merged and you decide to tackle that in another PR, let's open an issue to track it.

> As for error logs, I haven't really seen these errors yet, but I would like to look into this. Where do they appear for you? In the project.log?

They appear in expert.log
But again this may be an ElixirSense quirk, like it happens sometimes for module attributes

I think this is good to go, but I'll defer to @mhanberg on merging. I want to make sure we're good with this approach in the short term vs revamping the plugin system first.

apps/engine/lib/engine/code_intelligence/heex_normalizer.ex
katafrakt: I tried to do that, but I'm not sure if this can be done reliably. For example, I find it hard to detect things like `use MyAppWeb, :live_component`, which imports `sigil_H`, not to mention some potential more elaborate metaprogramming resulting in an import.

katafrakt:
I'm trying to understand if there's something special about this OTP 27.3.4.1 / Elixir 1.18.4 version combination, where tests for this fail on a seemingly unrelated test case, given that is passes with this elixir version and higher/lower OTP. I'm able to reproduce this locally, so it's not a CI issue.

UPDATE: I traced it to the `ElixirSense.definition` call in `Engine.CodeIntelligence.Definition`, so I think it's a problem there.

mhanberg:
I'll dig into this this week.

In general, I think we already have special code for heex templates, so probably fine to continue doing so for now.

apps/engine/lib/engine/code_intelligence/heex_normalizer.ex
mhanberg: Once we integrate spitfire, we should be able to detect any aliases and imports from use macros and metaprogramming. it can get you the current environment for a cursor location (imported functions, aliased modules, etc)

mhanberg:
@doorgan merge at your convenience

doorgan:
Let's fix(or workaround if it comes from ElixirSense) the failing test and we can merge

katafrakt:
@doorgan the test is fixed now

doorgan:
Thank you! 💜

PR #279, opened 2025-12-24
chore: speed up engine builds with incremental namespacing

doorgan:
Drastically improves Engine load times by incrementally namespacing engine files

Rather than complicating the transformer modules to make them aware of which files should be touched and which shouldn't, we:

1. Figure out which files are new, change and which should be deleted from the output directory
2. Delete the files that are flagged for deletion in the output folder
3. Copy the new+changed files to a temporary folder
4. Namespace that folder
5. Copy the contents of the temp folder to the output folder

This achieves the same with the slight overhead of creating and cleaning up a temporary folder.

In my test on my m2 macbook I got the times down from ~10 seconds to 800-1200ms.

### DIFF
--- a/apps/expert/lib/expert.ex
+++ b/apps/expert/lib/expert.ex
@@ -25,10 +25,7 @@ defmodule Expert do
 
   @dialyzer {:nowarn_function, apply_to_state: 2}
 
-  @external_resource "../../version.txt"
-  @version "../../version.txt" |> File.read!() |> String.trim()
-
-  def vsn, do: @version
+  def vsn, do: :expert |> Application.spec(:vsn) |> to_string()
 
   def get_lsp, do: :persistent_term.get(:expert_lsp, nil)
 


apps/expert/lib/expert/engine_node.ex
```diff
@@ -89,6 +89,7 @@ defmodule Expert.EngineNode do
 
             {:error, reason} ->
               IO.puts("error starting node: #{inspect(reason)}")
+              IO.puts("error starting node:\n #{inspect(reason)}")
```
Moosieus: nit: duplicate

apps/forge/lib/mix/tasks/namespace.ex
```diff
@@ -30,28 +30,69 @@ defmodule Mix.Tasks.Namespace do
...
+
+    entries_to_namespace =
+      classification.new ++ classification.changed
+
+    Mix.Shell.IO.info("""
+    Namespacing #{length(entries_to_namespace)} files:
+      New: #{length(classification.new)}
+      Changed: #{length(classification.changed)}
+      Deleted: #{length(classification.deleted)}
+    """)
```
Moosieus: nit: move this to a `log_entries_to_namespace(classification)` function

Moosieus:
I wonder if Briefly is handling Windows' file system properly.

doorgan:
It does! It creates something like `C:\Users\Dorgan\AppData\Local\Temp/briefly--5764607095903232-YiSQwcVeeltDoQnC53r7/briefly-5764606969264128-2dncVrKcfGsPTZ2JmgPO
`

Moosieus:
Maybe a file permission issue? I'm grasping at straws.

This PR reads solid otherwise.

mhanberg:
Question, we don't need to namespace everytime during a normal release, correct? meaning you downloaded expert via a package manager, github releases, or are building a specific release from source.

I think in that case, we don't need to namespace every time? Only the first time.

In the case that we are working in development, we can maybe have an env var or just use mix env dev to namespace every time.

I'll still review this tho.

doorgan:
@mhanberg we need to do that for nightly since the engine builds are "namespaced" by the expert version(which right now is always 0.1.0)
Unless we append the nightly commit hash to the version?

Even if we do that it's still beneficial to do incremental namespacing, it cuts down iteration time while developing

mhanberg:
Nightly should already have the commit hash appended to it. 

It's locally compiled releases that only have '0.1.0'

doorgan:
Ah, I didn't realize we do that to `version.txt` for nightly, that does simplify things. Changing `Expert.vsn/0` to read that file instead of the mix config does the trick for published releases

mhanberg (approved):

apps/expert/lib/expert.ex
```diff
@@ -25,7 +25,8 @@ defmodule Expert do
 
   @dialyzer {:nowarn_function, apply_to_state: 2}
 
-  @version Mix.Project.config()[:version]
+  @external_resource "../../version.txt"
+  @version "../../version.txt" |> File.read!() |> String.trim()
```
mhanberg: I think this can read from `Application.spec(:expert, :vsn)` at runtime instead of reading from the file at compile time.

PR #280, opened 2025-12-27
fix(expert): print to stderr when no transport argument provided

badosu:
This should fix initialization error not present in lsp logs for neovim, as stderr seems to be logged (and stdout ignored).


mhanberg:
❤️

PR #282, opened 2025-12-31
fix(expert): save new configuration after `workspace/didChangeConfiguration`

katafrakt:
State handler of workspace/didChangeConfiguration message was always returning old configuration, so the changes to dialyzer support were not preserved.

This is maybe not super important, as the dialyzer config does not seem to be used anywhere, but it gets the general `didChangeConfiguration` flow right for the future - and I would like to add `projectDir` change handled by `didChangeConfiguration`, as this is default way to tweak the setting per-project in Emacs.

### DIFF
--- a/apps/expert/lib/expert/state.ex
+++ b/apps/expert/lib/expert/state.ex
@@ -84,8 +84,6 @@ defmodule Expert.State do
         GenLSP.request(Expert.get_lsp(), request)
         {:ok, %__MODULE__{state | configuration: config}}
     end
-
-    {:ok, state}
   end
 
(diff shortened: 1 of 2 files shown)


mhanberg:
Can you clarify what you are seeing a bit more? We don't support dialyzer at all nor a projectDir configuration option.

katafrakt:
@mhanberg I'm not actually _seeing_ anything as such. I wanted to see if I can add support for dynamically setting `projectDir` via `didChangeConfiguration` (as otherwise I have problems with monorepo dir structure in Emacs), but found this on the way and decided to make a separate PR - because it does look like a bug in the code.

However, if you think it does not make sense or should be included in a potential larger PR, I'm fine with closing this one.

PR #284, opened 2026-01-04
feat: add module rename support

scottming:
### Context

While AI-assisted renaming is convenient, it's still slower than a native Language Server implementation. This PR ports the module rename feature from [Lexical PR #636](https://github.com/lexical-lsp/lexical/pull/636), which I've been using reliably for almost two years. Given that the Lexical repository is now archived, I wanted to bring this practical and well-tested feature to Expert.

### Changes

**Core Implementation (engine)**
- Add `Engine.CodeMod.Rename` module with submodules: `Prepare`, `Module`, `Entry`, `File`, and `Diff`
- Add `Engine.Commands.Rename` GenServer for tracking rename progress and triggering reindex
- Add `Engine.Commands.RenameSupervisor` for managing rename command lifecycle

**LSP Integration (expert)**
- Add `PrepareRename` and `Rename` handlers for LSP protocol
- Extend `Expert.EngineApi` with `prepare_rename/3` and `rename/5` APIs
- Register `rename_provider` with `prepare_provider: true` in server capabilities

**Infrastructure (forge)**
- Add `RenameFile` struct to `Forge.Document.Changes` for file rename operations
- Add `file_changed` and `file_saved` message types to `Forge.EngineApi.Messages`

**Key Differences from Lexical**
- Uses `:erpc.call` via `Expert.EngineApi.call/4` instead of `RemoteControl.call`
- Added reindexing mechanism after rename to keep search index consistent (see "Why These Changes")
- Namespace mapping: `Lexical.RemoteControl.*` → `Engine.*`, `Lexical.Server.*` → `Expert.*`

### Why These Changes

- **Progress tracking & reindex** (`59953157`): Without reindexing after rename, the search index retains old module names. This causes subsequent renames to fail - e.g., renaming `Accounts` → `Accounts2` works, but renaming back fails because the index doesn't reflect the new names.

- **Temporary document opening** (`f08fd155`): After a file rename, the new URI may not exist in `Document.Store`, causing "not_open" errors during reindex. Added `ensure_open/1` to temporarily open files from disk when needed.

- **Expanded test coverage** (`a38652ef`): Added comprehensive tests covering references, descendants, structs, edge cases, and file renaming with realistic module names.

### Extra

A demo video is attached showing the smooth rename workflow - renaming a module and then renaming it back. Tested successfully in both VSCode and Neovim.

https://github.com/user-attachments/assets/7f7bd67b-8cef-47b3-90f2-93146e534a90

### DIFF
--- a/apps/engine/lib/engine.ex
+++ b/apps/engine/lib/engine.ex
@@ -71,6 +71,10 @@ defmodule Engine do
 
   defdelegate runtime_versions, to: Forge.VM.Versions, as: :current
 
+  defdelegate prepare_rename(analysis, position), to: Engine.CodeMod.Rename, as: :prepare
+
+  defdelegate rename(analysis, position, new_name, client_name), to: Engine.CodeMod.Rename
+
   def list_apps do
     for {app, _, _} <- :application.loaded_applications(),
         not Forge.Namespace.Module.prefixed?(app),
--- a/apps/engine/lib/engine.ex
+++ b/apps/engine/lib/engine.ex
@@ -69,8 +69,6 @@ defmodule Engine do
 
   defdelegate workspace_symbols(query), to: CodeIntelligence.Symbols, as: :for_workspace
 
-  defdelegate runtime_versions, to: Forge.VM.Versions, as: :current
-
   defdelegate prepare_rename(analysis, position), to: Engine.CodeMod.Rename, as: :prepare
 
   defdelegate rename(analysis, position, new_name, client_name), to: Engine.CodeMod.Rename
@@ -78,6 +76,8 @@ defmodule Engine do
   defdelegate maybe_update_rename_progress(triggered_message),
     to: Engine.Commands.Rename,
     as: :update_progress
+
+  defdelegate runtime_versions, to: Forge.VM.Versions, as: :current
 
   def list_apps do
     for {app, _, _} <- :application.loaded_applications(),
--- a/apps/engine/lib/engine.ex
+++ b/apps/engine/lib/engine.ex
@@ -71,7 +71,8 @@ defmodule Engine do
 
   defdelegate prepare_rename(analysis, position), to: Engine.CodeMod.Rename, as: :prepare
 
-  defdelegate rename(analysis, position, new_name, client_name), to: Engine.CodeMod.Rename
+  defdelegate rename(analysis, position, new_name, client_name, rename_files?),
+    to: Engine.CodeMod.Rename
 
   defdelegate maybe_update_rename_progress(triggered_message),
     to: Engine.Commands.Rename,
--- a/apps/engine/lib/engine.ex
+++ b/apps/engine/lib/engine.ex
@@ -75,6 +75,10 @@ defmodule Engine do
 
   defdelegate rename(analysis, position, new_name, client_name), to: Engine.CodeMod.Rename
 
+  defdelegate maybe_update_rename_progress(triggered_message),
+    to: Engine.Commands.Rename,
+    as: :update_progress
+
   def list_apps do
     for {app, _, _} <- :application.loaded_applications(),
         not Forge.Namespace.Module.prefixed?(app),
--- a/apps/engine/lib/engine/application.ex
+++ b/apps/engine/lib/engine/application.ex
@@ -14,6 +14,7 @@ defmodule Engine.Application do
           Engine.CodeMod.Format.Cache,
           Engine.Api.Proxy,
           Engine.Commands.Reindex,
+          Engine.Commands.RenameSupervisor,
           Engine.Module.Loader,
           Engine.Compilation.TraceBuffer,
           Engine.Dispatch,
--- a/apps/engine/lib/engine/code_mod/aliases.ex
+++ b/apps/engine/lib/engine/code_mod/aliases.ex
@@ -1,11 +1,16 @@
 defmodule Engine.CodeMod.Aliases do
+  import Forge.Document.Line
+
   alias Engine.CodeMod.Directives
+  alias Forge.Ast
   alias Forge.Ast.Analysis
   alias Forge.Ast.Analysis.Alias
   alias Forge.Ast.Analysis.Scope
+  alias Forge.Document
   alias Forge.Document.Edit
   alias Forge.Document.Position
   alias Forge.Document.Range
+  alias Forge.Formats
 
   @doc """
   Returns the aliases that are in scope at the given range.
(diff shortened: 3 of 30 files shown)


doorgan (commented):
Some notes(I'm testing on a brand new `mix phx.new rename_test` project:

- If I am on module `RenameTestWeb.Router` and I rename it to `ExpertRenameTestWeb` then
  - No other module prefixed with `RenameTestWeb` is renamed
  - The `use RenameTestWeb, :router` isn't renamed either

- If I create `lib/rename_test/nested/foo.ex` defining `RenameTest.Nested.Foo` and I rename `RenameTest` in `lib/rename_test.ex`, then the child modules are renamed and the files are renamed too, but I end up with an empty `lib/rename_test/nested` folder

- When I try to rename a function(I know this is not supported by this PR, but still) I get this error:

```elixir
2026-01-14 16:34:27.690 [info] [Error - 4:34:27 PM] ** (UndefinedFunctionError) function XPGenLSP.ErrorResponse.fetch/2 is undefined (XPGenLSP.ErrorResponse does not implement the Access behaviour

You can use the "struct.field" syntax to access struct fields. You can also use Access.key!/1 to access struct fields dynamically inside get_in/put_in/update_in)
    (xp_gen_lsp 0.11.2) XPGenLSP.ErrorResponse.fetch(%XPGenLSP.ErrorResponse{message: "Failed to handle textDocument/prepareRename, {:error, :request_failed, \"Renaming :call is not supported for now\"}", code: -32603, data: nil}, :placeholder)
    (elixir 1.17.2) lib/access.ex:322: Access.get/3
    (xp_schematic 0.2.1) lib/schematic.ex:608: anonymous fn/5 in XPSchematic.map/1
    (elixir 1.17.2) lib/enum.ex:2531: Enum."-reduce/3-lists^foldl/2-0-"/3
    (xp_schematic 0.2.1) lib/schematic.ex:583: anonymous fn/3 in XPSchematic.map/1
    (xp_schematic 0.2.1) lib/schematic.ex:1001: anonymous fn/4 in XPSchematic.telemetry_wrap/3
    (xp_telemetry 1.3.0) /Users/dorgan/dev/expert/apps/expert/deps/telemetry/src/telemetry.erl:324: :xp_telemetry.span/3
    (xp_schematic 0.2.1) lib/schematic.ex:947: anonymous fn/3 in XPSchematic.oneof/1
```

Also we should have some progress reported, I'm testing on VSCode(it's easier for me to trigger these actions from this editor) and I don't see progress reported there, neither in the status bar nor in the Expert logs there

scottming:
Hi, @doorgan Thanks for the detailed feedback! Let me address each point:

**1. Renaming `RenameTestWeb.Router` to `ExpertRenameTestWeb`**

This is actually the expected behavior by design. The current rename implementation renames:
- The exact module and all its references
- Descendant modules (e.g., `MyApp.Users.Query` when renaming `MyApp.Users`)

When you rename `RenameTestWeb.Router`, other modules like `RenameTestWeb.PageController` are not descendants of `Router`, so they won't be renamed. Similarly, `use RenameTestWeb, :router` references the `RenameTestWeb` module, not `RenameTestWeb.Router`.

If you want to rename the `RenameTestWeb` prefix across all modules, you should trigger the rename on the `defmodule RenameTestWeb` definition itself.

**2. Empty folder left after renaming**

Good catch! This is a known limitation. The file rename operation is actually executed by the editor (VSCode/Neovim) via LSP's RenameFile - the editor handles creating parent directories automatically, but it doesn't clean up empty source directories afterward.

To fix this, we'd need to detect which directories become empty after the rename and send additional DeleteFile operations for them. Given that the rename feature is already quite complex (handling module references, descendants, file path conventions, Phoenix special folders, progress tracking, etc.), I'll note this as a potential future improvement, but it's lower priority for now.

**3. Error when trying to rename a function**

You're right - even though function renaming isn't supported yet, it should return a friendly error message like "Renaming :function is not supported for now" rather than throwing an `UndefinedFunctionError`. I'll look into this over the weekend since this is a higher priority issue.

**4. Progress reporting in VSCode**

Thanks for flagging this. We do have progress reporting implemented with VSCode-specific handling, but it seems something might be off. I'll investigate this as well.

scottming:
Hi, @doorgan  — the third and fourth issues you mentioned should both be fixed now. Please take a look at the two latest commits; they include detailed explanations of the problems and the solutions.

doorgan:
@scottming thanks for keeping this up to date and for the further fixes! I will take a better look after we're done with the full 0.1 release, this is a complex feature and I want to ensure we're doing the right thing in the right way. That you've been able to use it for so long is a great sign already though!

katafrakt:
I tested this today. Really nice work and there is definitely great potential here.
Few rough edges I encountered:

## Renaming usage with `use`

When I rename a macro, sometimes it did not rename `use The.Macro` calls. Other times it did. I haven't found yet under what conditions it happens.

## Progress notification sometimes hangs

Few times the "Renaming" progress notification hanged on me. I found this in the logs then:

```
12:23:04.858 instance_id=19CE1BC8F9F [debug] Node port message: 
12:23:04.858 [info] Starting rename tracking: 0 operations, 0 to reindex, 0 to delete
```

## Long time before progress notification shows for a large rename

Not sure why, but when I did a pretty big rename, it was a long time before the "Renaming" notification showed up the first time.

## Incorrect rename for nested modules

```elixir
defmodule MyApp.Test do
  defmodule Inner do
  end
end
```

In this case `Inner` gets renamed into full name:

```elixir
defmodule MyApp.Test do
  defmodule MyApp.Test.InnerRenamed do
  end
end
```

## Sometimes it triggers Refactorex errors

I believe this is not an issue with this PR per se, but the overall experience is that sometimes I see Refactorex errors after the rename. Often any subsequent renames are not possible.

Example errors:

```
    ** (FunctionClauseError) no function clause matching in XPSourceror.Zipper.replace/2
        (xp_sourceror 1.10.1) lib/sourceror/zipper.ex:340: XPSourceror.Zipper.replace(nil, {:extracted_variable, [], nil})
        (xp_refactorex 0.1.52) lib/refactorex/refactor/variable/extract_variable.ex:146: XPRefactorex.Refactor.Variable.ExtractVariable.replace_selection_by_variable/3
        (xp_refactorex 0.1.52) lib/refactorex/refactor/variable/extract_variable.ex:104: XPRefactorex.Refactor.Variable.ExtractVariable.extract_and_assign/4
        (xp_refactorex 0.1.52) lib/refactorex/refactor/variable/extract_variable.ex:41: XPRefactorex.Refactor.Variable.ExtractVariable.refactor/2
        (xp_refactorex 0.1.52) lib/refactorex/refactor/variable/extract_variable.ex:2: XPRefactorex.Refactor.Variable.ExtractVariable.visit/4
        (xp_sourceror 1.10.1) lib/sourceror/zipper.ex:612: XPSourceror.Zipper.do_traverse_while/3
        (xp_sourceror 1.10.1) lib/sourceror/zipper.ex:607: XPSourceror.Zipper.traverse_while/3
        (xp_refactorex 0.1.52) lib/refactorex/refactor/variable/extract_variable.ex:2: XPRefactorex.Refactor.Variable.ExtractVariable.execute/2
```

```
    ** (EXIT) an exception was raised:
        ** (FunctionClauseError) no function clause matching in anonymous fn/1 in XPRefactorex.Refactor.Function.next_available_function_name/2
            (xp_refactorex 0.1.52) lib/refactorex/refactor/function.ex:101: anonymous fn({:def, [trailing_comments: [], leading_comments: [], end_of_expression: [newlines: 2, line: 128, column: 41], line: 128, column: 3], [{:transact, [trailing_comments: [], leading_comments: [], closing: [line: 128, column: 40], line: 128, column: 7], [{:fun_or_multi, [trailing_comments: [], leading_comments: [], line: 128, column: 16], nil}, {:\\, [trailing_comments: [], line: 128, column: 35], [{:opts, [trailing_comments: [], leading_comments: [], line: 128, column: 30], nil}, {:__block__, [trailing_comments: [], leading_comments: [], closing: [line: 128, column: 39], line: 128, column: 38], [[]]}]}]}]}) in XPRefactorex.Refactor.Function.next_available_function_name/2
            (xp_refactorex 0.1.52) lib/refactorex/refactor/module.ex:32: anonymous fn/4 in XPRefactorex.Refactor.Module.next_available_name/4
            (elixir 1.19.4) lib/enum.ex:2520: Enum."-reduce/3-lists^foldl/2-0-"/3
            (xp_refactorex 0.1.52) lib/refactorex/refactor/module.ex:30: XPRefactorex.Refactor.Module.next_available_name/4
            (xp_refactorex 0.1.52) lib/refactorex/name_cache.ex:16: anonymous fn/2 in XPRefactorex.NameCache.consume_name_or/1
            (elixir 1.19.4) lib/agent/server.ex:20: Agent.Server.handle_call/3
            (stdlib 7.2) gen_server.erl:2470: :gen_server.try_handle_call/4
            (stdlib 7.2) gen_server.erl:2499: :gen_server.handle_msg/3
```

## Renaming files

I honestly don't like that it renames the files too. This is not what I asked for. But I think it could be ok if it only renamed the files when current filename matches current module name by convention. Unfortunately, it also renames when the convention is not met, i.e. when I rename `MyApp.Orders` in `my_app/index.ex` file to `MyApp.Ordering`, it lands in `my_app/ordering.ex`.

## Repo renaming?

This is strange, but it never ever allowed me to rename `MyApp.Repo` (as in: the repo for Ecto). Tried few times but to no avail.

doorgan:
Sorry it took me so long to get back to this one. I fixed all the issues we pointed out, plus a few others that I found while doing more testing(broken multi-alias renames, not respecting the client's advertised capabilities, wrong edits on changed but not saved files, etc). I think this PR is in good standing now. Generally speaking both the design and the code itself look reasonable and make sense to me.

I'd appreciate an extra pair of eyes though, as this is a large change and it's likely I missed something.

katafrakt:
I'll test it early next week.

But a question about strategy @doorgan: do we want this for 0.2? Because I also started "regular" rename PR and these two step on each other toes + I found way more edge cases than I though, so I paused it. Maybe we can postpone it post-0.2 and ship more or less complete renaming support in 0.3?

doorgan:
@katafrakt no, we'll release them for 0.3, we'll need more testing on both before we can merge them
I also need to take a look at your PR again to see how we can get to a common base for both implementations

katafrakt:
In that case, I think let's merge this one (when it's ready) after 0.2 release and I will rebase my renaming change onto it. TBH I don't remember now all the edge cases with my PR, but I think it's not yet ready for review.

PR #288, opened 2026-01-06
feat: use Spitfire to provide document symbols on documents with syntax errors

doorgan:
Switches the parser to Spitfire

Spitfire produces AST even when it finds syntax errors. Rather than just swapping the parser, this change integrates that feature by providing document symbols on invalid documents to exercise this new functionality

### DIFF
--- a/apps/forge/config/config.exs
+++ b/apps/forge/config/config.exs
@@ -1,5 +1,13 @@
 import Config
 
+parser =
+  case System.get_env("EXPERT_PARSER") do
+    "elixir" -> :elixir
+    _ -> :spitfire
+  end
+
+config :forge, :parser, parser
+
 config :snowflake,
   machine_id: 1,
   # First second of 2024
--- a/apps/forge/lib/forge/ast.ex
+++ b/apps/forge/lib/forge/ast.ex
@@ -77,6 +77,11 @@ defmodule Forge.Ast do
   require Logger
   require Sourceror
 
+  @parser (case Application.compile_env(:forge, :parser, :spitfire) do
+             :spitfire -> Forge.Ast.Parser.Spitfire
+             :elixir -> Forge.Ast.Parser.Elixir
+           end)
+
   @typedoc "Return value from `Code.Fragment.cursor_context/2`"
   @type cursor_context :: any()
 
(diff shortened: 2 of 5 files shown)


apps/engine/lib/engine/search/indexer/quoted.ex
```diff
@@ -14,16 +14,20 @@ defmodule Engine.Search.Indexer.Quoted do
...
 
   def index(%Analysis{valid?: true} = analysis, extractors) do
+    {:ok, extract_entries(analysis, extractors)}
+  end
+
+  def index(%Analysis{valid?: false}, _extractors) do
+    {:ok, []}
+  end
+
+  def extract_entries(%Analysis{} = analysis, extractors) do
```
doorgan, line 24: We will probably want to do this anyways in the future, but for now I opted to NOT index documents with syntax errors, this change helps with that.
I'm not 100% happy with calling Indexer code when to code is not actually being Indexed, but I couldn't come up with a better place to move this to, and I also have the gut feeling we'll end up indexing code with errors anyways

doorgan:
Question: should we add a flag to enable/disable Spitfire like NextLS, or should we just go ahead and fully adopt it?
I'm under the impression that flag was added when it was still very young, but now projects like Igniter use it, that's as good a seal of battle tested for me

mhanberg:
> Question: should we add a flag to enable/disable Spitfire like NextLS, or should we just go ahead and fully adopt it? I'm under the impression that flag was added when it was still very young, but now projects like Igniter use it, that's as good a seal of battle tested for me

Having a configuration option is probably still good, because if someone continually runs into fuel errors or exceptions with it, they could disable it while we work on a fix. The battle tested-ness of it being used by igniter is probably more around correct code than it is incomplete code. 

But, enable it by default.

But we can always ship it with no option then put an option in there later.

apps/forge/lib/test/detection_case.ex
```diff
@@ -132,7 +132,7 @@ defmodule Forge.Test.DetectionCase do
 
     quote generated: true do
       test unquote(test_name) do
-        assert_detected unquote(context), unquote(assertion_text)
+        assert_detected(unquote(context), unquote(assertion_text))
```
mhanberg: remove these changes if they weren't on purpose.

doorgan:
@mhanberg got it, I have to leave for my judo classes now, I'll add the option and address your comments when I'm back

mhanberg:
> judo classes

Note to self, never attempt to fight Dorgan

PR #290, opened 2026-01-07
feat(expert): support projectDir in `initializationOptions`

katafrakt:
I am struggling to make Expert work with a "monorepo" setup, where actual Elixir code is in a subdirectory (`api` in my case), while the parent directory is open. This works in VSCode, but not in other editors, in my experience.

By checking the VSCode extensions in seems that both Lexical and Expert extensions achieve that by [directly setting the path](https://github.com/expert-lsp/vscode-expert/blob/e59ade26c7e99110b56927da4c41319ae0e278cb/src/configuration.ts#L29-L34) based on the first workspace path and the extension's `projectDir` setting. This is then used to [build `workspaceFolders`](https://github.com/expert-lsp/vscode-expert/blob/e59ade26c7e99110b56927da4c41319ae0e278cb/src/extension.ts#L81-L85) and as I understand VSCode sets `rootUri` in `initialize` based on that, which is then used by Expert.

However, this method does not work well with other clients. For example, in Emacs/eglot attempt to overwrite `rootUri` is ignored (eglot seems to always set it on its own). Other editors don't even expose a theoretical possibility to do so. And even if, this should be an absolute path, which is not always easy to set.

As a resolution, I propose supporting `projectDir` in `initializationOptions`, which is a standard way for configuring LSPs during `initialize` request.

I tested this approach successfully in Emacs and NeoVim with following configurations:

```elisp
((elixir-ts-mode
 . ((eglot-server-programs
     . ((elixir-ts-mode . ("start_expert.sh" :initializationOptions
                           (:projectDir "api"))))))))
```

```lua
vim.lsp.config('expert', {
  cmd = { 'expert_darwin_arm64', '--stdio' },
  root_markers = { 'mix.exs', '.git' },
  filetypes = { 'elixir', 'eelixir', 'heex' },
  settings = {
    projectDir = 'api'
  }
})
```

In the future, I think this could also be supported in `workspace/didChangeConfiguration`, if possible, which I think is required for Zed (although I might have missed something and doing `initializationOptions` is possible there too.

### DIFF
--- a/apps/expert/lib/expert/configuration.ex
+++ b/apps/expert/lib/expert/configuration.ex
@@ -29,14 +29,35 @@ defmodule Expert.Configuration do
 
   @dialyzer {:nowarn_function, set_dialyzer_enabled: 2}
 
-  @spec new(Forge.uri(), map(), String.t() | nil) :: t
-  def new(root_uri, %Structures.ClientCapabilities{} = client_capabilities, client_name) do
+  @spec new(Forge.uri(), map(), String.t() | nil, map()) :: t
+  def new(
+        root_uri,
+        %Structures.ClientCapabilities{} = client_capabilities,
+        client_name,
+        init_options \\ %{}
+      ) do
     support = Support.new(client_capabilities)
-    project = Project.new(root_uri)
+    effective_root_uri = resolve_project_dir(root_uri, init_options)
+    project = Project.new(effective_root_uri)
 
     %__MODULE__{support: support, project: project, client_name: client_name}
     |> tap(&set/1)
   end
+
+  defp resolve_project_dir(root_uri, %{"projectDir" => project_dir})
+       when is_binary(project_dir) and project_dir != "" do
+    project_path =
+      if Path.type(project_dir) == :absolute do
+        project_dir
+      else
+        root_path = Forge.Document.Path.absolute_from_uri(root_uri)
+        Path.join(root_path, project_dir) |> Path.expand()
+      end
+
+    Forge.Document.Path.to_uri(project_path)
+  end
+
+  defp resolve_project_dir(root_uri, _init_options), do: root_uri
 
   @spec new(keyword()) :: t
   def new(attrs \\ []) do
--- a/apps/expert/lib/expert/state.ex
+++ b/apps/expert/lib/expert/state.ex
@@ -47,7 +47,10 @@ defmodule Expert.State do
         _ -> nil
       end
 
-    config = Configuration.new(event.root_uri, event.capabilities, client_name)
+    init_options = event.initialization_options || %{}
+
+    config =
+      Configuration.new(event.root_uri, event.capabilities, client_name, init_options)
 
     response = initialize_result()
     new_state = %__MODULE__{state | configuration: config, initialized?: true}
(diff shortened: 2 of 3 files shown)


mhanberg:
I don't want to reintroduce this pattern from ElixirLS. It doesn't work when you have multiple Elixir apps in subdirectories. 

This is what workspace folders are for, there is an open PR for it. Our efforts are best put towards getting that over the line.

mkaput:
Workspace folders also might not be useful in that monorepo that @katafrakt works on (I do also have access there) because top-level `mix.exs` is not in workspace root; while other root files (like AGENTS.md etc) are.

Heck, there are even 2 or 3 top-level `mix.exs` files there 🙃 

I don't know anything yet about stuff's structure in Expert; but in [CairoLS](https://github.com/software-mansion/cairols) what worked veeery well was to maintain a separate project management logic, which was bookkeeping an map of _workspace root path_ → _crates_/here mix projects analyzed by the LS. This map was automatically updated whenever some source file was opened and we walked up the directory tree to detect the top-level workspace manifest file (here `mix.exs`, there `Scarb.toml`). We didn't even bother with workspace folders with this design.

Perhaps this is worth discussing on some chat form somewhere?

doorgan:
@mkaput #18 already implements that partially, it'd be trivial to support that case in that PR. I'll do that.

mhanberg:
> Workspace folders also might not be useful in that monorepo that katafrakt works on (I do also have access there) because top-level mix.exs is not in workspace root; while other root files (like AGENTS.md etc) are.

I don't really understand what you are saying here. With workspace folders, the mix.exs for a folder doesn't need to be in the root, it needs to be in the folder. Other top level files don't really matter in this case.

> Heck, there are even 2 or 3 top-level mix.exs files there 🙃

Are you saying this is in the actual project root or in the subdirectory? Regardless, I don't really understand how that would work.

>  top-level workspace manifest file

Elixir does not have this concept.

katafrakt:
I tested lates version from `main` and it does indeed address this problem. Sorry for the unnecessary noise.

mkaput:
sorry if I'm saying something dumb about Expert here, haven't get time to dig into internals yet and just saying stuff based on my assumptions on how stuff may be done. my goal right now is to share experiences

> Are you saying this is in the actual project root or in the subdirectory? Regardless, I don't really understand how that would work.

we are talking about following repository structures

```shell
repo/
+ README.md
+ AGENTS.md # and other stuff like this, in general nobody will want to open a subdirectory in the editor

+ elixir_app/
  + mix.exs
+ subdirectory/
  + elixir_app_2/
    + mix.exs
  + ...
+ other_app_1/
```

> I don't really understand what you are saying here. With workspace folders, the mix.exs for a folder doesn't need to be in the root, it needs to be in the folder. Other top level files don't really matter in this case.

I was talking about [workspace/workspaceFolders](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders). Firstly, the editor will only report `repo/` as a workspace folder. Secondly, in CairoLS we decided not to trust editors to actually report `workspace/workspaceFolders` at all. I don't remember exactly, but we spotted editors which only used `didOpen` et al. Obviously nobody wrote down what apps did this, eh


> > top-level workspace manifest file
> 
> Elixir does not have this concept.

Umbrella projects are such thing aren't they?

```shell
+ kv_umbrella
  + mix.exs  # Expert should only know about this file, isn't it?
  + apps
    + kv
      + mix.exs
    + kv_server
      + mix.exs
```

---

please correct me if I seem not to understand something 🙏🏻

PR #293, opened 2026-01-09
feat: configure Workspace Symbols to return all symbols on empty query

### LINKED ISSUE #140: Workspace Symbols
## Description

I believe workspace symbols work slightly differently between Expert and Next LS, so we should hash out any changes needed to be made.

In particular, in Next LS if you don't provide a query, it returns all symbols. This allows the editor to just fuzzy filter them (depending on the editor) and is preferred by some people (me).

### LINKED ISSUE #155: Configuration
## Description

Expert doesn't currently have any concept of user configuration.

I believe that we will won't some kind of configuration, as there are nuances to how the features work and some people might prefer to have them work in slightly different ways.

For example, Expert I believe will not return any workspace symbols unless you provide a query, but Next LS returns all of the symbols. I prefer the latter, as I like to have the editor to filter them. This could be configurable.

## Solution

- [ ] The server can be configured by the editor.

Note: This scope does not include a configuration file.

doorgan:
Closes #140
Closes #155

There's a rather big change to the way we do configuration, not only to support this new option, but also becase there was a race condition that prevented config changes from being applied reliably, as requests other than didChangeConfiguration could be handled at the same time as eg textDocument/didOpen, which would return a state with the old config, overwriting the new one.

The new approach is to use `:persistent_term` instead. I opted for this rather than a genserver or an agent because configuration changes are infrequent, and removing the config from the server state is enough to eliminate that race condition

### DIFF
--- a/apps/expert/lib/expert/configuration.ex
+++ b/apps/expert/lib/expert/configuration.ex
@@ -5,7 +5,6 @@ defmodule Expert.Configuration do
 
   alias Expert.Configuration.Support
   alias Expert.Configuration.WorkspaceSymbols
-  alias Expert.Dialyzer
   alias Expert.Protocol.Id
   alias GenLSP.Notifications.WorkspaceDidChangeConfiguration
   alias GenLSP.Requests
@@ -14,20 +13,16 @@ defmodule Expert.Configuration do
   defstruct support: nil,
             client_name: nil,
             additional_watched_extensions: nil,
-            dialyzer_enabled?: false,
             workspace_symbols: %WorkspaceSymbols{}
 
   @type t :: %__MODULE__{
           support: support | nil,
           client_name: String.t() | nil,
           additional_watched_extensions: [String.t()] | nil,
-          dialyzer_enabled?: boolean(),
           workspace_symbols: WorkspaceSymbols.t()
         }
 
   @opaque support :: Support.t()
-
-  @dialyzer {:nowarn_function, set_dialyzer_enabled: 2}
 
   @spec new(Structures.ClientCapabilities.t(), String.t() | nil) :: t
   def new(%Structures.ClientCapabilities{} = client_capabilities, client_name) do
@@ -83,22 +78,10 @@ defmodule Expert.Configuration do
   defp apply_config_change(%__MODULE__{} = old_config, %{} = settings) do
     new_config =
       old_config
-      |> set_dialyzer_enabled(settings)
       |> set_workspace_symbols(settings)
       |> set()
 
     maybe_watched_extensions_request(new_config, settings)
-  end
-
-  defp set_dialyzer_enabled(%__MODULE__{} = old_config, settings) do
-    enabled? =
-      if Dialyzer.check_support() == :ok do
-        Map.get(settings, "dialyzerEnabled", true)
-      else
-        false
-      end
-
-    %__MODULE__{old_config | dialyzer_enabled?: enabled?}
   end
 
   defp set_workspace_symbols(%__MODULE__{} = config, settings) do
--- a/apps/expert/lib/expert/dialyzer.ex
+++ b/apps/expert/lib/expert/dialyzer.ex
@@ -1,5 +1,0 @@
-defmodule Expert.Dialyzer do
-  def check_support do
-    :ok
-  end
-end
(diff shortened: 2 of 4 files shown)


apps/expert/lib/expert/configuration.ex
```diff
@@ -60,36 +64,30 @@ defmodule Expert.Configuration do
...
+    apply_config_change(get(), change.params.settings)
   end
 
   defp apply_config_change(%__MODULE__{} = old_config, %{} = settings) do
-    old_config
-    |> set_dialyzer_enabled(settings)
-    |> maybe_add_watched_extensions(settings)
+    new_config =
+      old_config
+      |> set_dialyzer_enabled(settings)
```
mhanberg: I'm confused, do we support dialzyer?

doorgan: The option was already there and I didn't remove it. Looking into the codebase we're not doing anything with it so I'll remove it altogether.
Looking at discussions from when this was still Lexical, there were some attempts to add dialyzer and there was a TODO for it: https://github.com/lexical-lsp/lexical/issues/95
But it never got implemented, and this config option seems to be a leftover from that time

mhanberg: Roger

pages/configuration.md
```diff
@@ -0,0 +1,19 @@
...
+    "minQueryLength": 2
+  }
+}
+```
+
+## Available Settings
+
+| Setting | Type | Default | Description |
+|---------|------|---------|-------------|
+| `workspaceSymbols.minQueryLength` | integer | `2` | Minimum characters required before workspace symbol search returns results. Set to `0` to return all symbols with an empty query. |
```
mhanberg, line 19: Do these need to be camelCased? I feel like some servers have them camel cased even if their language is not cameCased, possibly because the VSCode settings use that convention.

doorgan: They don't need to be, I was following convention as you mention. I have no opinion on which casing to use, we can use snake_case instead

apps/expert/lib/expert/configuration/workspace_symbols.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Expert.Configuration.WorkspaceSymbols do
+  @moduledoc false
+
+  defstruct min_query_length: 2
```
mkaput, line 4: out of curiosity - what's the reason behind having this default to `2` and allowing user to configure this altogether? performance? because to me it looks like `0` as hardcoded would be a reasonable thing to have and I don't recall other LSes having such config

pages/configuration.md
mhanberg: Yeah, I can't tell if I want it to be snake cased cuz I think it matters or if my brain just sees camelCase and is like "must destroy"

I'll lean toward doesn't matter who cares. 

But we'll want this documented on the website after we merge.

apps/expert/lib/expert/configuration/workspace_symbols.ex
doorgan: It's the previous behavior, I had to type 2 characters to get completions before this change. Hardcoding `0` makes us return every result and let the client filter them, which is the NextLS behavior. Having >0 causes the server to filter results first, which is the Lexical behavior.

pages/configuration.md
doorgan: Yep, I'll update expert-lsp.org after this

apps/expert/lib/expert/configuration/workspace_symbols.ex
mkaput: Yeah I get this. I was wondering why `2` was in the first place here 😅

doorgan: Not sure! Maybe because a single letter wasn't considered specific enough for filtering to be meaningful?

mkaput: I think this might be editor-specific. VSCode and IntelliJ I know have even some ML models for sorting such things. otoh probably helix/neovim (not using them) just throw stuff as they get from the LSP. So I guess, perhaps the original author picked what worked in his editor? 🤔

PR #295, opened 2026-01-11
fix: plain Release failing to start Engine due to RELEASE_* environment pollution

scottming:
### Context

`just burrito-local` and `just release` produce different release types:

| Feature | `just burrito-local` | `just release` (Plain) |
|---------|---------------------|------------------------|
| Output | Single executable binary | Standard Elixir release directory |
| Includes ERTS | Yes (downloaded remotely) | Yes (local ERTS bundled) |
| Startup | Self-contained, isolated environment | Shell scripts that set `RELEASE_*` env vars |
| `RELEASE_*` vars at runtime | 2 | 15 |

When running Expert via Plain Release, the startup script (`bin/plain`) sets 15 `RELEASE_*` environment variables and prepends the release's `erts-*/bin` and `bin` directories to `PATH`.

This caused Engine startup to fail with:
```
Runtime terminating during boot ({'cannot get bootfile','/path/to/rel/plain/bin/start.boot'})
```

**Root cause**: When Expert spawned an external elixir process to build/start the Engine:
1. The polluted `PATH` caused `erl` to resolve to the release's bundled `erl` instead of the system's (via asdf)
2. The release's `erl` sets `ROOTDIR` to the release directory
3. Erlang then tried to load `{ROOTDIR}/bin/start.boot`, which doesn't exist

---

### Changes

- Extract `resolve_path/1` and `sanitized_env/1` helpers for cleaner code organization
- Filter `RELEASE_*` environment variables from child processes using `{key, false}` (Erlang's mechanism to unset env vars)
- Remove release directory paths from `PATH` before passing to child processes
- Preserve `false` atom in `ensure_charlists/1` instead of converting to string
- Add Windows path separator support in `remove_release_paths/2`
- Improve error messages with project context

---

### Why These Changes

Child processes spawned via `Port.open` inherit the parent's environment. Simply omitting `RELEASE_*` vars from the `:env` option doesn't remove them—they're still inherited. We must explicitly pass `{Name, false}` to unset them.

Similarly, the release's ERTS binaries in `PATH` would shadow the system's Erlang/Elixir. Cleaning `PATH` ensures the child process uses the correct runtime for the target project (respecting version managers like asdf/mise).

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -38,6 +38,18 @@ defmodule Expert.Port do
   end
 
   def elixir_executable(%Project{} = project) do
+    {path, error_context} = resolve_path(project)
+
+    case :os.find_executable(~c"elixir", to_charlist(path)) do
+      false ->
+        {:error, :no_elixir, "Couldn't find an elixir executable#{error_context}"}
+
+      elixir ->
+        {:ok, elixir, sanitized_env(path)}
+    end
+  end
+
+  defp resolve_path(%Project{} = project) do
     if Forge.OS.windows?() do
       # Remove the burrito binaries from PATH
       path =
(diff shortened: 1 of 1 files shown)


jorams:
Welp I guess should have looked for open pull requests because I just debugged the exact same problem and was writing up an issue.

Notably the `PATH` issue won't happen on every system because the value for `$PATH` is determined in `Expert.Port.path_env_at_directory` by starting a shell in the project root directory, and on some systems (e.g. Debian) the value for `$PATH` will be set to a specific value in `/etc/profile`. On others (e.g. Arch, Fedora) the existing value will be kept and defaults will only be appended. The likes of ASDF and Mise can also prefix working paths, hiding the problem.

Also note that it's not `bin/plain` prepending the `erts-*/bin` and `bin` directories to `PATH`, that only happens inside the Erlang process.

The simplest possible fix to make it start successfully is to unset `$PATH` in the environment for the shell so you always end up with a clean end result. That could also remove paths the user does want, but on systems like Debian it will do that anyway:


```diff
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -89,8 +89,12 @@ defmodule Expert.Port do
     # or we get an incomplete PATH not including erl or any other version manager
     # managed programs.

-    # Disable shell session history to reduce noise
-    env = [{"SHELL_SESSIONS_DISABLE", "1"}]
+    env = [
+      # Disable shell session history to reduce noise
+      {"SHELL_SESSIONS_DISABLE", "1"},
+      # Start with a clean PATH to get the correct new value
+      {"PATH", nil}
+    ]

     args =
       case Path.basename(shell) do
```

scottming:
@jorams 

I tried the simpler approach you mentioned, and it seems to work as well.

I think it would be more appropriate for you to submit this PR, so I’ll close mine.

doorgan:
@jorams thanks, your suggestion makes a lot of sense, let's try with that first

jorams:
I'll open a PR once I get off work.

Issue #275, opened 2025-12-22
Engine is compiled with system Erlang, despite using Mise (MacOS, fish)


katafrakt:
OS: MacOS Tahoe
Editor: Emacs / eglot
Shell: fish
Expert version: built from git
Version manager: mise

This is probably very niche, but Expert stopped working for me today at all. Whenever it tries to start, after compiling the engine it fails with something like:

```
3:06:42.859 [debug] Node port message: 
13:06:42.851 [error] beam/beam_load.c(150): Error loading module 'Elixir.XPForge.NodePortMapper':
  corrupt atom table



13:06:42.865 [debug] Node port message: 
13:06:42.865 [error] beam/beam_load.c(150): Error loading module 'Elixir.XPForge.EPMD':
  corrupt atom table
```

Even worse, now when I try to do anything in my project, I'm getting similar error with `Error loading module 'Elixir.Hex`. Every time, I need to `mix local.hex --force`.

After some digging I found that when engine building does it's own `mix local.hex --force`, it does that in Elixir 1.19.4-otp27 directory, but using Erlang/OTP 28 installed from Homebrew. Even further down the line, I found that `Expert.Port` when detecting `fish`, in `path_Env_at_directory`, starts a new fish shell using `fish -l -c`, which modifies the path. This is probably because `-l` reevaluates fish config, runs `eval "$(/opt/homebrew/bin/brew shellenv)"` from ~/.config/fish/conf.d/env.fish` (I believe this is autogenerated by Homebrew, as I don't recall creating this file).

As a result, the beginning of path when I run `echo $PATH` in my shell:

```
/Users/pawel.sw/.opam/4.14.0/bin /Users/pawel.sw/Downloads/google-cloud-sdk/bin /Users/pawel.sw/.local/share/mise/installs/elixir/1.19.4-otp-27/bin
```

but when I log the path in `Expert.Port`, after `path_env_in_directory`, it is:

```
/opt/homebrew/bin:/opt/homebrew/sbin:/Users/pawel.sw/.local/bin:/Users/pawel.sw/dev/lexical/_build/dev/package/lexical/bin
```

The directories are in a completely different order, and indeed, `erl` now points to the Homebrew installation.

The issue also manifests in log file with this message:

```
12:43:20.143 [debug] Building engine: * creating /Users/pawel.sw/.local/share/mise/installs/elixir/1.19.4-otp-27/.mix/archives/hex-2.3.1-otp-28

[...]

PR #298, opened 2026-01-12
fix: inherited PATH pollutes project environment

jorams:
This is a followup to the discussion in #295.

To determine the correct value for PATH to start the engine build and project node with, `Expert.Port.path_env_at_directory` starts a new shell in the project directory. When doing so the server's own PATH variable is inherited by that shell.

In the context of an Expert release built using `just release`, this variable contains paths pointing to the release, including one containing an `erl` executable. The surrounding directory structure does not contain a `start.boot` file in the location `erl` expects.

On some systems (e.g. Debian) the new shell will set PATH to a completely new value or prefix it with paths containing a complete Erlang environment. On others (e.g. Arch, Fedora) the new shell will only append defaults to the value, leaving the unwanted paths in place.

In the latter case Expert will launch `elixir`, which in turn launches `erl`, which won't be able to find a `start.boot`-file and therefore crash.

By providing the new shell an unset PATH it is fully responsible for determining an appropriate value without being affected by the environment of Expert.

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -89,8 +89,12 @@ defmodule Expert.Port do
     # or we get an incomplete PATH not including erl or any other version manager
     # managed programs.
 
-    # Disable shell session history to reduce noise
-    env = [{"SHELL_SESSIONS_DISABLE", "1"}]
+    env = [
+      # Disable shell session history to reduce noise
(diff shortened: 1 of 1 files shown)


doorgan (approved):
Thank you! 💜

katafrakt:
Also fixes #275 for me

PR #300, opened 2026-01-13
fix: fallback to packaged or system elixir

### LINKED ISSUE #299: Expert goes looking for Elixir in a fresh PATH in a fresh shell, but my dev environment can't work that way
Hiya! This is my third time giving Expert a time; each time I've had a different show-stopper.

Today we get further than ever before, but bring-up fails at:

```
...
23:16:37.752 [info] Child :net_kernel of Supervisor :net_sup started
Pid: #PID<0.186.0>
Start Call: :net_kernel.start_link(%{name: :"expert-manager-kv-20199@127.0.0.1", supervisor: :net_sup_
dynamic, net_tickintensity: 4, net_ticktime: 60, name_domain: :longnames, clean_halt: false})
Restart: :permanent
Shutdown: 2000
Type: :worker
23:16:37.753 [debug] Child :net_sup_dynamic of Supervisor :kernel_sup started
Pid: #PID<0.183.0>
Start Call: :erl_distribution.start_link(%{name: :"expert-manager-kv-20199@127.0.0.1", supervisor: :ne
t_sup_dynamic, net_tickintensity: 4, net_ticktime: 60, name_domain: :longnames, clean_halt: false})
Restart: :permanent
Shutdown: 1000
Type: :supervisor
23:16:37.779 [debug] sent notification server -> client window/logMessage
23:16:37.780 [error] Failed to find an elixir executable, shutting down
23:16:37.780 [debug] sent notification server -> client $/progress
23:16:37.784 [error] Process #PID<0.182.0> terminating
** (exit) :ok
    (stdlib 6.2.2.2) gen_server.erl:2210: :gen_server.init_it/6
    (stdlib 6.2.2.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
```

Reading the source, it turns out Expert tries to launch a new shell without a `PATH` variable set, navigate to the project directory, and then look for `elixir` in `PATH`:

https://github.com/elixir-lang/expert/blob/8e1bb3dc5328ddc6fcab2203767b44a92db0909a/apps/expert/lib/expert/port.ex#L85-L97

I use Nix and dev shells (`nix develop`) to manage my toolchains. This will never find an `elixir` executable, since I don't (and won't ever) have one setup outside a dev shell. Indeed, the correct `elixir` executable is the one that was already on `PATH`, because I'm in a dev shell when I start my editor! This is how my other LSPs manage to find their tools.

Can we perhaps fall back to an `elixir` found on the `PATH` in Expert's environment itself if this method doesn't produce anything?

doorgan:
Fix #299

Fallback to the elixir packaged by burrito, or the system elixir if no elixir is found for the project.

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -38,6 +38,20 @@ defmodule Expert.Port do
   end
 
   def elixir_executable(%Project{} = project) do
+    case find_project_elixir(project) do
+      {:ok, _, _} = success ->
+        success
+
+      {:error, :no_elixir, reason} ->
+        Logger.warning(
+          "Failed to find elixir for project, falling back to packaged elixir: #{reason}"
+        )
+
+        fallback_elixir()
+    end
+  end
+
+  defp find_project_elixir(%Project{} = project) do
     if Forge.OS.windows?() do
       # Remove the burrito binaries from PATH
       path =
@@ -79,6 +93,17 @@ defmodule Expert.Port do
 
           {:ok, elixir, env}
       end
+    end
+  end
+
+  # Fallback to using whatever elixir Expert was packaged with.
+  defp fallback_elixir do
+    case System.find_executable("elixir") do
+      nil ->
+        {:error, :no_elixir, "Couldn't find any elixir executable"}
+
+      elixir ->
+        {:ok, to_charlist(elixir), []}
     end
   end
 


mhanberg:
I also use nix and devshells, let me test current main and then test this.

doorgan:
@mhanberg for context, I think this started happening after we merged #298 
That PR fixed `just release`, but it seems it broke it for nix/devshells users.

mhanberg:
Yes, current main does not work.

This branch seems to pick up the correct path from the devshell that nvim was started in.

PR #305, opened 2026-01-14
fix: use minimal PATH on unix instead of fully removing it

doorgan:
I started seeing errors about `mv` being missing on `main`, this fixes that by using a minimal PATH instead of an empty one.


doorgan:
I'm merging this, this issue is being very disruptive as it flat out prevents expert from working on macos

katafrakt:
It worked fine for me on MacOS before this on fish. Which shell are you using?

doorgan:
@katafrakt  I'm using `zsh`

Issue #303, opened 2026-01-14
Timeouts on querying search store


katafrakt:
On a fairly large codabase, I very often see timeouts like this in the logs:

```
09:15:13.016 [error] Task #PID<0.243.0> started from XPExpert terminating
** (stop) {:exception, {:timeout, {GenServer, :call, [XPEngine.Search.Store, {:exact, "JumpWeb.Admin.FeatureFlagDefinitionsLive.rollout_state_badge/2", [subtype: :definition]}, 5000]}}}
    (kernel 10.2.7.3) erpc.erl:1366: :erpc.call/5
    (xp_expert 0.1.0) lib/expert/provider/handlers/go_to_definition.ex:21: XPExpert.Provider.Handlers.GoToDefinition.handle/2
    (xp_expert 0.1.0) lib/expert.ex:121: XPExpert.handle_request/2
    (xp_gen_lsp 0.11.2) lib/gen_lsp.ex:369: anonymous fn/2 in XPGenLSP.loop/3
    (xp_telemetry 1.3.0) /Users/pawel.sw/dev/expert/apps/expert/deps/telemetry/src/telemetry.erl:324: :xp_telemetry.span/3
    (xp_gen_lsp 0.11.2) lib/gen_lsp.ex:368: anonymous fn/5 in XPGenLSP.loop/3
    (xp_gen_lsp 0.11.2) lib/gen_lsp.ex:592: anonymous fn/4 in XPGenLSP.attempt/4
    (elixir 1.19.4) lib/task/supervised.ex:105: Task.Supervised.invoke_mfa/2
Function: #Function<0.130722175/0 in XPGenLSP.attempt/4>
    Args: []
09:15:13.018 [error] Child :undefined of Supervisor :expert_task_queue terminated
** (exit) {:exception, {:timeout, {GenServer, :call, [XPEngine.Search.Store, {:exact, "JumpWeb.Admin.FeatureFlagDefinitionsLive.rollout_state_badge/2", [subtype: :definition]}, 5000]}}}
```

After seeing this, functions like go to definition and hover seem to be severely impaired in the project (although definition sometimes works, I assume by falling back to ElixirSense).

I'm seeing this for `:exact`, `:find_by_subject` and `:delete_by_path`.

[...]

PR #308, opened 2026-01-16
fix(engine): don't attempt search when ETS checkpoint is loading

katafrakt:
The checkpoint file can be quite large (1.2G in my case) and it may take quite a while to load it (8.2s in my case). During that time, all the GenServer calls with the default timeout of 5000ms might timeout, leading to the whole GenServer termination and general instability of the search functionality.

This adds guards for state.loaded? which return {:error, :loading} if it's not loaded.

This addresses #303, probably only partially. I have more suspects of taking too long on a large project:
* `query_by_subject` during delete (as described in #303)
* `reduce` takes 2-3 seconds for my project
* apply large WAL file - I once saw it grow to 17M, but I didn't manage to measure how long does it take to apply it

### DIFF
--- a/apps/engine/lib/engine/search/store.ex
+++ b/apps/engine/lib/engine/search/store.ex
@@ -3,6 +3,7 @@ defmodule Engine.Search.Store do
   A persistent store for search entries
   """
 
+  alias Engine.Dispatch
   alias Engine.Search.Store
   alias Engine.Search.Store.State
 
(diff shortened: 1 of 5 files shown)


mhanberg:
What is the result of adding these error return values?

Does the editor display an error message?

katafrakt:
@mhanberg it is handled by the caller, like for example [here](https://github.com/elixir-lang/expert/blob/03ec5c6cd614bc17c7ec3d68d1cf5f7c2ed185bb/apps/engine/lib/engine/code_intelligence/definition.ex#L176-L182):

```elixir
    case Store.exact(subject, condition) do
      {:ok, entries} ->
        entries

      _ ->
        []
    end
```

One thing that could maybe be improved is to handle it inside [call_or_default](https://github.com/elixir-lang/expert/blob/03ec5c6cd614bc17c7ec3d68d1cf5f7c2ed185bb/apps/engine/lib/engine/search/store.ex#L292), but I'm not sure. What do you think?

Either way, it's not visible to the editor. Client just gets an empty result set (and rightly so, I think, it's not an error - search index is simply not available at the moment).

mhanberg:
Gotcha, I think it could be worth to keep the behavior as you have it but also post a window/showMessage notification that says the index is still loading. 

Or else to the user they think it's just not working.

katafrakt:
Added. Works in my tests.
Screenshot 2026-01-16 at 18 49 59

mhanberg (approved):

mhanberg:
It might also be good to emit a progress notification when loading the search thing

PR #313, opened 2026-01-19
feat(engine): indicate progress when loading checkpoint

katafrakt:
This implements the idea from #308 to add progress notification while the ets checkpoint is loading.

Screenshot 2026-01-19 at 11 03 29


doorgan (approved):
Looks good, thanks!

Issue #291, opened 2026-01-08
Hover does not work for shorthand notation inside ~H sigil


katafrakt:
While testing #278, @doorgan discovered that go-to-definition works, but hover does not. I confirmed this today. Creating an issue, as agreed in the discussion.

(I plan to look into this soon)

mhanberg:
@katafrakt can this be closed?

katafrakt:
Yes, I think let's close it. It's not perfect, but much better than when I created the issue. If remaining things come up, I think it's better to take them case-by-case in new issues.

Issue #270, opened 2025-12-19
Macro related Go to definition support and other integrations


doorgan:
Opening this to track support for third party library integrations to Expert

Some libraries like Phoenix LiveView, Nx and Ash have special needs when it comes to go to definition inside their macros.

For example:
- LiveView uses the `~H` sigil, and go to definition should work there for function components or interpolated content
- Nx would benefit from smarter go to definition for calls related to `defn` and `deftransform`
- Ash could benefit from enriching the results we provide to the editor, similar to what the [Ash Studio](https://marketplace.visualstudio.com/items?itemName=ketupia.ash-studio) extension does

We need better direct integration with Expert to support libraries. The general sentiment seems to be that the current plugin architecture is lacking for this.

PR #314, opened 2026-01-19
fix(engine): resolve correct arity from inside ~H sigil

katafrakt:
This fixes a deficiency of a previous implementation of jump-to-definition from HEEx templates (#278). While it worked for some cases, it always looked up a function of arity 2. This happened because `arity_at_point` function looked up arity of `sigil_H` function, which is two.

This change introduces partial parsing of the contents of the ~H sigil and adds more precise arity resolving for these cases. This also enables some hover inside ~H (#291).

### Additional notes
* I renamed `HeexNormalizer` to `Heex`, because it's not only normalizing anymore. Once the proper support for plugins is added (#270), this module would contain LiveView-specific addons in one place.
* Curly braces handling is not ideal, but should cover most of the basic cases. TBH I think we can only get so far without using `Phoenix.LiveView.TagEngine`, but that would create a runtime dependency, which I think is not good without proper extensions support.


katafrakt:
There's a Credo failure. here caused by a bug in Credo that has been recently fixed: https://github.com/rrrene/credo/issues/1164

doorgan:
We can update credo to latest

katafrakt:
Updated Credo, now it passes 👍

mhanberg:
@katafrakt rebase your changes and then we can merge

katafrakt:
@mhanberg rebased

doorgan (approved):

PR #321, opened 2026-01-20
fix: interpolation_ranges/1 should work for empty interpolations

doorgan:
With Spitfire we now parse things like `"foo#{}bar"`. The previous code wasn't able to calculate ranges for empty interpolations and crashed.


mhanberg:
I'm curious as to why this wasn't possible before

doorgan:
Same, I found how to reproduce it but I'm not 100% sure why it happens just now, the ASTs from both parsers are identical

PR #326, opened 2026-01-21
fix: update document store on didchange even without the engine running

doorgan:
We were dropping this event if the engine for that project wasn't running because it would attempt to trigger a document compilation.

Skipping the document compilation is correct, but we should still be updating the document store with the new document contents.

### DIFF
--- a/apps/expert/test/expert/expert_test.exs
+++ b/apps/expert/test/expert/expert_test.exs
@@ -561,4 +561,63 @@ defmodule ExpertTest do
       assert_project_alive?(nested_subproject)
     end
   end
+
+  describe "text document changes" do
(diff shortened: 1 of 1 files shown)


apps/expert/lib/expert/state.ex
```diff
@@ -128,30 +128,23 @@ defmodule Expert.State do
     projects = ActiveProjects.projects()
     project = Project.project_for_uri(projects, uri)
 
-    with true <- ActiveProjects.active?(project),
-         {:ok, updated_source} <-
-           Document.Store.get_and_update(
-             uri,
-             # TODO: this function needs to accept the GenLSP data structure
```
doorgan, line 135: This comment was stale, we already support that structure

PR #328, opened 2026-01-22
fix(engine): improve entity resolution for HEEx components with curly braces

katafrakt:
Previous implementation had a bug where if there was a curly braces expression on the line with a HEEx component, its arity was always taken, even if the cursor was on the component. This lead to incorrect resolution (usually to not be able to resolve the component call).

This is improvement on top of #314


doorgan (approved):
Works great, thanks!

PR #329, opened 2026-01-22
feat(engine): hover for module attributes

katafrakt:
Something I've been missing is the ability to quickly look up how the module attribute is defined. It just shows the definition, does not try to evaluate it, which is sufficient in my opinion for most cases.

Go to definition already worked for module attributes.

Screenshot 2026-01-22 at 17 31 43

### DIFF
--- a/apps/expert/test/expert/provider/handlers/hover_test.exs
+++ b/apps/expert/test/expert/provider/handlers/hover_test.exs
@@ -1,4 +1,5 @@
 defmodule Expert.Provider.Handlers.HoverTest do
+  alias Engine.Search
   alias Expert.EngineApi
   alias Expert.Protocol.Convert
   alias Expert.Provider.Handlers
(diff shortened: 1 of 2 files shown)


apps/expert/test/expert/provider/handlers/hover_test.exs
```diff
@@ -727,6 +728,115 @@ defmodule Expert.Provider.Handlers.HoverTest do
...
+    Enum.reduce_while(1..50, :not_ready, fn _, _ ->
+      case EngineApi.call(project, Search.Store, :loaded?, []) do
+        true ->
+          {:halt, :ok}
+
+        _ ->
+          Process.sleep(100)
+          {:cont, :not_ready}
+      end
+    end)
```
doorgan, line 837: Maybe we should produce a message like the one we have for `project_index_ready` and wait for it rather than polling the store

doorgan: We can do that in a follow-up PR, as it's not that related to the changes here anyways

katafrakt: Actually, I believe we can use `project_index_ready` here, because it's emitted after `loaded?` is set to true.

PR #332, opened 2026-01-23
chore(expert): wait for message instead of polling in tests

katafrakt:
addresses https://github.com/elixir-lang/expert/pull/329#discussion_r2718122139


mhanberg:
For future reference, this would be a `test`. Not a `chore`.

PR #333, opened 2026-01-23
test: correctly expect lexicographic sorting

leonqadirie:
**I'm a bit confused as I can't find any reference to this failing for others and CI seems to pass.**
But on my machine, `just test expert --warnings-as-errors` reliably fails 2 test assertions on `main` and is _very_ timeout-flaky for some others. This fixes the 2 reliably failing assertions not matching the expectations of lexicographic ordering of `apps/expert/lib/expert/engine.ex:103-114`.

My machine:
- macOS 26.2
- Erlang/OTP 28.3.1
- Elixir 1.19.5
- nushell 0.110.0

---
Reproducible test failures:
```elixir
  1) test run/1 - clean subcommand with --force stops deleting after first error and returns error code 1 (Expert.EngineTest)
     test/expert/engine_test.exs:81
     Assertion with =~ failed
     code:  assert Enum.at(attempted_dirs, 1) =~ "0.2.0/foobar"
     left:  "/path/to/repo/expert/apps/expert/tmp/Expert.EngineTest/test-run-1---clean-subcommand-with---force-stops-deleting-after-first-error-and-returns-error-code-1-052d261f/0.2.0/bazbeau"
     right: "0.2.0/foobar"
     stacktrace:
       test/expert/engine_test.exs:122: (test)

  2) test run/1 - clean subcommand interactive mode handles multiple directories with mixed responses (Expert.EngineTest)
     test/expert/engine_test.exs:169
     Expected truthy, got false
     code: assert File.exists?(dir2)
     arguments:

         # 1
         "/path/to/repo/expert/apps/expert/tmp/Expert.EngineTest/test-run-1---clean-subcommand-interactive-mode-handles-multiple-directories-with-mixed-responses-adb406df/0.2.0/foobar"

     stacktrace:
       test/expert/engine_test.exs:184: (test)
```

### DIFF
--- a/apps/expert/test/expert/engine_test.exs
+++ b/apps/expert/test/expert/engine_test.exs
@@ -111,7 +111,7 @@ defmodule Expert.EngineTest do
       assert output =~ "Error deleting"
       assert output =~ dir3
 
-      # Should only attempt dir1 and dir2, not dir3
+      # Sorted order is dir1, dir3, dir2; should only attempt dir1 and dir3
       attempted_dirs =
         agent_pid
         |> Agent.get(& &1)
@@ -119,7 +119,7 @@ defmodule Expert.EngineTest do
 
       assert length(attempted_dirs) == 2
       assert Enum.at(attempted_dirs, 0) =~ "0.1.0/foobar"
-      assert Enum.at(attempted_dirs, 1) =~ "0.2.0/foobar"
+      assert Enum.at(attempted_dirs, 1) =~ "0.2.0/bazbeau"
     end
   end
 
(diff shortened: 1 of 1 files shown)


mhanberg:
This would be a `test` and not a `fix`

leonqadirie:
> This would be a `test` and not a `fix`

So a `fix` must be user-facing akin to a patch in semantic versioning, fixing a logic error doesn't suffice?

Adjusted.

mhanberg:
Fixes are patch, features are minor, breaking changes use a bang with either. 

Other types don't show up in the changelog.

leonqadirie:
More meant as differentiation criteria for me in the future - so if I fix sth. that is not noticeable to the consumer (such as the test case here), I'd not put it into `fix` but the next-best category. Got it.

I'm still a bit confused why CI was green before and now is still green with explicitly changed expectations. There are test jobs that run.

PR #341, opened 2026-01-28
docs: add enigmatic sponsor logo

danj3:
enigmatic sponsorship via eef earmark fund.


danj3:
needs to be added to expert website

PR #343, opened 2026-01-28
fix(engine): handle matches against any struct

katafrakt:
Using an "any struct match" (I don't know the proper term, sorry) causes errors being logged during indexing:

```
14:58:09.984 [debug] Node port message: 
14:58:09.984 [error] Could not expand alias: {:_, [line: 1590, column: 35], nil} at /path/to/file.ex 1590:34
```

The code causing it is, for example:

```elixir
  defp normalize_suggestion_data(%_{} = struct) do
    Map.from_struct(struct)
  end
```

This change adds proper handling to this kind of matches (ignoring them, as they are not useful, but without emitting an error).

### DIFF
--- a/apps/engine/test/engine/search/indexer/extractors/struct_reference_test.exs
+++ b/apps/engine/test/engine/search/indexer/extractors/struct_reference_test.exs
@@ -55,6 +55,24 @@ defmodule Engine.Search.Indexer.Extractors.StructReferenceTest do
       assert decorate(doc, struct.range) == "variable = «%MyStruct{}»"
     end
 
+    test "in an any-struct match on the left side and struct on the right" do
+      {:ok, [struct], doc} = ~q[%_{} = %MyStruct{}] |> index()
+
+      assert struct.type == :struct
+      assert struct.subtype == :reference
+      assert struct.subject == Subject.module(MyStruct)
+      assert decorate(doc, struct.range) == "%_{} = «%MyStruct{}»"
+    end
+
+    test "in an any-struct match with binding on the left side and struct on the right" do
+      {:ok, [struct], doc} = ~q[%str_name{} = %MyStruct{}] |> index()
+
+      assert struct.type == :struct
+      assert struct.subtype == :reference
+      assert struct.subject == Subject.module(MyStruct)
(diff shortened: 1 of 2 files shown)


apps/engine/test/engine/search/indexer/extractors/struct_reference_test.exs
```diff
@@ -69,6 +87,24 @@ defmodule Engine.Search.Indexer.Extractors.StructReferenceTest do
...
+        |> index()
+    end
+
+    test "in an any-struct pattern with binding in params" do
+      {:ok, [], _doc} =
+        ~q[
+        def my_fn(%struct_name{} = first) do
+        end
+        ]
+        |> index()
```
doorgan:
```suggestion
      assert {:ok, [], _doc} =
        ~q[
        def my_fn(%struct_name{} = first) do
        end
        ]
        |> index()
```

apps/engine/test/engine/search/indexer/extractors/struct_reference_test.exs
```diff
@@ -69,6 +87,24 @@ defmodule Engine.Search.Indexer.Extractors.StructReferenceTest do
       assert decorate(doc, struct.range) == ~S[def my_fn(«%MyStruct{}» = first) do]
     end
 
+    test "in an any-struct pattern in params" do
+      {:ok, [], _doc} =
+        ~q[
+        def my_fn(%_{} = first) do
+        end
+        ]
+        |> index()
```
doorgan:
```suggestion
      assert {:ok, [], _doc} =
        ~q[
        def my_fn(%_{} = first) do
        end
        ]
        |> index()
```

apps/engine/test/engine/search/indexer/extractors/struct_reference_test.exs
katafrakt: I had it initially, but the formatter makes it look weird. I brought it back now.

doorgan: Seems like a bug, we should probable file an issue in the elixir repo if it's not fixed in main

PR #352, opened 2026-02-04
fix(forge): improve log when Spitfire crashes

katafrakt:
This is a bit opinionated, but I'd like to tone down a message about Spitfire crash in logs.

Over last few days I received questions from a few people about Expert vaguely not working, pointing to the Spitfire error logs in `expert.log`:

```
07:20:21.182 [debug] Node port message: 
07:20:21.182 [error] Spitfire parser crashed: ** (CaseClauseError) no case clause matching:

    :"}"

    (xp_spitfire 0.3.0) lib/spitfire.ex:892: XPSpitfire.parse_stab_expression/2
    (xp_spitfire 0.3.0) lib/spitfire.ex:1136: anonymous fn/1 in XPSpitfire.parse_do_block/2
    (xp_spitfire 0.3.0) lib/spitfire/while.ex:5: XPSpitfire.While2.recurse/3
    (xp_spitfire 0.3.0) lib/spitfire/while.ex:10: XPSpitfire.While2.recurse/3
    (xp_spitfire 0.3.0) lib/spitfire.ex:1132: anonymous fn/1 in XPSpitfire.parse_do_block/2
    (xp_spitfire 0.3.0) lib/spitfire/while.ex:5: XPSpitfire.While2.recurse/3
    (xp_spitfire 0.3.0) lib/spitfire.ex:1130: XPSpitfire.parse_do_block/2
    (xp_spitfire 0.3.0) lib/spitfire.ex:259: XPSpitfire.parse_expression/6
```

However, these messages do not affect the whole language server, they just mean that some file could not be parsed. In case of this project, these are Phoenix template files, which are actually EEx files with `.ex` extension. So I thought it could be a good idea to improve these messages with:

- Indicating the file where this happaned
- Making it a warning instead of error, because it's not critical (and even works, I think it's not fixable)
- Collapsing stacktrace into one line - this does not attract that much attention, but retains the possibility to debug further, if needed

The result:

```
12:57:15.039 [debug] Node port message: 
12:57:15.039 [warning] Spitfire crashed when parsing /Users/pawel.sw/jump/Jump/api/deps/phoenix/priv/templates/phx.gen.auth/settings_live.ex
message: "no case clause matching:\n\n    :\"}\"\n"
stacktrace:     (xp_spitfire 0.3.0) lib/spitfire.ex:892: XPSpitfire.parse_stab_expression/2\n    (xp_spitfire 0.3.0) lib/spitfire.ex:1136: anonymous fn/1 in XPSpitfire.parse_do_block/2\n    (xp_spitfire 0.3.0) lib/spitfire/while.ex:5: XPSpitfire.While2.recurse/3\n    (xp_spitfire 0.3.0) lib/spitfire/while.ex:10: XPSpitfire.While2.recurse/3\n    (xp_spitfire 0.3.0) lib/spitfire.ex:1132: anonymous fn/1 in XPSpitfire.parse_do_block/2\n    (xp_spitfire 0.3.0) lib/spitfire/while.ex:5: XPSpitfire.While2.recurse/3\n    (xp_spitfire 0.3.0) lib/spitfire.ex:1130: XPSpitfire.parse_do_block/2\n    (xp_spitfire 0.3.0) lib/spitfire.ex:259: XPSpitfire.parse_expression/6\n
```

Let me know what you think.

### DIFF
--- a/apps/forge/lib/forge/ast.ex
+++ b/apps/forge/lib/forge/ast.ex
@@ -174,7 +174,9 @@ defmodule Forge.Ast do
     |> case do
       {:error, :crashed, exception, stacktrace} ->
         log_parser_crash(exception, stacktrace, document.path)
-        {:error, {[line: 1, column: 1], "parser crashed: #{Exception.message(exception)}"}}
+
+        {:error,
+         {[line: 1, column: 1], "parser crashed: #{Exception.format_banner(:error, exception)}"}}
 
       other ->
         other
@@ -215,7 +217,9 @@ defmodule Forge.Ast do
 
       {:error, :crashed, exception, stacktrace} ->
         log_parser_crash(exception, stacktrace, document.path)
-        {:error, {[line: 1, column: 1], "parser crashed: #{Exception.message(exception)}"}}
+
+        {:error,
+         {[line: 1, column: 1], "parser crashed: #{Exception.format_banner(:error, exception)}"}}
 
       _error ->
         # https://github.com/elixir-lang/elixir/issues/12673#issuecomment-1626932280
@@ -226,7 +230,10 @@ defmodule Forge.Ast do
         case do_container_cursor_to_quoted(document_fragment) do
           {:error, :crashed, exception, stacktrace} ->
             log_parser_crash(exception, stacktrace, document.path)
-            {:error, {[line: 1, column: 1], "parser crashed: #{Exception.message(exception)}"}}
+
+            {:error,
+             {[line: 1, column: 1],
+              "parser crashed: #{Exception.format_banner(:error, exception)}"}}
 
           other ->
             other
@@ -532,11 +539,11 @@ defmodule Forge.Ast do
   end
 
   defp log_parser_crash(exception, stacktrace, path) do
-    path_info = if path, do: " when parsing #{path}", else: ""
-    stacktrace_line = Exception.format_stacktrace(stacktrace) |> String.replace("\n", "\\n")
+    path_info =
+      if path, do: " when parsing #{path}. This file will not be used in search index.", else: ""
 
     Logger.warning(
-      "Spitfire crashed#{path_info}\nmessage: #{inspect(Exception.message(exception))}\nstacktrace: #{stacktrace_line}"
+      "Spitfire crashed#{path_info}\n#{Exception.format(:error, exception, stacktrace)}"
     )
   end
 
--- a/apps/forge/lib/forge/ast/parser/spitfire.ex
+++ b/apps/forge/lib/forge/ast/parser/spitfire.ex
@@ -1,15 +1,14 @@
 defmodule Forge.Ast.Parser.Spitfire do
   @moduledoc false
 
-  defp opts do
-    [
-      literal_encoder: &{:ok, {:__block__, &2, [&1]}},
-      token_metadata: true,
-      columns: true,
-      unescape: false
-    ]
-  end
+  @type parse_error :: {location :: keyword(), message :: binary()}
+  @type comment :: map()
 
+  @spec string_to_quoted(binary()) ::
+          {:ok, Macro.t(), [comment()]}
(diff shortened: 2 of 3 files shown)


doorgan (commented):
I think what's valuable here is that it tells you which file failed to parse, and having it be a warning is a good change.

I think we *do* want it to attract attention, failing to parse valid files is a rather serious bug and we do want it reported. We do not know if it failed to parse a valid elixir file(serious bug) or it failed to parse code with errors(we need to fix but not as serious)

> However, these messages do not affect the whole language server

It does, indexing relies on the ability to parse valid elixir

Making it clear that *this happened, but we recovered from it* in the message may be more useful.

doorgan:
As a side note, we should properly handle .eex files rather than feeding them to an elixir parser, the current behavior is incorrect and will always fail for those files

katafrakt:
I understand your point, but:

> failing to parse valid files is a rather serious bug and we do want it reported

Sure, but the thing is that the errors I see are from not valid Elixir files (this is the case of a fresh Phoenix project too). So it is a kind of false positive.

> It does, indexing relies on the ability to parse valid elixir

Maybe I misspoke, but what I meant is that it simply does not parse this particular file. Does not crash the indexer as a whole or a batch of files. So when we are talking about invalid Elixir files, it's not really a problem for the user.

I was also going to say that detecting these files would be very tricky, but when writing the comment I actually figured that _maybe_ it would make sense to treat the files in `deps` specially. It's not like they are being actively changed by the user.

doorgan:
My concern is that, since we moved to Spitfire and Spitfire currently has some scenarios where it misparses valid elixir(I'm currently working on fixing all that) I worry that making those parse errors less visible will cause people to miss them and not report them.

So what I think we should do is:
1. Not make this error less visible, but word it in a way that makes it clear it's recoverable (Spitifre parser crashed makes it look more  catastrophic than it really is and I think that's what confuses people not familiar with how expert works)
2. Not try to parse EEx files

I'm good with wording the warning differently in this PR to address 1. TBH I think making it a warning and adding the file path that failed to parse is of great help already.

We also need to track and fix 2. I think it's a longstanding bug that was hidden before because we didn't report failures to parse files as prominently as we do now(because it would happen for any file while you're typing and that was normal).

mhanberg:
For releasable changes like log messages, use the `fix` commit type. Or else this (in a normal, post 0.1.0 workflow) couldn't get released until something else was merged in that would be added to the changelog.

katafrakt:
@doorgan I changed the message to:

```
14:47:35.682 [warning] Spitfire crashed when parsing /Users/pawel.sw/jump/Jump/api/deps/phoenix_live_view/deps/phoenix/priv/templates/phx.gen.auth/settings_controller.ex. This file will not be used in search index.
** (CaseClauseError) no case clause matching:

    :"}"

    (xp_spitfire 0.3.0) lib/spitfire.ex:892: XPSpitfire.parse_stab_expression/2
    (xp_spitfire 0.3.0) lib/spitfire.ex:1136: anonymous fn/1 in XPSpitfire.parse_do_block/2
    (xp_spitfire 0.3.0) lib/spitfire/while.ex:5: XPSpitfire.While2.recurse/3
    (xp_spitfire 0.3.0) lib/spitfire/while.ex:10: XPSpitfire.While2.recurse/3
    (xp_spitfire 0.3.0) lib/spitfire.ex:1132: anonymous fn/1 in XPSpitfire.parse_do_block/2
    (xp_spitfire 0.3.0) lib/spitfire/while.ex:5: XPSpitfire.While2.recurse/3
    (xp_spitfire 0.3.0) lib/spitfire.ex:1130: XPSpitfire.parse_do_block/2
    (xp_spitfire 0.3.0) lib/spitfire.ex:259: XPSpitfire.parse_expression/6
```

WDYT?

I also added some type specs to understand what Dialyzer complained about. I can remove them.

@mhanberg alright, thanks for the explanation.

PR #367, opened 2026-02-09
fix: log project's erl path

### LINKED ISSUE #365: Log path to `erl` executable
## Description

To help with bug reports, we should also log the path to the `erl` executable that the engine server will be able to find. We've seen some weird issues like #306 where we probably weren't using the correct (or maybe we used to not use the correct) `erl`.

doorgan:
This makes us check that we find an erlang executable(if we don't then we'll fail later anyways), and prints which executable we found.

Resolves #365 

There is a test failure in forge that exists on main, I'll fix that in another pr


mhanberg:
Needs rebase now, my PR took care of the deps issue 😅

doorgan:
Done!

PR #369, opened 2026-02-10
fix(engine): isolate search store from async load crashes

QuinnWilton:
This is part of a set of 4 PRs that arose out of some static analysis tooling I'm working on:
- https://github.com/elixir-lang/expert/pull/369
- https://github.com/elixir-lang/expert/pull/370
- https://github.com/elixir-lang/expert/pull/371
- https://github.com/elixir-lang/expert/pull/372

That means that these aren't crashes or issues that I've observed in practice, however based on my reading of the code, they do represent issues worth addressing.

## Problem:
`State.prepare_backend_async/2` uses `Task.async/1` to build the search index. If this task raises, the `Store` crashes, and the task silently fails. The store is supervised under a `:one_for_one` strategy, and as far as I can tell, no indexed data is permanently lost, however:

1) During the restart window, callers of the `Store` will be returned `:noproc` instead of  `{:error, :loading}`, and the crash will cascade
2) If any `project_compiled` events are dispatched between the crash and the restart, these will be missed by the `Store`. Since these events are never replayed, the event will be lost, and the server will remain unloaded until the next compilation

## Solution
By using `Task.Supervisor.async_nolink/2` to start the search index task under a dedicated `Engine.TaskSupervisor`, we can instead handle a crash in the task by logging the error, and clearing `async_load_ref`, so that the server is immediately ready to process further requests.

During this time, the server remains available, callers are isolated from the crash, no `project_compiled` events are lost while waiting for the restart, and previously loaded indexes remain loaded.

In the successful case, the monitor is flushed to avoid `:DOWN` messages building up and causing a mailbox leak.

### DIFF
--- a/apps/engine/lib/engine/search/store.ex
+++ b/apps/engine/lib/engine/search/store.ex
@@ -154,7 +154,16 @@ defmodule Engine.Search.Store do
 
   # handle the result from `State.async_load/1`
   def handle_info({ref, result}, {update_ref, %State{async_load_ref: ref} = state}) do
+    Process.demonitor(ref, [:flush])
     {:noreply, {update_ref, State.async_load_complete(state, result)}}
+  end
+
+  def handle_info(
+        {:DOWN, ref, :process, _pid, reason},
+        {update_ref, %State{async_load_ref: ref} = state}
+      ) do
+    Logger.error("Search index async load crashed: #{inspect(reason)}")
+    {:noreply, {update_ref, %State{state | async_load_ref: nil}}}
   end
 
   def handle_info(:flush_updates, {_, %State{} = state}) do
--- a/apps/engine/test/engine/search/store_test.exs
+++ b/apps/engine/test/engine/search/store_test.exs
@@ -368,6 +368,7 @@ defmodule Engine.Search.StoreTest do
   defp with_a_started_store(project, backend) do
     destroy_backend(backend, project)
 
+    start_supervised!({Task.Supervisor, name: Engine.TaskSupervisor})
     start_supervised!(Dispatch)
     start_supervised!(backend)
     start_supervised!({Store, [project, &default_create/1, &default_update/2, backend]})
@@ -400,6 +401,7 @@ defmodule Engine.Search.StoreTest do
     setup %{project: project} do
       destroy_backend(Ets, project)
 
+      start_supervised!({Task.Supervisor, name: Engine.TaskSupervisor})
       start_supervised!(Dispatch)
       start_supervised!(Ets)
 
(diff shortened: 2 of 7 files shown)


apps/engine/test/engine/search/store_test.exs
```diff
@@ -458,4 +460,37 @@ defmodule Engine.Search.StoreTest do
...
+      end)
+
+      {:ok, project: project}
+    end
+
+    test "store survives when async load task crashes" do
+      pid = Process.whereis(Store)
+      Store.enable()
+      # Allow time for the task to crash and the DOWN message to be processed.
+      Process.sleep(100)
```
mhanberg, line 491: I suggest having the function that raises send a messsage to the test process instead, and we can assert_receive on that instead of waiting.

apps/engine/lib/engine/search/store.ex
```diff
@@ -154,9 +154,18 @@ defmodule Engine.Search.Store do
...
   def handle_info({ref, result}, {update_ref, %State{async_load_ref: ref} = state}) do
+    Process.demonitor(ref, [:flush])
     {:noreply, {update_ref, State.async_load_complete(state, result)}}
   end
 
+  def handle_info(
+        {:DOWN, ref, :process, _pid, reason},
+        {update_ref, %State{async_load_ref: ref} = state}
+      ) do
+    Logger.error("Search index async load crashed: #{inspect(reason)}")
```
mhanberg, line 165:
```suggestion
    Logger.error("Failed to prepare search store backend: #{inspect(reason)}")
```

apps/engine/test/engine/search/store_test.exs
```diff
@@ -458,4 +460,37 @@ defmodule Engine.Search.StoreTest do
...
+      crashing_create = fn _project ->
+        raise "index build exploded"
+      end
+
+      start_supervised!({Store, [project, crashing_create, &default_update/2, Ets]})
+
+      assert_eventually alive?()
+
+      on_exit(fn ->
+        after_each_test(Ets, project)
```
mhanberg, line 481:
```suggestion
        destroy_backend(Ets, project)
```

mhanberg: this after each test function can probably be deleted, but we can deal with that in a different pr (not you, just as a note for myself)

doorgan (changes requested):
I'm not sure this PR fixes the issue it claims to fix.
If I'm understanding this correctly, this will prevent the Store from crashing, but will leave it in a broken state where it's alive, but nothing was loaded and nothing can be loaded either. That would be hiding a crash and for a user it would still be broken.
I think this needs more changes to get the Store into a healthy state after recovering from a load crash, or to crash even more loudly if this would render the engine unusable

doorgan:
I'm closing this one to clean up, feel free to open a new PR if you want to continue working on it

Issue #363, opened 2026-02-06
Add a functionality to collect bug report data


mkaput:
There is a need to make is rather simple UX-wise to collect bug data, so that making a great bug report is as effortless as possible.

It is TBD yet how this will exactly work. Let's discuss needs first. When we will reach a conclusion, then this issue description should be edited.

mkaput:
On highest level, what I, personally, foresee is to have a VSCode command that collect bug data and leaves a file with a report, ready to share. Why VSCode? From a pragmatic perspective - that's where most of the user base stays, I guess*,

@mhanberg has suggested using _expert-as-cli_ mode and have a command there to build a similar report. That also sounds very cool to me - my vscode command may just expose a trivial UI layer on top of that.

---

What do we need to collect?

- Expert version (with commit hash precision)
- User project's Elixir/OTP/Mix versions (also with commit hash precision)
- Editor version - whatever we get from lsp initialization protocol, verbatim.
- Operation + path+line+column if we're talking about a crash and it can be narrowed down to particular user context? (not always possible)

Thoughts?

lukaszsamson:
LSP has telemetry and VSCode has mechanism for collecting the extensions telemetry. This allows collection of extension and server crash logs and metrics to Azure Logs and App Analytics

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#telemetry_event
https://code.visualstudio.com/api/extension-guides/telemetry
https://www.npmjs.com/package/@vscode/extension-telemetry

PR #380, opened 2026-02-12
feat: add `instance_id` metadata to logs

Draggu:
This allows us to easily grab logs from a single expert run.

Also, make logs single-line where possible for future ease of parsing

### DIFF
--- a/apps/expert/config/runtime.exs
+++ b/apps/expert/config/runtime.exs
@@ -17,5 +17,6 @@ config :logger, :default_handler,
   config: [
     file: String.to_charlist(log_file_name),
     max_no_bytes: :infinity,
-    max_no_files: 0
+    max_no_files: 0,
+    formatter: Logger.Formatter.new(metadata: [:instance_id])
   ]
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -204,7 +204,16 @@ defmodule Expert.EngineNode do
     start_net_kernel(project)
 
     node_name = Project.node_name(project)
-    bootstrap_args = [project, Document.Store.entropy(), all_app_configs(), Node.self()]
+
+    bootstrap_args = [
+      project,
+      Document.Store.entropy(),
+      all_app_configs(),
+      Node.self(),
+      # Copy logger global metadata to engine instances.
+      # Everything spawned from single expert instance will use same `instance_id`
+      :logger.get_primary_config().metadata
+    ]
 
     with {:ok, node_pid} <- EngineSupervisor.start_project_node(project),
          {:ok, glob_paths} <- glob_paths(project),
--- a/apps/engine/lib/engine/bootstrap.ex
+++ b/apps/engine/lib/engine/bootstrap.ex
@@ -11,7 +11,15 @@ defmodule Engine.Bootstrap do
 
   require Logger
 
-  def init(%Project{} = project, document_store_entropy, app_configs, manager_node) do
+  def init(
+        %Project{} = project,
+        document_store_entropy,
+        app_configs,
+        manager_node,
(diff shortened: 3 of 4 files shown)


apps/expert/lib/expert/engine_node.ex
```diff
@@ -187,7 +187,14 @@ defmodule Expert.EngineNode do
...
 
     node_name = Project.node_name(project)
-    bootstrap_args = [project, Document.Store.entropy(), all_app_configs(), Node.self()]
+
+    bootstrap_args = [
+      project,
+      Document.Store.entropy(),
+      all_app_configs(),
+      Node.self(),
+      :logger.get_primary_config().metadata
```
mkaput, line 215: I miss a comment explaining why we pass metadata here

apps/expert/lib/expert/engine_node.ex
```diff
@@ -426,7 +433,9 @@ defmodule Expert.EngineNode do
 
   @impl true
   def handle_info({_port, {:data, data}}, %State{} = state) do
-    message = to_string(data)
+    message =
+      data |> to_string() |> String.trim() |> (&Regex.replace(~r/\r\n|\n|\r/, &1, ", ")).()
```
mkaput: why?

Draggu: I want logs to be single-line if possible for future parsing

apps/expert/lib/expert/engine_node.ex
Draggu: Done

apps/expert/lib/expert/engine_node.ex
katafrakt: I don't have strong opinions on this, but it was discussed in https://github.com/elixir-lang/expert/pull/352#issuecomment-3849974444 that some log messages should stand out to the reader. Collapsing them into one line is probably is the opposite direction of that.

mhanberg: Can you expand what you mean by future parsing?

Draggu: https://github.com/elixir-lang/expert/issues/363
When creating the report command, we will need to retrieve entries from a single instance. So we need to parse the log file and filter out other entries.

Draggu: I don’t see this as a readability issue, logs are still clear enough. If you’d rather do it another way, I’m fine with that.

@mhanberg if the current version is fine, can we merge?

mhanberg: Removing new lines from logs is an extreme readability issue. Especially if they are stack traces. Also, users probably want to be able to redact things from the logs before sending them to us, so having them legible will help I think.

I think its probably enough to add the instance id and then bundle all the logs rather than parsing them and only doing that session. we can easily scroll to the top to find the first line.

apps/expert/config/runtime.exs
```diff
@@ -16,7 +16,8 @@ if Code.ensure_loaded?(LoggerFileBackend) do
 
   config :logger, :general_log,
     path: log_file_name,
-    level: :debug
+    level: :debug,
+    metadata: [:instance_id]
```
mhanberg: I think this can be `metadata: [instance_id: Integer.to_string(System.os_time(:millisecond), 16)]` and you don't have to have lines 16-18 in apps/expert/lib/expert/application.ex. but you'll have to double check

Draggu: It does not work, `metadata` expects an atom list

mhanberg:
> I recall there was a line that turned logs into a single line for "future parsing" but I don't see it anymore

I had said I think it's a readability issue. I believe he removed it after.

Issue #403, opened 2026-02-17
Cannot build on nix


Alt-iOS:
When building on nix-darwin, the build fails, because it tries to link to a read-only directory, here are the logs:
```nix
┃        > Running phase: configurePhase
┃        > ln: failed to create symbolic link 'apps/engine/deps/erlang27.3.4.7/src': Permiss…
┃        For full logs, run:
┃          nix log /nix/store/ghppb22sh3md9whhpz89v5dm7nipbsbh-expert-0.1.0-.drv
```
It seems to me that tries to create the link on a src directory directly

Alt-iOS:
This happened when running `nh darwin switch`. I did a quick fix on my local machine, and seems to work, created a PR in case its useful. I am not that familiar with nix packaging, so if theres a better way lmk and I can update the PR

mhanberg:
This is odd, since i use it on nix with darwin everyday.

But, i use it via home-manager, but I don't think that should make a difference.

Did you edit the logs you shared? I am confused by the path `apps/engine/deps/erlang27.3.4.7/src`

Alt-iOS:
I can share you the full log
```nix
ln: failed to create symbolic link 'apps/engine/deps/erlang27.3.4.7/src': Permission denied
error: Cannot build '/nix/store/ghppb22sh3md9whhpz89v5dm7nipbsbh-expert-0.1.0-.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/w93xxv42zfcsfjhmnhjmhfkpc701c6sx-expert-0.1.0-
       Last 7 log lines:
       > Running phase: unpackPhase
       > unpacking source archive /nix/store/blxqqyp9712l3i58shbv7qc3vmc803bh-source
       > source root is source
       > Running phase: patchPhase
       > Running phase: updateAutotoolsGnuConfigScriptsPhase
       > Running phase: configurePhase
       > ln: failed to create symbolic link 'apps/engine/deps/erlang27.3.4.7/src': Permission denied
       For full logs, run:
         nix log /nix/store/ghppb22sh3md9whhpz89v5dm7nipbsbh-expert-0.1.0-.drv
error: Cannot build '/nix/store/3q8r40qz9h6nmf5gx58vgy0l8gjn4l4v-expert-0.1.0--fish-completions.drv'.
       Reason: 1 dependency failed.
       Output paths:
[...]

PR #404, opened 2026-02-17
fix: nix build fails in preConfigure phase

Alt-iOS:
It creates a temporary writable dir for linking in the preCofnigure phase, instead of linking directly to source which because its a read-only didnt allow it, deleting if it previously existed. Fixes issue at #403

### DIFF
--- a/nix/expert.nix
+++ b/nix/expert.nix
@@ -33,14 +33,24 @@ beamPackages.mixRelease rec {
 
   preConfigure = ''
     # copy the logic from mixRelease to build a deps dir for engine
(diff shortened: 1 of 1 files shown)


mhanberg:
@adamcstephens could you take a look at this?

adamcstephens:
I'd be surprised if this is an actual fix? Anyway I added more context in #403.

Alt-iOS:
It does build in my local mac and the expert binary is available and working haha, but like you said in the #403, it was failing because of the overriding of `inputs.nixpkgs.follow`

PR #405, opened 2026-02-17
feat: prompt user to fetch deps when they get out of sync

doorgan:
Adds a prompt for the user to fetch dependencies if we encounter a dependency issue when compiling their project.

I copied the behavior of NextLS here, so if you say "yes" to the prompt we will:
- remove build artifacts
- run mix deps.get for you
- start the engine again

I had to add another state tracking to `ActiveProjects` because stopping the engine after the deps error would get it restarted upon document notifications and other random functions. Ready wasn't enough since they represent different things.

I'm not sure what to do if the user says "no" or dismisses the message though, because the engine for that project won't be restarted. They'll have to restart the LSP. Maybe that's ok?

To reproduce the issue this PR addresses:
- Add `{:sourceror, "~> 0.5"}` to your deps and have expert initialize and compile it
- Change it to `{:sourceror, "~> 1.10"}`, with this PR you'll see the prompt

### DIFF
--- a/apps/expert/config/test.exs
+++ b/apps/expert/config/test.exs
@@ -1,5 +1,5 @@
 import Config
 
-# Withuot this option, ExUnit would randomly not terminate or tests would behave
+# Without this option, ExUnit would randomly not terminate or tests would behave
 # erratically.
 config :gen_lsp, :exit_on_end, false


mhanberg:
> I'm not sure what to do if the user says "no" or dismisses the message though, because the engine for that project won't be restarted. They'll have to restart the LSP. Maybe that's ok?

In Next LS I just had the engine shut down and it operated in a Degraded state after that. Basically just document symbols. Well in next LS everything except autocomplete and hover basically operated on the server so definition and references still worked, but they wouldn't update as you made changes.

doorgan:
@mhanberg we can do that(just did that in latest push), the engine continues to run and to provide completions etc, but they need to fetch deps to get everything working properly again


On a side note, there is a test that is failing consistently in this PR but is unrelated to the changes. I've seen it fail once or twice locally but not on CI, I'll take a look tomorrow and fix in a separate pr.

doorgan:
I don't yet get why, but some tests that pass on main CI were failing in this PR. I suspect it has something to do with the change in the `gen_lsp` config(the windows path one certainly does). They only manifest in this PR and they are small changes so I decided to fix that here instead of a different PR.

katafrakt:
This is a fantastic UX improvement! I tested it in VSCode and it works flawlessly.

However, in Zed and Emacs the prompt shows twice for me:

Screenshot 2026-02-19 at 12 47 28

And it's indeed coming twice from the server:

```
[jsonrpc] e[13:00:07.602] <-- window/showMessageRequest[24] {"id":24,"jsonrpc":"2.0","method":"window/showMessageRequest","params":{"actions":[{"title":"yes"},{"title":"no"}],"message":"The Expert engine failed with errors on dependencies for project zlobex. Would you like to fetch them?","type":1}}
[jsonrpc] e[13:00:07.622] <-- $/progress {"jsonrpc":"2.0","method":"$/progress","params":{"token":1858,"value":{"kind":"end"}}}
[jsonrpc] e[13:00:07.622] <-- window/showMessageRequest[25] {"id":25,"jsonrpc":"2.0","method":"window/showMessageRequest","params":{"actions":[{"title":"yes"},{"title":"no"}],"message":"The Expert engine failed with errors on dependencies for project zlobex. Would you like to fetch them?","type":1}}

```

apps/expert/lib/expert.ex
```diff
@@ -244,13 +246,134 @@ defmodule Expert do
...
+        GenLSP.request(
+          lsp,
+          %Requests.WindowShowMessageRequest{
+            id: Id.next(),
+            params: %Structures.ShowMessageRequestParams{
+              type: Enumerations.MessageType.error(),
+              message:
+                "The Expert engine failed with errors on dependencies for project #{project_name}. Would you like to fetch them?",
+              actions: [
+                %Structures.MessageActionItem{title: "yes"},
```
mhanberg: Let's make these capital Yes and No, rather than yes and no

But also wondering if Allow and Deny are better labels

doorgan: I like Yes/No more, Allow/Deny sounds to me like language for permissions, while this case is about course of action

mhanberg: Roger 👍

doorgan:
@katafrakt should be fixed now

apps/expert/lib/expert.ex
```diff
@@ -244,13 +246,137 @@ defmodule Expert do
...
+          "Engine failed due to dependency errors. Run 'mix deps.get' to fetch dependencies for #{Project.name(project)}."
+        )
+      end
+
+      if !supports_show_message do
+        log_error(
+          lsp,
+          project,
+          "Engine failed due to dependency errors, but client does not support showing messages. Run 'mix deps.get' to fetch dependencies for #{Project.name(project)}."
+        )
```
mhanberg, line 321: The message should also say to fetch and then restart the LSP/editor

katafrakt:
@doorgan unfortunately I still see it twice in both Emacs and Zed. I might be able to look into this a bit later today or tomorrow, if you want.

doorgan:
@katafrakt that's odd, I can't reproduce in zed.
Yeah let's wait until you can debug it, I'll keep trying to reproduce on my end

mhanberg:
I wonder if maybe its reacting to both of the potential dependency errors strings and then sending two requests?

katafrakt:
Perhaps a naive fix is to check whether we already recorded a `deps_error` in state before sending the notification

```
diff --git a/apps/expert/lib/expert/engine_node.ex b/apps/expert/lib/expert/engine_node.ex
index a7945780..08d81f0c 100644
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -448,8 +448,10 @@ defmodule Expert.EngineNode do
     Logger.debug("Node port message: #{message}")
 
     if State.detect_deps_error(message) do
-      if lsp = Expert.get_lsp() do
-        send(lsp.pid, {:deps_error, state.project, %{last_message: message}})
+      if not state.deps_error do
+        if lsp = Expert.get_lsp() do
+          send(lsp.pid, {:deps_error, state.project, %{last_message: message}})
+        end
       end
 
       {:noreply, %{state | last_message: message, deps_error: true}}
```

(verified in Zed)

doorgan:
I'm still not able to reproduce that issue, but your suggestion makes sense so I added it

katafrakt:
Checked Zed and Emacs, works great now!

apps/expert/config/test.exs
```diff
@@ -1 +1,5 @@
 import Config
+
+# Withuot this option, ExUnit would randomly not terminate or tests would behave
```
mhanberg:
```suggestion
# Without this option, ExUnit would randomly not terminate or tests would behave
```

mhanberg (approved):

PR #407, opened 2026-02-18
fix: provide typespec docs on hover for private functions

doorgan:
There is a test that is failing consistently for me and I'm able to reproduce it:

```
  1) test call hover private function (Expert.Provider.Handlers.HoverTest)
Error:      test/expert/provider/handlers/hover_test.exs:482
     match (=) failed
     code:  assert {:ok, %Structures.Hover{} = result} = hover(project, hovered)
     left:  {:ok, %GenLSP.Structures.Hover{} = result}
     right: {:ok, nil}
     stacktrace:
       test/expert/provider/handlers/hover_test.exs:495: anonymous fn/1 in Expert.Provider.Handlers.HoverTest."test call hover private function"/1
       test/expert/provider/handlers/hover_test.exs:72: anonymous fn/3 in Expert.Provider.Handlers.HoverTest.with_compiled_in/3
       test/expert/provider/handlers/hover_test.exs:91: Expert.Provider.Handlers.HoverTest.with_tmp_file/3
       test/expert/provider/handlers/hover_test.exs:494: (test)
```

Given code like this:

```elixir
defmodule MyModule do
  @spec my_fun(integer()) :: integer()
  defp my_fun(x), do: x + 1
  def my_other_fun(x, y), do: my_fun(x) + my_fun(y)
end
```

Hovering on the `my_fun/1` calls inside `my_other_fun/2` provides no typespecs on hover. 
The current code has the specs for the private functions available, but doesn't use them. It only uses the docs for functions with doc chunks(which makes sense at first glance) and discards the rest.
The fix in this PR is to gather the specs for functions with no docs and produce entries for them too.

### DIFF
--- a/apps/engine/lib/engine/code_intelligence/docs.ex
+++ b/apps/engine/lib/engine/code_intelligence/docs.ex
@@ -49,7 +49,9 @@ defmodule Engine.CodeIntelligence.Docs do
           module: module,
           doc: Entry.parse_doc(module_doc),
           functions_and_macros:
-            parse_entries(module, function_entries ++ macro_entries, spec_defs),
+            parse_entries(module, function_entries ++ macro_entries, spec_defs,
+              include_not_documented?: true
+            ),
           callbacks: parse_entries(module, callback_entries, callback_defs),
           types: parse_entries(module, type_entries, type_defs)
         }
@@ -65,7 +67,7 @@ defmodule Engine.CodeIntelligence.Docs do
     kind
   end
 
-  defp parse_entries(module, raw_entries, defs) do
+  defp parse_entries(module, raw_entries, defs, opts \\ []) do
     defs_by_name_arity =
       Enum.group_by(
         defs,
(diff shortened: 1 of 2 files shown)


apps/engine/lib/engine/code_intelligence/docs.ex
```diff
@@ -73,13 +73,51 @@ defmodule Engine.CodeIntelligence.Docs do
...
+      normalize_signature("#{name}#{rest}")
+    else
+      _ -> "#{name}/#{arity}"
+    end
+  end
+
+  defp normalize_signature(signature) do
+    # Remove parenthesis for the function head signature, e.g. `foo(integer())`
+    # becomes `foo(integer)`.
+    Regex.replace(~r/([A-Za-z0-9_\.]+)\(\)/, signature, "\\1")
```
mkaput: but do we want `SomeStruct.t()` to become `SomeStruct.t`? `t()` is more common to my knowledge

doorgan: This is for the signature, not the spec, but you still raise a good point.
The signature is something like `MyModule.my_fun(arg1)`, the spec is the actual typespec.
For the signature we don't want to actually refer to the spec but rather to signal that it takes an argument, so I updated to show arg1, arg2, etc

Screenshot 2026-02-18 at 5 58 21 AM

PR #418, opened 2026-02-20
fix: forward logs through window log handler

### LINKED ISSUE #382: Log more with window/logMessage
## Problem

We often have to ask users to check the log file, but that is an un-intuitive step when they usually can check an in editor log panel (like the Output Panel in VSCode). We currently do most of our logging to the regular logger, but we should also be logging things with window/logMessage.

## Solution

Couple options

- Add a logger handler that calls GenLSP.log
- Create a logger process that calls normal logger and GenLSP logger

## Notes

- need to think about the server logs as well as the engine logs
- the server logs always go to the expert processes current working directory, but the project logs go into the actual projects root directory. these can be different depending on how the editor starts the LSP. For example, Neovim will search up from the file to find the root directory to send in the intialize request, but still starts the editor in neovims current working directory. So for example if you start neovim in the expert repo, it'll put the server logs in `.expert/expert.log`, but the project logs in `apps/engine/.expert/project.log`.
    - there is also something weird about if it puts the .gitignore in that folder. i dont think its putting it in the root .expert folder, only the project .expert folder, need to double check that.

## Questions

- Do we need the sasl logs? they take a lot of space whenever the apps starts and i dont think it helps most of the time.
- What log level should we log at? I believe it currently is set to debug in the server logs, i dont know about the project logs. It would probably just be info.

## Ideas

- we should log the locations of the log files, so they can see them in the editors log panel to go find them.

mkaput:
Add a centralized logger-to-LSP bridge for window/logMessage instead of ad hoc GenLSP.log/info/error calls. This routes operational logs through one path and keeps window/showMessage for explicit errors.

Implement Expert.Logging.WindowLogHandler with runtime attachment, recursion guards, OTP/SASL/report filtering, and OTP 25/26+ compatibility via Expert.Logging.LoggerHandler. Wire attachment during app boot and test bootstrap, and route engine init events to lsp.pid so non-default test servers receive them.

Add client-based gating for Emacs/Eglot in configuration (issue #382 workaround), plus tests covering config behavior, handler idempotency/replacement, and Emacs suppression semantics.

fix #382

https://github.com/user-attachments/assets/aa84e338-7d4e-4850-aef0-c7219102441e

### DIFF
--- a/apps/expert/test/expert_test.exs
+++ b/apps/expert/test/expert_test.exs
@@ -9,6 +9,11 @@ defmodule Expert.ExpertTest do
 
   setup do
     :persistent_term.erase(Expert.Configuration)
+    with_patched_transport()
+
+    # These tests call `Expert.handle_info/2` directly (bypassing `Expert.Application`),
+    # so we must start `Expert.ActiveProjects` to create its ETS tables first.
+    start_supervised!({Expert.ActiveProjects, []})
 
     # window/logMessage comes from Logger via WindowLogHandler,
     # so tests that assert on log notifications must allow :info events through.
@@ -22,9 +27,6 @@ defmodule Expert.ExpertTest do
   end
 
   test "sends an error message on engine initialization error" do
-    with_patched_transport()
-    start_supervised!({Expert.ActiveProjects, []})
-
     project = Fixtures.project()
     lsp = initialize_lsp(project)
 
@@ -54,8 +56,6 @@ defmodule Expert.ExpertTest do
   end
 
   test "suppresses window/logMessage for emacs client" do
-    with_patched_transport()
-
     project = Fixtures.project()
     lsp = initialize_lsp(project, client_name: "Emacs")
     reason = :something_bad


apps/expert/lib/expert/logging/window_log_handler.ex
```diff
@@ -0,0 +1,149 @@
...
+
+  @metadata_recursion_key :expert_window_log_handler
+  @excluded_log_domains [:otp, :sasl]
+
+  @doc """
+  Registers the logger handler in a release-safe way.
+
+  We intentionally register the handler with `__MODULE__` at runtime instead of `Logger.add_handlers/1` config.
+  In namespaced releases, we observed app env config retaining `Expert.Logging.WindowLogHandler` while the runtime
+  module is `XPExpert.Logging.WindowLogHandler`, causing handler registration to fail.
```
doorgan, line 26: I'm surprised by this, I expected this to have been namespaced, there might be a bug in config namespacing?

PR #422, opened 2026-02-20
fix: build expert on latest nixpkgs

### LINKED ISSUE #403: Cannot build on nix

idabmat:
Fixes #403
Picking up on the changes from [NixOS/nixpkgs#475224](https://github.com/NixOS/nixpkgs/pull/475224)

I don't know if there are more changes that @adamcstephens wants to
include. But at least it now builds on latest nixpkgs.

I've refrained from updating the `flake.lock` to avoid any unforeseen
changes though. Let me know if I should bump it.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -10,11 +10,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1763144771,
-        "narHash": "sha256-VWF3RK2lNPX5tz14LnJM7DXkP1YwgrTuu3RqeVEymCk=",
+        "lastModified": 1771594669,
+        "narHash": "sha256-bjqfo3IobpY0se/LOL6U7J0AodvMTZjZad1x3XLc7HY=",
         "owner": "elixir-tools",
         "repo": "nix-beam-flakes",
-        "rev": "0927b7964669ff9301e90ec5753ac81885a4b1bf",
+        "rev": "da950d6652b5b4bc43ee7c440a39097b4d96e448",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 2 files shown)


adamcstephens:
I started working on this the other day, but got sidetracked going a bit deeper on some other improvements that @mhanberg and I were discussing on discord.

The `flake.lock` is pretty old in this repo. I'd probably recommend going ahead and including that update if you're comfortable doing so. While I think this PR is probably backwards compatible, it's simpler to support one implementation on the nixpkgs side.

Eventually nixpkgs will even have a hook for this deps dir building, but not until I get to refactoring mixRelease.

mhanberg:
Going to test this locally and with update nixpkgs.

idabmat:
Updated the PR with an updated `flake.lock` file.

mhanberg:
@adamcstephens do you know why it seems to tack on an extra dash in the derivation name? `expert-0.1.0-rc.0-`

adamcstephens:
> @adamcstephens do you know why it seems to tack on an extra dash in the derivation name? `expert-0.1.0-rc.0-`

This is something in the nixpkgs builder. Probably mixRelease. But I haven’t tracked it down yet.

PR #438, opened 2026-02-23
ci: check if new commits since last nightly release

### LINKED ISSUE #430: Only publish nightly release if new commits
## Description

We currently publish  new nightly everyday no matter what. 

We should try and only publish if there have been new commits since the last release.

MikaelFangel:
Fixes #430 

I have created two dependent jobs where the process is

1. Check that `nightly` tag exists
```bash
git rev-parse nightly >/dev/null 2>&1
```

2. Get commits since last
```bash
git log nightly..HEAD --oneline
```
3. Check if the output of the commits since last is empty
4. Create an output with result
5. Check in the `Nightly build` job if it should run based on the output from the `Check for new commits` job

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -5,8 +5,31 @@ on:
     - cron: 0 0 * * *
 
 jobs:
+  check:
+    name: Check for new commits
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


mhanberg:
Thanks!

mhanberg:
image

It works!

Issue #446, opened 2026-02-24
Use custom MIX_HOME, HEX_HOME, and REBAR_CACHE_DIR for engine


mhanberg:
## Description

Usually hex and rebar are installed into a global directory at ~/.mix, which i believe is creating a bunch of vague problems with building the engine like in #385.

Setting these should inside the mix build path of the engine should allow us to always ensure we have a compatible one installed and it wont be removed by the user accidentally.

mhanberg:
Closed by #454

PR #454, opened 2026-02-24
fix: isolate engine builds and retry them on dependency errors

### LINKED ISSUE #444: Retry engine build if it fails with dependency error
## Description 

If the engine build is interrupted (or is a local release and the dependencies have been updated) the next time it builds/loads it will fail. 

At this point it requires manual intervention. 

Instead, we should run it again with force: true (MIX_INSTALL_FORCE=1).

doorgan:
Fixes #444 

This started as *just* a fix for #444 with a minor refactor to move the engine builder to a different module and GenServer to make it easier to read and test, but I was *constantly* running into various issues so I had to address several related issues too, namely:

- Hex archives would constantly get corrupted as I tried expert on different repos with different versions. The fix seems to be a combination of isolating the engine builds on a different mix home, and to pass `--if-missing` to project builds
- Hex would sometimes not be able to resolve dependencies, it seemed to be related to the above point, and was apparently fixed by addressing #446 
- I got to reproduce #431 at some point, resetting `MIX_ENV` solved that
- I would get weird issues where maybe the engine builds, but then I got hex archive corruption errors on the project build and viceversa, making them share the same `MIX_HOME` fixed that

After those changes I got it stable, and was able to run several editor instances on several repos, run tests, build releases etc without running into hex archive issues again.

As for the changes for #444 the way I could reproduce it to produce this PR was a bit contrived:
1. Build expert with `just release`
2. Find the `start_expert` script in the plain build folder, and add `export EXPERT_ENGINE_PATH = "/Users/dorgan/dev/expert/apps/engine"`
3. Change the `mix.exs` file for the engine to add a dependency
4. Start an editor and see Expert fail to build the engine without the change, and to see it retry with `force: true` and succeed on the second attempt

### DIFF
--- a/apps/expert/priv/build_engine.exs
+++ b/apps/expert/priv/build_engine.exs
@@ -14,8 +14,8 @@ force? = Keyword.get(args, :force, false)
 
 expert_data_path = :filename.basedir(:user_data, "Expert", %{version: expert_vsn})
 
-elixir_otp_vsn = "elixir-#{System.version()}-otp-#{:erlang.system_info(:otp_release)}"
-tooling_path = Path.join([expert_data_path, "tooling", elixir_otp_vsn])
+elixir_erts_vsn = "elixir-#{System.version()}-erts-#{:erlang.system_info(:version)}"
+tooling_path = Path.join([expert_data_path, "tooling", elixir_erts_vsn])
 
 System.put_env("MIX_INSTALL_DIR", expert_data_path)
 System.put_env("MIX_HOME", Path.join(tooling_path, "mix_home"))


mhanberg:
Since we now have the logger handler for window/logMessage, I think you'll need to convert any usages of GenLSP.info/log/warning/error in this code to just use Logger.

apps/expert/lib/expert/engine_node/builder.ex
```diff
@@ -0,0 +1,213 @@
...
+            {:ok, port} ->
+              %State{state | port: port}
+
+            _ ->
+              state
+          end
+
+        {:noreply, %State{state | attempts: state.attempts + 1}}
+      else
+        Logger.error("Maximum build attempts reached. Failing the build.")
```
mhanberg: I wonder if we can put the project in the logger metadata and in the handler, prepend the `[<project>]` like we do in the other logs, but make it consistent.

apps/expert/lib/expert.ex
```diff
@@ -489,7 +489,7 @@ defmodule Expert do
 
   defp initialization_error_message({:shutdown, {:failed_to_start_child, child, reason}}) do
     case child do
-      {Project.Node, node_name} ->
+      {Expert.Project.Node, node_name} ->
```
mhanberg: Was this just never matching before?

apps/expert/lib/expert/engine_node/builder.ex
```diff
@@ -0,0 +1,213 @@
...
+        )
+
+        {:stop, :normal, state}
+      end
+    else
+      {:noreply, %State{state | last_line: line}}
+    end
+  end
+
+  def handle_info({_port, {:exit_status, status}}, state) do
```
mhanberg, line 105: Does this message only come in if it fails in some way?

apps/expert/lib/expert/engine_node.ex
```diff
@@ -216,16 +222,24 @@ defmodule Expert.EngineNode do
     ]
 
     with {:ok, node_pid} <- EngineSupervisor.start_project_node(project),
-         {:ok, glob_paths} <- glob_paths(project),
+         {:ok, {glob_paths, mix_home}} <- prepare_engine(project),
          :ok <- Progress.report(token, message: "Starting Erlang node..."),
-         :ok <- start_node(project, glob_paths),
+         :ok <- start_node(project, glob_paths, mix_home: mix_home),
```
mhanberg, line 227: Do we need the build script to send back the hex home and rebar cache paths as well?

apps/expert/lib/expert/engine_node/builder.ex
doorgan: Yes, we're passing `:exit_status` to `Port.open` and we weren't handling this return value (we do for project builds). I hit `{:exit_status, 1}` a few times working on this

apps/expert/lib/expert.ex
doorgan: This is a leftover of a larger refactor I was planning on doing and discarded, I'll revert it. I was planning on changing `Expert.EngineNode` to `Expert.Engine.Node`(same for the other `Expert.Engine*` modules) but I didn't see it improve readability *and* modules names `Node` or `Supervisor` would conflict with a lot of other things when aliased so I had to fully qualify them. I ultimately discarded that and this is a leftover.

apps/expert/lib/expert/engine_node.ex
doorgan: These don't touch the archives as far as I can see, and if I understand correctly, keeping the user's `HEX_HOME` lets Expert user their private repos when building their project if they are using Oban Pro for example. `REBAR_CACHE_DIR` didn't seem to cause issues either

doorgan: The issue this is trying to fix is that without isolating the `MIX_HOME` we'd try to use `~/.mix`, which on a forced hex build would overwrite the user's archive and cause crashes. I still don't 100% understand why the engine build caused that though

doorgan: #459 is a perfect example of the kind of crashes I was seeing before this

apps/expert/lib/expert/engine_node/builder.ex
doorgan: Done!

apps/expert/lib/expert/engine_node/builder.ex
```diff
@@ -0,0 +1,217 @@
...
+
+    def close_port(_port), do: :ok
+  else
+    # In dev and prod environments, the engine source code is included in the
+    # Expert release, and we build it on the fly for the project elixir+opt
+    # versions if it was not built yet.
+    defp start_build(%Project{} = project, from, opts \\ []) do
+      with {:ok, elixir, env} <- Expert.Port.project_executable(project, "elixir"),
+           {:ok, erl, _env} <- Expert.Port.project_executable(project, "erl") do
+        Logger.info("Using path: #{System.get_env("PATH")}", project: project)
```
mhanberg, line 154:
```suggestion
        Logger.info("Using PATH: #{System.get_env("PATH")}", project: project)
```

apps/expert/lib/expert/logging/window_log_handler.ex
```diff
@@ -108,13 +109,14 @@ defmodule Expert.Logging.WindowLogHandler do
     end
   end
 
-  defp extract_message(%{msg: {:string, message}}) do
+  defp extract_message(%{msg: {:string, message}, meta: metadata}) do
```
mhanberg, line 112: I generally dont pattern match on keyword lists in function clauses because its the same as `[{:meta, metadata}]` which only matches if its a single item list with the exact tuple

doorgan: It's a map key though, the `}` before the comma there is from `{:string, message}`

apps/expert/lib/expert/engine_node.ex
```diff
@@ -216,16 +222,24 @@ defmodule Expert.EngineNode do
...
          :ok <- Progress.report(token, message: "Bootstrapping engine..."),
          :ok <- :rpc.call(node_name, Engine.Bootstrap, :init, bootstrap_args),
          :ok <- ensure_apps_started(node_name, token) do
       {:ok, node_name, node_pid}
     end
   end
 
+  defp prepare_engine(project) do
+    Expert.Progress.with_progress("[#{Project.name(project)}] Preparing engine", fn _token ->
```
mhanberg, line 236: in a later pr, perhaps we could mod this to be like `Expert.Progress.with_progress(project_name, text, callback)`

apps/expert/lib/expert/logging/window_log_handler.ex
mhanberg: oh... i should learn to read

mhanberg:
Trying this out, i didnt have a previous engine build for the elixir/erl combo, it succesfully built the engine, but output this warning to my editor

```
   Error  08:13:42 AM notify.error [blog] Failed to start child {XPExpert.Project.Node, "blog"}: {:error, "Build script exited with status: 0", "Rewriting 0 config scripts.\n"}
```

I think the exit_status sends the 0 exit code when it is a success.

It still correctly continued on tho.

mhanberg:
I rm'd a dep from that engine build and opened the editor back up, it correctly failed and retried

```
[Log] [blog] Engine build output: Unchecked dependencies for environment dev:
* refactorex (Hex package)
  the dependency is not available, run "mix deps.get"
[Warning] [blog] Detected dependency errors during engine build, retrying... (attempt 1/1)
```

mhanberg:
I noticed that for the engine we namespace the directory by erts version, but the tooling we do it by OTP version, should those be consistent?

```
└  expert engine ls
/home/mitchell/.local/share/Expert/0.1.0-rc.2/elixir-1.17.3-erts-15.2.7
/home/mitchell/.local/share/Expert/0.1.0-rc.2/elixir-1.18.4-erts-16.0.1
/home/mitchell/.local/share/Expert/0.1.0-rc.2/tooling

┌ 08:21:48 AM
│mitchell in 🌐 nublar in blog on  main +3 [!] via 🥟 v1.2.20 via  v1.18.4 (OTP 28) via  impure (nix-shell-env)
└  ls /home/mitchell/.local/share/Expert/0.1.0-rc.2/tooling
elixir-1.18.4-otp-28
```

doorgan:
@mhanberg good catches

re: status code 0, yes that's the success code that's why it still got the paths and carried on, we don't need to log anything there so I added a clause for that.

re: tooling path, good catch, I fixed that

PR #458, opened 2026-02-25
fix: add Project.unique_name to allow multiple project with same root folder name

### LINKED ISSUE #450: Ignores umbrella sub-app with same name as folder of umbrella
We have an umbrella with a few sub-apps, and where a sub-app is a dependency of the `_web` app that matches the name of the Git repo (and so becomes the default folder name).

e.g.
folder: my_app

top-level mix.exs: 
```elixir
defmodule MyAppUmbrella.MixProject do
  use Mix.Project

  def project do
    [
      name: "MyApp",
      apps_path: "apps",
      # ...
```

In "apps/" we have `my_app_web` which has an `in_umbrella` dep to `my_app`, also in "apps/".

If we freshly delete `.expert/` and restart the LSP, it will work when in a `my_app_web` file/module (go-to-def, format, etc), and even go-to-def to a `my_app` module. But once in that `my_app` module, it will not format or seem to process anything, and seems to ignore as though it were not started (sorry, I forgot to grab the logs, but can grab some later if needed).

If I then delete `.expert/` and rename the top-level folder to `my_app_umbrella` it works (at least in VSCode. Still need to test in Zed and nvim).

I'm assuming this is a name collision with the umbrella's folder name, but let me know if I can provider any more debugging. (I'll try and get the exact log later)

katafrakt:
This attempts to fix #450 by introducing `Project.unique_name`, including a hash of the absolute path of the project.

When two or more mix projects with the same containing folder are present, Expert attempts to start two supervisors with the same name. Upon failing to start the second one, it assumes it was already started. This leads to LSP partially not working (the second project was not indexed, for example).

Here we add `Project.unique_name`, which is a name of the containing folder + a hash of the full absolute path.

The situation where this issue was detected was an umbrella app inside `my_app` directory with `apps/my_app` inside. Now, since these two mix projects have different absolute paths, they will have different `unique_name` and two supervisors will correctly be started.

### DIFF
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -476,7 +476,7 @@ defmodule Expert.EngineNode do
   end
 
   def name(%Project{} = project) do
-    :"#{Project.name(project)}::node_process"
+    :"#{Project.unique_name(project)}::node_process"
   end
 
   @deps_apps Mix.Project.deps_apps()
--- a/apps/expert/lib/expert/engine_supervisor.ex
+++ b/apps/expert/lib/expert/engine_supervisor.ex
@@ -8,7 +8,7 @@ defmodule Expert.EngineSupervisor do
 
   def child_spec(%Project{} = project) do
     %{
-      id: {__MODULE__, Project.name(project)},
+      id: {__MODULE__, Project.unique_name(project)},
       start: {__MODULE__, :start_link, [project]}
     }
   end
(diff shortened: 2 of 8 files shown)


mhanberg:
Shouldn't we be starting a single engine for the whole umbrella project?

I feel like this implies we're starting one for the umbrella and the child app.

katafrakt:
Yes, from my tests we are starting engine for every sub-app. I can try to see how we could not do that for umbrellas.

mhanberg:
Roger, I feel like maybe that is a hold over from before we had workspace folders. I dont think we should be doing that at all.

katafrakt:
I'll leave this one open anyway, because I think it also fixes some unorthodox setups, such as a following monorepo:

```
service1/
  api/
    mix.exs
  frontend/
service2/
  api/
    mix.exs
  frontend/
```

doorgan (approved):
I think this is useful regardless of umbrellas, expert shouldn't care if project names collide

Issue #465, opened 2026-02-26
Expert (VSCode) crashed and will not process new commands -> crash on rc4 install


sb8244:
```
12:00:48.806 [error] GenServer XPEngine.Commands.Reindex terminating
** (stop) exited in: GenServer.call({:global, {XPForge.Document.Store, 1253}}, {:fetch, "file:///Users/stephenbussey/src/supered/super/lib/super/sync_engine/query/sync_data_configuration_store.ex", :analysis}, 5000)
    ** (EXIT) time out
    (elixir 1.15.7) lib/gen_server.ex:1074: GenServer.call/3
    (xp_engine 0.1.0-rc.2) lib/engine/commands/reindex.ex:58: XPEngine.Commands.Reindex.State.entries_for_uri/1
    (xp_engine 0.1.0-rc.2) lib/engine/commands/reindex.ex:27: XPEngine.Commands.Reindex.State.reindex_uri/2
    (xp_engine 0.1.0-rc.2) lib/engine/commands/reindex.ex:126: XPEngine.Commands.Reindex.handle_cast/2
    (stdlib 5.1.1) gen_server.erl:1103: :gen_server.try_handle_cast/3
    (stdlib 5.1.1) gen_server.erl:1165: :gen_server.handle_msg/6
    (stdlib 5.1.1) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
Last message: {:"$gen_cast", {:reindex_uri, "file:///Users/stephenbussey/src/supered/super/lib/super/sync_engine/query/sync_data_configuration_store.ex"}}
State: %XPEngine.Commands.Reindex.State{reindex_fun: #Function<2.99500343/1 in XPEngine.Commands.Reindex.do_reindex>, index_task: nil, pending_updates: %{}}
2026-02-26 12:00:48.965 [info] [name: XPExpert, buffer: XPExpert.Buffer, task_supervisor: :expert_task_queue, dynamic_supervisor: XPExpert.DynamicSupervisor, assigns: XPExpert.Assigns]
```

I reloaded VSCode, expert updated (rc4), and my install failed:

```
2026-02-26 12:03:19.142 [info] [super] Engine build output: /bin/sh: /Users/stephenbussey/Library/Application: No such file or directory
[...]

PR #467, opened 2026-02-26
fix: build engines to `:user_cache` instead of `:user_data`

doorgan:
`:user_data` produces a path like `/Users/dorgan/Library/Application Support/Expert/...`, that path has a space in it which causes the issue in #465:

```
2026-02-26 12:03:19.142 [info] [super] Engine build output: /bin/sh: /Users/stephenbussey/Library/Application: No such file or directory
```

While I can't reproduce it on my macbook, the issue there is that the path is being split on the space, so it tries to execute `/Users/stephenbussey/Library/Application`.

This PR changes the build path to `:user_cache` instead, which is something like `/Users/dorgan/Library/Caches/Expert/...`, which doesn't have a space in it.

The `/bin/sh` call is coming from elixir's `Mix.Tasks.Deps.Compile` on elixir 1.15.7, so not something we can easily patch.
Reconsidering my previous choices, having this be a cache makes more sense than it being user data, as it's really a cache of engines, not really "user data" per se, so it makes sense to change it regardless of the issue.

### DIFF
--- a/apps/expert/priv/build_engine.exs
+++ b/apps/expert/priv/build_engine.exs
@@ -4,15 +4,17 @@
     strict: [
       vsn: :string,
       source_path: :string,
-      force: :boolean
+      force: :boolean,
+      cache_dir: :string
     ]
   )
 
 expert_vsn = Keyword.fetch!(args, :vsn)
 engine_source_path = Keyword.fetch!(args, :source_path)
 force? = Keyword.get(args, :force, false)
+cache_dir = Keyword.fetch!(args, :cache_dir)
 
-expert_data_path = :filename.basedir(:user_cache, "Expert", %{version: expert_vsn})
+expert_data_path = Path.join(cache_dir, expert_vsn)
 
 elixir_erts_vsn = "elixir-#{System.version()}-erts-#{:erlang.system_info(:version)}"
 tooling_path = Path.join([expert_data_path, "tooling", elixir_erts_vsn])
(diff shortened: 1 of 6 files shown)


mhanberg:
Link to fix in Elixir v1.16.1 where this was patched: https://github.com/elixir-lang/elixir/pull/13303

apps/expert/priv/build_engine.exs
```diff
@@ -12,7 +12,7 @@ expert_vsn = Keyword.fetch!(args, :vsn)
 engine_source_path = Keyword.fetch!(args, :source_path)
 force? = Keyword.get(args, :force, false)
 
-expert_data_path = :filename.basedir(:user_data, "Expert", %{version: expert_vsn})
```
mhanberg, line 15: Let's centralize this base path in forge

Also, lets lowercase the E in Expert since we're already changing it. its a nit but it scratches the wrong part of brain when i see it 😅

doorgan: It'd be nice to do it here but this is a standalone script, we'd need to load forge paths to it to make it available. Doable but I'm a bit weary of adding that.

I will centralize it for the rest of the code though!

mhanberg: Pass it in as `--cache-dir` to the script like we do the other stuff

doorgan: I had a brain fart, I was also thinking that we still need to generate it using `:erlang.system_info` which runs on the script, but we can pass the root path from the app anyways

PR #470, opened 2026-02-27
chore(expert): log Expert version on startup

### LINKED ISSUE #439: Log the current version
We should log the current version when the app boots up.

rma97:
Resolves #439

### DIFF
--- a/apps/expert/lib/expert/application.ex
+++ b/apps/expert/lib/expert/application.ex
@@ -86,10 +86,12 @@ defmodule Expert.Application do
     buffer_opts =
       cond do
         opts[:stdio] ->
+          Logger.info("Expert v#{Expert.vsn()} starting on stdio")
           []
 
         is_integer(opts[:port]) ->
           IO.puts("Starting on port #{opts[:port]}")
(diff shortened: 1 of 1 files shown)


apps/expert/lib/expert/application.ex
```diff
@@ -86,10 +86,11 @@ defmodule Expert.Application do
     buffer_opts =
       cond do
         opts[:stdio] ->
+          Logger.info("Expert v#{Expert.vsn()} starting on stdio")
           []
 
         is_integer(opts[:port]) ->
-          IO.puts("Starting on port #{opts[:port]}")
```
mhanberg, line 92: Let's keep this IO.puts in addition to the log.

mhanberg:
> I think this sholdn't shown unconditionally, but:
> - When showing the main "help" message
> - When starting on stdio
> - When starting on tcp mode

If by help message you mean cli --help then we're good, it already logs there

doorgan:
Yep that's what I meant!

apps/expert/lib/expert/application.ex
rmand97: done

Issue #359, opened 2026-02-06
Snowflake dependency for `Forge.Identifier`


katafrakt:
Forge currently uses [Snowflake](https://github.com/blitzstudios/snowflake) for generating identifiers. However, I do have some concerns about this dependency:

1. It hasn't been updated in 8 years
2. It uses deprecated language features, meaning it can simply stop work in the future versions
3. I occasionally get a `{:error, :backwards_clock}` error from it, even though I don't do anything with my clock

From my understanding, the `Forge.Identifier` needs to ba a unique ID that encodes a timestamp. This can be achieved by UUIDv7 too. I created a branch with replacing Snowflake with UUIDv7 using [uniq](https://github.com/bitwalker/uniq) (to avoid namespacing issues). It seems to run fine, tests pass. Here are some performance notes:

1. Building a search index from scratch seems slightly faster with UUIDv7 (5:15 vs 5:30 min, this is within a margin of error though)
2. The checkpoint file is slightly larger with UUIDv7 (because it's twice as large, I think, as the Snowflake id). This was 1.5G vs 1.6G.

## Benchmarks

I also ran some benchmarks of the two.

### ID generation

```
Name                            ips        average  deviation         median         99th %
Uniq.UUID.uuid7(:raw)        2.20 M        0.45 μs  ±1366.47%        0.42 μs        0.54 μs
Snowflake.next_id/0          0.84 M        1.19 μs   ±841.98%        1.04 μs        1.38 μs

Comparison: 
Uniq.UUID.uuid7(:raw)        2.20 M
Snowflake.next_id/0          0.84 M - 2.60x slower +0.73 μs
```

### Comparison (greater than) of two elements

```
Name                            ips        average  deviation         median         99th %
Uniq (binary >)             25.00 M       40.00 ns   ±100.23%          42 ns          42 ns
Snowflake (integer >)       24.90 M       40.16 ns   ±108.12%          42 ns          42 ns

Comparison: 
Uniq (binary >)             25.00 M
Snowflake (integer >)       24.90 M - 1.00x slower +0.164 ns
```

### Sorting a list of 200 elements

(200 was just a random guess, I actually don't know how large are the lists we are comparing in Expert)

```
Name                            ips        average  deviation         median         99th %
Snowflake (Enum.sort)       55.02 K       18.18 μs    ±27.73%       17.79 μs       23.46 μs
[...]

PR #479, opened 2026-03-03
perf: use "Expert ID" instead of Snowflake

### LINKED ISSUE #471: Indexing always fails with backwards_clock error
Around the release of rc.0 the indexing started to regularly fail on my machine with the following error. I am not sure when exactly it started but I don't think this error was there since I started using Expert which was right after the first version of the VS Code extension was released.

```
14:11:16.444 [error] Task #PID<0.14197.0> started from #PID<0.14170.0> terminating
** (MatchError) no match of right hand side value:

    {:error, :backwards_clock}

    (xp_forge 0.1.0-rc.5) lib/forge/identifier.ex:7: XPForge.Identifier.next_global!/0
    (xp_forge 0.1.0-rc.5) lib/forge/search/indexer/entry.ex:85: XPForge.Search.Indexer.Entry.reference/6
    (xp_engine 0.1.0-rc.5) lib/engine/search/indexer/extractors/function_reference.ex:178: XPEngine.Search.Indexer.Extractors.FunctionReference.entry/6
    (xp_engine 0.1.0-rc.5) lib/engine/search/indexer/source/reducer.ex:97: anonymous fn/2 in XPEngine.Search.Indexer.Source.Reducer.apply_extractors/2
    (elixir 1.19.5) lib/enum.ex:2520: Enum."-reduce/3-lists^foldl/2-0-"/3
    (xp_engine 0.1.0-rc.5) lib/engine/search/indexer/quoted.ex:27: anonymous fn/2 in XPEngine.Search.Indexer.Quoted.extract_entries/2
    (elixir 1.19.5) lib/macro.ex:663: anonymous fn/4 in Macro.do_traverse_args/4
    (stdlib 7.2) lists.erl:2673: :lists.mapfoldl_1/3
Function: &:erlang.apply/2
    Args: [#Function<4.128477729/1 in XPEngine.Search.Indexer.do_process_chunks/5>, [[...]]]
```

The files have been redacted but it is no specific file where it fails. Sometimes it fails at around 5% progress, sometimes at 30% and the highest I have seen was around 50%. But there wasn't any clear pattern like time, files in chunk, etc.

I already tried re-installing the extension and deleting the `.expert` folder to no success. I assume #359 would resolve the issue as the error appears to originate from Snowflake.

### Specs
**IDE**: Visual Studio Code (1.109.5)
**Installation method**: VS Code Extension store
**Expert version**: 0.1.0-rc.5
**OS**: Windows with WSL (project and VS code running in WSL)
**OTP / Elixir version**: 28.3.1 / 1.19.5 installed with ASDF

katafrakt:
This replaces Snowflake ID generation with a simplified implementation proposed in https://github.com/elixir-lang/expert/issues/359#issuecomment-3884075978.

These things are hard to measure reliably (note to self: do not ever do that on battery), but here are results from indexing my work project which I think are most reliable:

| Implementation | Time to build | Diff | Index dump size | Diff |
|----|-----|----|----|----|
| Current (Snowflake) | 362.7 s | - | 2350461879 | - |
| UUIDv7 | 313.4 s | -13.6% | 2470686299 | +5.1% |
| Expert ID (this PR) | 306.7 s | -15.4% | 2304117361 | -2% |

This also includes a leftover commit where I added elapsed time to "indexed finished" message. I added it just to measure, but maybe it can stay. It's in a separate commit, so easy to remove if we don't want it.

Fixes #471

### DIFF
--- a/apps/engine/deps.nix
+++ b/apps/engine/deps.nix
@@ -318,23 +318,6 @@ let
         in
         drv;
 
-      snowflake =
-        let
-          version = "1.0.4";
-          drv = buildMix {
-            inherit version;
-            name = "snowflake";
-            appConfigPath = ./config;
-
-            src = fetchHex {
-              inherit version;
-              pkg = "snowflake";
-              sha256 = "badb07ebb089a5cff737738297513db3962760b10fe2b158ae3bebf0b4d5be13";
-            };
-          };
-        in
-        drv;
-
       sourceror =
         let
           version = "1.10.1";
--- a/apps/expert/deps.nix
+++ b/apps/expert/deps.nix
@@ -422,23 +422,6 @@ let
         in
         drv;
 
-      snowflake =
-        let
-          version = "1.0.4";
-          drv = buildMix {
-            inherit version;
-            name = "snowflake";
-            appConfigPath = ./config;
-
-            src = fetchHex {
-              inherit version;
-              pkg = "snowflake";
-              sha256 = "badb07ebb089a5cff737738297513db3962760b10fe2b158ae3bebf0b4d5be13";
-            };
-          };
-        in
-        drv;
-
       sourceror =
         let
           version = "1.10.1";
--- a/apps/expert_credo/deps.nix
+++ b/apps/expert_credo/deps.nix
@@ -241,23 +241,6 @@ let
         in
         drv;
 
-      snowflake =
-        let
-          version = "1.0.4";
-          drv = buildMix {
-            inherit version;
-            name = "snowflake";
-            appConfigPath = ./config;
-
-            src = fetchHex {
-              inherit version;
-              pkg = "snowflake";
-              sha256 = "badb07ebb089a5cff737738297513db3962760b10fe2b158ae3bebf0b4d5be13";
-            };
-          };
-        in
-        drv;
-
       sourceror =
         let
           version = "1.10.1";
--- a/apps/forge/deps.nix
+++ b/apps/forge/deps.nix
@@ -241,23 +241,6 @@ let
         in
         drv;
 
-      snowflake =
-        let
-          version = "1.0.4";
-          drv = buildMix {
-            inherit version;
-            name = "snowflake";
-            appConfigPath = ./config;
-
-            src = fetchHex {
-              inherit version;
-              pkg = "snowflake";
-              sha256 = "badb07ebb089a5cff737738297513db3962760b10fe2b158ae3bebf0b4d5be13";
-            };
-          };
-        in
-        drv;
-
       sourceror =
         let
           version = "1.10.1";


mhanberg:
Make sure to run this when changing deps to ensure that the nix derivation works still. (this is in zsh, might need to change or run each line manually for bash/fish)

```
for proj in "expert" "forge" "engine" "expert_credo"; do
  just mix $proj deps.nix
done
```

katafrakt:
@mhanberg thanks, I did that. Maybe there should be a CI check that verifies that?

On the topic of CI, I think tests are failing here because of some CI misconfig. I'm not sure this works:

```
key: ${{ runner.os }}-mix-${{ matrix.project }}-${{ env.DEFAULT_ELIXIR }}-${{ env.DEFAULT_OTP }}-${{ hashFiles('apps/${{matrix.project}}/mix.lock') }}
```

In the [action log](https://github.com/elixir-lang/expert/actions/runs/22671804808/job/65717611673?pr=479) I see:

```
Cache hit for: Linux-mix-expert_credo-1.15.8-otp-25-25.3.2.4-
Received 6653359 of 6653359 (100.0%), 39.4 MBs/sec
```

This suggests that `hashFiles('apps/${{matrix.project}}/mix.lock')` evaluates to empty string. I think interpolation within interpolation might be illegal (?). As a result it pulls the `_build` with Snowflake present and attempts to start its app, I think.

```
Generated expert_credo app

Notice: 13:39:28.675 [notice] Application gen_lsp exited: :stopped

Notice: 13:39:28.681 [notice] Application schematic exited: :stopped

Notice: 13:39:28.683 [notice] Application telemetry exited: :stopped

Notice: 13:39:28.683 [notice] Application nimble_options exited: :stopped

Notice: 13:39:28.683 [notice] Application jason exited: :stopped

Notice: 13:39:28.683 [notice] Application typed_struct exited: :stopped

Notice: 13:39:28.683 [notice] Application eex exited: :stopped

Notice: 13:39:28.685 [notice] Application sasl exited: :stopped
** (Mix) Could not start application snowflake: could not find application file: snowflake.app
error: Recipe `mix` failed with exit code 1
error: Recipe `test` failed on line 33 with exit code 1
```

The tests pass for me locally.

mhanberg:
Last night I pulled it down and they passed for me as well. 

I'll take a look at what is in the cache.

mhanberg:
You were 100% right about the cache key being wrong.

Should be fixed now, please rebase on main and lets see if she works

katafrakt:
tests pass now 👍  thanks!

mhanberg:
Is there any chance this suffers from a similar problem since it has a time element? Or is it not an issue because system_time pulls the time from the os and that's always monotonic at an OS level?

doorgan:
@mhanberg this doesn't have the time issue because it doesn't compare clocks at all(as in, there won't be an error), it's a naive implementation partly because of that(and faster due to that too). Comparing clocks it could reveal a time shift issue if for example you use devcontainers(this is where I recall seeing this issue in lexical back when I mainly used windows and wsl), but I'm not sure about how much it matters

mhanberg:
Sweet, I guess I meant could it be subtly wrong, rather than alert us. 

Sounds good tho.

Issue #484, opened 2026-03-04
Mix is running interactively


Draggu:
Mix is running interactively and prints to stdout, which leads to a permanent communication failure. I looked through calls, but there are already `--force --if-missing` flags. I guess it is coming from a different command calling `mix local.hex` under the hood

```
2026-03-02 14:22:42.640 [info] [Error - 2:22:42 PM] Client Expert: connection to server is erroring.
Header must provide a Content-Length property.
{"mix requires the hex package manager to fetch dependencies\nshall i install hex? (if running non-interactively, use \"mix local.hex --force\") [yn] content-length":"172"}
Shutting down server.
```

PR #485, opened 2026-03-04
feat: halt system on significant child process failure

### LINKED ISSUE #428: Correctly shutdown in case we shut down the buffer
> I believe this issue is from the genlsp process crashing and then being restarted by the supervisor and then the buffer restarting, which then starts to consume a stdio stream in the middle and you just can't recover from that.
> 
> We should probably think of the proper restart strategy for those processes, but its most likely that the app should just terminate if the expert process crashes. 

 _Originally posted by @mhanberg in [#366](https://github.com/elixir-lang/expert/issues/366#issuecomment-3892877619)_

Draggu:
`gen_lsp` part https://github.com/elixir-tools/gen_lsp/pull/80

Close #428, I tried to fix the source problem here (writing to stdout), but I'm not sure if [case](#484
) I found is the only one. I don't think there exists a valid restart option since clients can't recover from gibberish written to output.


apps/engine/mix.exs
```diff
@@ -50,7 +50,9 @@ defmodule Engine.MixProject do
       {:elixir_sense,
        github: "elixir-lsp/elixir_sense", ref: "e3ddc403554050221a2fd19a10a896fa7525bc02"},
       {:forge, path: "../forge"},
-      {:gen_lsp, "~> 0.11.3"},
+      {:gen_lsp,
+       git: "https://github.com/elixir-tools/gen_lsp",
```
mhanberg: Tip: you can use the github option and branch option together, a little more ergonomic and easier to update when you push changes.

apps/expert/lib/expert/application.ex
```diff
@@ -115,9 +113,23 @@ defmodule Expert.Application do
...
+
+    case Supervisor.start_link(children_spec, opts) do
+      {:ok, pid} ->
+        spawn(fn ->
+          ref = Process.monitor(pid)
+
+          receive do
+            {:DOWN, ^ref, :process, _, _} ->
+              Logger.error("Expert application has terminated unexpectedly")
+              # Signifficant process we cant recover (like the GenLSP) has crashed
```
mkaput: typos


```suggestion
              # A significant process we can't recover from (like the GenLSP) has crashed.
```

mhanberg:
testing this today

apps/expert/lib/expert/application.ex
```diff
@@ -117,9 +117,23 @@ defmodule Expert.Application do
...
+              :timer.apply_after(50, System, :halt, [1])
+          end
+        end)
```
mhanberg: I don't think this is the right approach. When i was playing aroudn with this locally, i was never seeing the log hit the system, probably because it was not getting handled before we halted the system.

I think the better option is to implement the c:stop callback on the application and halt the system from there when the application shuts down (which it will when we shutdown the supervisor, which shuts down from the significant process dying).

there is one issue tho, from the perspective of that callback, we can't really know if the application was terminated normally or due to a crash. This same callback is executed when the client sends the exit notification, since we call System.stop there, which gracefully shuts down the system. I think perhaps what we want to do is set a piece of global state (persistent_term maybe) if we exit on purpose, and check that in the callback. if its present, then we can just let it continue to shut down, else we can halt with exit code 1.

This method uses existing otp conventions more or less, whereas the approach in this pr i have never seen before (which is not to say its incorrect, it just is possibly non-idiomatic).

It would ideally be nice if we could list an application as significant, and if it shutdown, the whole beam should terminate. I had sort of assumed that was possible, but doesn't seem like it.

apps/expert/mix.exs
```diff
@@ -86,7 +86,7 @@ defmodule Expert.MixProject do
       # assume a roundtrip to a project node is made.
       {:engine, path: "../engine", only: [:test]},
       {:forge, path: "../forge"},
-      {:gen_lsp, "~> 0.11.3"},
+      {:gen_lsp, github: "Draggu/gen_lsp", branch: "restart"},
```
mhanberg: instead of making changes to gen_lsp, lets just override the child spec in the application child list. it might be better in the long run, but we can make that change if this seems to work

mhanberg (commented):
In testing this, I may have found a bug in the neovim lsp client. when you close it, it seems to send the shutdown request multiple times (which we handle gracefully, but was pretty confusing when encountering).

apps/expert/lib/expert/application.ex
Draggu: Done. Thanks for the tip

PR #486, opened 2026-03-04
ci: run through zizmor and fix broken caching

mhanberg:
Zizmor suggested

- add dependabot cool down
- avoid command injection by putting dynamic bits in env vars
- reducing permissions blast radius
- pinning the hashes (apparently if you pin them with the version
  comment dependabot actually will honor that)
- setting persist-credentials to false on checkout action

This also fixes the cache problem that @katafrakt was seeing in #479.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ env:
 permissions:
   contents: read
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   static-analysis:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -13,6 +13,10 @@ env:
 permissions:
   contents: read
 
+defaults:
+  run:
+    shell: bash
```
mhanberg, line 18: apparently this sets it to git bash on windows, which allows us to use the same shell syntax everywhere

PR #490, opened 2026-03-05
fix: fetch `hex` and `rebar` in all places

### LINKED ISSUE #484: Mix is running interactively

Draggu:
Close #484

I'm not sure which call it came from, but simply fetching `hex` and `rebar`, as was done before, with a few calls already in all the necessary places, should do the trick.

### DIFF
--- a/apps/engine/lib/engine/build/project.ex
+++ b/apps/engine/lib/engine/build/project.ex
@@ -3,6 +3,7 @@ defmodule Engine.Build.Project do
   alias Engine.Build.Isolation
   alias Engine.Plugin
   alias Engine.Progress
+  alias Forge.Internet
   alias Forge.Project
   alias Mix.Task.Compiler.Diagnostic
 
@@ -53,6 +54,7 @@ defmodule Engine.Build.Project do
       Mix.Task.clear()
       Progress.report(token, message: "Compiling #{Project.display_name(project)}")
       result = compile_in_isolation()
+      Project.ensure_hex_and_rebar()
       Mix.Task.run(:loadpaths)
       result
     end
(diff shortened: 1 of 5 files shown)


apps/engine/lib/engine/build/project.ex
```diff
@@ -53,6 +53,8 @@ defmodule Engine.Build.Project do
       Mix.Task.clear()
       Progress.report(token, message: "Compiling #{Project.display_name(project)}")
       result = compile_in_isolation()
+      Mix.Task.run("local.hex", ~w(--force --if-missing))
+      Mix.Task.run("local.rebar", ~w(--force --if-missing))
```
mkaput: worth extracting into named function isn't it?

doorgan:
It'd be nice to know where exactly this problem happens but iirc you mentioned this issue is not reproducible reliably.
Should we check if the user may not be connected to internet before this? Otherwise it looks like Expert wouldn't work offline(not that we have *great* support for offline but still)

apps/engine/lib/engine/build/project.ex
Draggu: done

doorgan:

Issue #493, opened 2026-03-06
Code intelligence does not work on a project


katafrakt:
I built Expert from `main` and opened a previously opened project. Suddenly, whatever I was trying to do (hover, go to definition) resulted in an error:

```

17:08:21.949 instance_id=19CC3D1BC35 [error] ** (MatchError) no match of right hand side value:

    {:error,
     {:not_enabled,
      {:resolve_mfa, JumpWeb.BlahController, :blah_blah, 2}}}

    (xp_expert 0.1.0-rc.5) lib/expert/provider/handlers/hover.ex:87: XPExpert.Provider.Handlers.Hover.hover_content/2
    (xp_expert 0.1.0-rc.5) lib/expert/provider/handlers/hover.ex:28: XPExpert.Provider.Handlers.Hover.handle/1
    (xp_expert 0.1.0-rc.5) lib/expert.ex:119: XPExpert.handle_request/2
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:390: anonymous fn/2 in XPGenLSP.loop/3
    (xp_telemetry 1.3.0) /Users/pawel.sw/dev/expert/apps/expert/deps/telemetry/src/telemetry.erl:324: :xp_telemetry.span/3
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:389: anonymous fn/5 in XPGenLSP.loop/3
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:559: anonymous fn/4 in XPGenLSP.attempt/4
    (elixir 1.19.4) lib/task/supervised.ex:105: Task.Supervised.invoke_mfa/2
```

This was preceeded by various errors related to ETS.

```
17:02:21.456 [error] GenServer XPEngine.Search.Store terminating
** (stop) exited in: GenServer.call({:global, {:ets_search, "api"}}, {:find_by_subject, [:_, :struct, :definition]}, 5000)
    ** (EXIT) an exception was raised:
        ** (ArgumentError) errors were found at the given arguments:

  * 2nd argument: not a key that exists in the table

            (stdlib 7.2) :ets.lookup_element(:expert_search_v3, {:by_id, 4486756823203841, :struct, :definition}, 2)
            (xp_engine 0.1.0-rc.5) lib/engine/search/store/backends/ets/state.ex:99: anonymous fn/2 in XPEngine.Search.Store.Backends.Ets.State.find_by_subject/4
[...]

PR #494, opened 2026-03-09
fix(engine): add missing rescue for `:ets.lookup_element`

katafrakt:
All other cases where `:ets.lookup_element` is used have rescue on `ArgumentError` to provide fallback. We need it here as well.

This addressed #493, although probably the full recovery step here involves dropping the search index and letting Expert to create it anew. I wonder if we should add it to release notes somehow. The other option is to add new schema version (although schema was not changed, just the mechanism to build IDs).

### DIFF
--- a/apps/engine/lib/engine/search/store/backends/ets/state.ex
+++ b/apps/engine/lib/engine/search/store/backends/ets/state.ex
@@ -96,10 +96,9 @@ defmodule Engine.Search.Store.Backends.Ets.State do
       id_keys
     end)
     |> MapSet.new()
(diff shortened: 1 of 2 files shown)


katafrakt:
I wasn't sure TBH. I think since this is because of ID generation change, it will fails for all anyway. But maybe it is a good idea to change that (and wrap `lookup_element` in geenral.

doorgan:
Yeah wrapping lookup_element and make callers deal with the result value sounds better in general

doorgan:

katafrakt:
Done wrapping. Added some tests too.

PR #495, opened 2026-03-09
fix(expert): display user-friendly message on filesystem error

### LINKED ISSUE #476: badrpc message during find definition
I use `expert` (version is `v0.1.0-rc.5`, taken from [here](https://github.com/elixir-lang/expert/releases/download/v0.1.0-rc.5/expert_linux_amd64)) with `GNU Emacs 30.2`, `eglot` and `eglot-booster` under Guix System (the NixOS equivalent with Guile Scheme instead of Nix DSL).

Sometimes I get the following message:

```
[eglot] Server reports (type=1): [lib] Failed to start child {XPExpert.Project.Node, "lib"}: {:badrpc, {:EXIT, {{:badmatch, {:error, :erofs}}, [{XPForge.Project, :ensure_workspace_directory, 1, [file: ~c"lib/forge/project.ex", line: 241]}, {XPForge.Project, :ensure_workspace, 1, [file: ~c"lib/forge/project.ex", line: 224]}, {XPEngine.Bootstrap, :init, 5, []}]}}}
```

This can be reproduced as follows. Place the text editor cursor on something like `Application.spec(:myapp, :vsn)`, then have the LSP server perform a `textDocument/definition` (in `Emacs`: `M-x xref-find-definitions`), i.e., switch current buffer to the implementation code, in this case `Application.spec/2` in the `application.ex` file, which belongs to Elixir. I guess the error may be related to the fact that the file is immutable due to Guix System specifics. In my case it is `/gnu/store/1i5wvfhr3k39aldksjv42rl4w2kpa76d-elixir-1.19.5-src/source/lib/elixir/lib/application.ex`, and everything at `/gnu/...` is read-only for a common user.

katafrakt:
This handles situations such as read-only directories or read-only filesystems (NixOS). Currently when attempting to create an `.expert` directory, it fails with an error that is difficult to understand.

Fixes #476 although I only simulated the NixOS situation with a directory without write permissions.

Screenshot 2026-03-09 at 19 08 07

### DIFF
--- a/apps/expert/lib/expert/engine_node.ex
+++ b/apps/expert/lib/expert/engine_node.ex
@@ -249,7 +249,7 @@ defmodule Expert.EngineNode do
 
   defp start_net_kernel(%Project{} = project) do
     manager = Project.manager_node_name(project)
-    Node.start(manager, name_domain: :longnames)
+    Node.start(manager, :longnames)
   end
 
   defp ensure_apps_started(node, token) do


apps/expert/lib/expert/project/node.ex
```diff
@@ -112,4 +119,13 @@ defmodule Expert.Project.Node do
       error -> error
     end
   end
+
+  defp bootstrap_error_message(reason) do
+    case reason do
+      :eacces -> "Project directory has insufficient permissions"
+      :erofs -> "Project is in a read-only filesystem"
+      :enospc -> "No disk space available"
+      _ -> "Unable to create workspace: #{inspect(reason)}"
```
doorgan: Shouldn't this be something like `Unable to bootstrap engine: `? A workspace is a different concept

apps/expert/lib/expert.ex
```diff
@@ -482,7 +482,7 @@ defmodule Expert do
 
   defp initialization_error_message({:shutdown, {:failed_to_start_child, child, reason}}) do
     case child do
-      {Project.Node, node_name} ->
+      {Expert.Project.Node, node_name} ->
```
doorgan, line 484: Is this a leftover?

apps/expert/lib/expert/project/node.ex
doorgan: We should also probably tell the user what permissions we expect

apps/expert/lib/expert.ex
katafrakt: Actually, this module aliases both `Expert.Project` and `Forge.Project`, and I thought it is a bug - it should match `Expert.Project.Node`, because it's called with results from `Expert.Project.Supervisor.ensure_node_started(project)`. But maybe it should not be here in this pull request.

doorgan: Ahh got it, no we can keep it

apps/expert/lib/expert/project/node.ex
katafrakt: Good catch, updated the messages.

apps/expert/lib/expert.ex
katafrakt: I guess we can remove the shadowed alias too.

apps/expert/lib/expert/engine_node.ex
```diff
@@ -242,7 +249,7 @@ defmodule Expert.EngineNode do
 
   defp start_net_kernel(%Project{} = project) do
     manager = Project.manager_node_name(project)
-    Node.start(manager, :longnames)
+    Node.start(manager, name_domain: :longnames)
```
katafrakt: not sure why Dialyzer started to complain about this now, but it seems like a legit complaint

doorgan:

apps/expert/lib/expert/engine_node.ex
katafrakt: nevermind, I didn't realize it does not work on Elixir 1.16, which Expert supports

PR #499, opened 2026-03-12
fix(forge): id generator must not create duplicate IDs

### LINKED ISSUE #498: Go To Definition returns spurious results
Using go-to-definition in a large-ish project shows multiple references that are nonsensical. They don't contain the function name in question or they're either too small or out of bounds for the file.

This is because `Forge.Identifier` uses the process dictionary to store state, and when the indexer runs, it uses Task.async_stream, which creates multiple processes. This means that entries in the index can have the same ID, which is the reason for the spurious entries. 

PR coming soon

scohen:
Recently, the id generator was replaced with a different implementation that was faster. However, that implementation relied on storing state in a process dictionary. When expert indexes source code, it uses Task.async_stream, which will cause the id generator to run inside of multiple processes, which can make it generate duplicate IDs.

When this happens, features that use the search index will return spurious results.

This change moves the id generator to use erlang's :atomics module to store the relevant state. It also changes the format of the IDs. Since we're only ever going to run on a single node, the machine number in the snowflake format was pretty useless. The new ID generator reclaims those ten bits, and uses them for the timestamp. This allows timestamps represent microseconds. This means that the sequence number is only used if two ids are generated in the same microsecond.

Also added tests

Fixes #498

### DIFF
--- a/apps/forge/lib/forge/identifier.ex
+++ b/apps/forge/lib/forge/identifier.ex
@@ -1,5 +1,5 @@
 defmodule Forge.Identifier do
-  # Microseconds since Jan 1, 2024 UTC
+  # Microseconds from the Unix epoch to January 1, 2024, UTC
   @epoch 1_704_070_800_000_000
   @atomics_key {__MODULE__, :atomics_key}
   @max_seq 4096


scohen:
Note: I also added a benchmark for this new implementation, and it does around 12MM ops/sec on my M4 max Macbook Pro.

apps/forge/lib/forge/identifier.ex
```diff
@@ -1,35 +1,55 @@
 defmodule Forge.Identifier do
-  import Bitwise
+  # Microseconds since Jan 1, 2024 UTC
```
rmand97:
```suggestion
  # Microseconds from the Unix epoch to January 1, 2024, UTC
```

PR #505, opened 2026-03-12
fix: prevent recursive alias generation for nested modules

### LINKED ISSUE #497: Fetching function docs crashes
I'm on rc.5 and if I have something like `Map.fetch(...)` and I'm on the `Map` and press `K` in neovim, I get the docs for `Map` (the "hover" functionality). But, if I instead am on `.fetch` and do the same, I just get `No information available`. It seems to be the same for all functions.

If I check my lsp log I can see errors like this

```
[ERROR][2026-03-10 12:16:12] .../vim/lsp/buf.lua:66	-32603	"** (MatchError) no match of right hand side value: {:error, {:not_enabled, {:resolve_mfa, Map, :fetch, 2}}}\n    (xp_expert 0.1.0-rc.5) lib/expert/provider/handlers/hover.ex:87: XPExpert.Provider.Handlers.Hover.hover_content/2\n    (xp_expert 0.1.0-rc.5) lib/expert/provider/handlers/hover.ex:28: XPExpert.Provider.Handlers.Hover.handle/1\n    (xp_expert 0.1.0-rc.5) lib/expert.ex:119: XPExpert.handle_request/2\n    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:390: anonymous fn/2 in XPGenLSP.loop/3\n    (xp_telemetry 1.3.0) /Users/runner/work/expert/expert/apps/expert/deps/telemetry/src/telemetry.erl:324: :xp_telemetry.span/3\n    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:389: anonymous fn/5 in XPGenLSP.loop/3\n    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:559: anonymous fn/4 in XPGenLSP.attempt/4\n    (elixir 1.17.3) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2\n"
```

LSP settings for Expert

```
vim.lsp: Enabled Configurations ~
- expert:
  - capabilities: {
      textDocument = {
        completion = {
          completionItem = {
            commitCharactersSupport = false,
            deprecatedSupport = true,
            documentationFormat = { "markdown", "plaintext" },
            insertReplaceSupport = true,
            insertTextModeSupport = {
              valueSet = { 1 }
            },
            labelDetailsSupport = true,
            preselectSupport = false,
            resolveSupport = {
              properties = { "documentation", "detail", "additionalTextEdits", "command", "data" }
            },
            snippetSupport = true,
            tagSupport = {
              valueSet = { 1 }
            }
          },
          completionList = {
            itemDefaults = { "commitCharacters", "editRange", "insertTextFormat", "insertTextMode", "data" }
          },
          contextSupport = true,
          insertTextMode = 1
        }
      }
    }
  - cmd: { "/Users/myuser/.local/bin/expert", "--stdio" }
  - filetypes: elixir, eelixir, heex, surface
  - root_markers: { "mix.exs", ".git" }
```

Once in a blue moon it _does_ work though which is a bit odd.

Anything I'm doing wrong or it's a bug?

Draggu:
Close #497

We used only first segment of nested modules to create alias, which was failing if nested module name started with segment that was already an alias

### DIFF
--- a/apps/forge/lib/forge/ast/analysis.ex
+++ b/apps/forge/lib/forge/ast/analysis.ex
@@ -442,21 +442,33 @@ defmodule Forge.Ast.Analysis do
     state
   end
 
-  defp maybe_push_implicit_alias(%State{} = state, [first_segment | _], document, quoted)
+  defp maybe_push_implicit_alias(
+         %State{} = state,
+         [first_segment | _] = module_segments,
+         document,
+         quoted
+       )
        when is_atom(first_segment) do
     segments =
       case State.current_module(state) do
         # the head element of top-level modules can be aliased, so we
         # must expand them
         [] ->
-          expand_alias([first_segment], state)
+          expand_alias(module_segments, state)
 
-        # if we have a current module, we prefix the first segment with it
+        # if we have a current module, we prefix the segments with it
         current_module ->
-          current_module ++ [first_segment]
+          current_module ++ module_segments
       end
 
-    implicit_alias = Alias.implicit(document, quoted, segments, first_segment)
+    implicit_alias =
+      Alias.implicit(
+        document,
+        quoted,
+        segments,
+        module_segments
+      )
+
     State.push_alias(state, implicit_alias)
   end
 
@@ -484,7 +496,7 @@ defmodule Forge.Ast.Analysis do
     alias_map = state |> State.current_scope() |> Scope.alias_map()
 
     case alias_map do
-      %{^first => existing_alias} ->
+      %{[^first] => existing_alias} ->
         existing_alias.module ++ rest
 
       _ ->
--- a/apps/engine/lib/engine/analyzer.ex
+++ b/apps/engine/lib/engine/analyzer.ex
@@ -49,7 +49,7 @@ defmodule Engine.Analyzer do
 
     if is_nil(maybe_imported_mfa) do
       aliases = aliases_at(analysis, position)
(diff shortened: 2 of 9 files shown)


apps/forge/lib/forge/ast/analysis.ex
```diff
@@ -442,21 +442,33 @@ defmodule Forge.Ast.Analysis do
...
+          current_module ++ module_segments
       end
 
-    implicit_alias = Alias.implicit(document, quoted, segments, first_segment)
+    implicit_alias =
+      Alias.implicit(
+        document,
+        quoted,
+        segments,
+        module_segments |> Enum.join(".") |> String.to_atom()
```
mkaput: String.to_atom() is dangerous. Allocated atoms are never deallocated. Is this necessary here? (I don't know the context)

mhanberg: For modules you want to use Module.concat

Draggu: > String.to_atom() is dangerous.

I thought about using list instead (no need to create new atoms), but it would require me to rework other parts too, would be better to do in follow-up

> For modules you want to use Module.concat

It will prefix these with `Elixir.` which I want to explicitly omit here

Draggu:
Reworked a bit to not construct more atoms than necessary, kept this logic in tests for convenience.

PR #509, opened 2026-03-13
fix: build a single engine for multiple projects using the same toolchain versions

### LINKED ISSUE #466: Build the engine once for every elixir+otp combination
Currently if you have a workspace with multiple workspace folders and all of them use the same elixir+otp versions, we will try to build the engine for each one of them.
In other words, if your workspace has 3 folders, 3 engines will be built

To reduce work and speed up statup times, we should only try to build the engine once for each elixir+otp pair during the same session, and cache the engine path for any subsequent engine restart.
If we try to build the same engine twice at the same time, only a single build should run and both engines should use the results of that single run.

doorgan:
Fix #466 

When resuming a session on Expert itself, I would regularly see issues due to multiple engine builders trying to modify the same files. This PR introduces a coordinator that keeps track of finished and in-flight builds, and ensures that:

- An engine for an elixir+otp pair is built once during the lifetime of the Expert instance, and reused upon project restarts
- If multiple projects using the same versions request an engine at the same time, or while that version is being built, they all wait for the same, single builder to finish instead of starting new ones
- Additionally, ensures engines for different versions can be built concurrently

Besides fixing that issue, it also cuts down startup time for Expert by a lot, since we builds the engine once instead of thrice(forge, engine, expert), once one engine finishes building the rest are instant.


mhanberg:
@katafrakt can you give this a review when you get a chance.

apps/expert/lib/expert/engine_builds.ex
```diff
@@ -0,0 +1,179 @@
...
+          {:ok, pending} ->
+            pending = %{pending | waiters: [from | pending.waiters]}
+            pending_map = Map.put(state.pending, key, pending)
+
+            {:noreply, %State{state | pending: pending_map}}
+
+          :error ->
+            case start_builder(project, build, key) do
+              {:ok, pid} ->
+                pending = %{pid: pid, ref: Process.monitor(pid), waiters: [from]}
```
katafrakt, line 55: I would probably personally use a struct for this pending, but no big deal

apps/expert/lib/expert/engine_builds.ex
```diff
@@ -0,0 +1,179 @@
...
+    case toolchain_versions(project, elixir, env) do
+      {output, 0} ->
+        output = String.trim(output)
+
+        with {:ok, binary} <- Base.decode64(output),
+             {elixir_version, erts_version} <- :erlang.binary_to_term(binary) do
+          {:ok, {elixir_version, erts_version}}
+        else
+          _ -> {:error, "Failed to determine Elixir/OTP runtime for project"}
+        end
```
katafrakt, line 126: I feel like this should still be a part of `toolchain_versions`, which should return result tuple, not a raw base64-encoded command result. WDYT?

apps/expert/lib/expert/engine_builds.ex
```diff
@@ -0,0 +1,179 @@
...
+        {:error,
+         "Failed to determine Elixir/OTP runtime for project: #{String.trim(output)} (status #{status})"}
+    end
+  end
+
+  defp toolchain_versions(%Project{} = project, elixir, env) do
+    cmd =
+      "{System.version(), to_string(:erlang.system_info(:version))} |> :erlang.term_to_binary() |> Base.encode64() |> IO.write()"
+
+    System.cmd(to_string(elixir), ["--eval", cmd],
```
katafrakt, line 138: I wonder if this would cause the problems again on Windows, as fixed in #518. I'll try to verify that.

katafrakt (commented):
The change looks great. I just left 2 code style comments + a doubt about Windows.

apps/expert/lib/expert/engine_builds.ex
katafrakt: nevermind, it's fine as we don't use `\c` switch.

mhanberg:
This will definitely be an improvement.

I am wondering if it can be even further improved by yoinking some of the "build lock" code that they have put into mix for the compiler lock that could potentially be used to lock them between lsp instances (obviously without actually clustering the LSPs. just using the socket implementation they have)

katafrakt (approved):

PR #519, opened 2026-03-18
perf(engine): load module store after build

scohen:
After expert starts, a user's completion request trigger's elixir_sense's module store to load, which on my project takes 12 seconds. This makes expert feel sluggish.

This change automatically loads the module store after the project's build completes and provides progress while this is happening. After the module store is built, completions are fast.


lukaszsamson:
Did you consider dropping the module store? AFAIK only ash plugin was using it

scohen:
@lukaszsamson I'm not sure what you mean, the module store is created internally by elixir_sense

Issue #504, opened 2026-03-12
`.expert/.gitignore` not created


Draggu:
Sometimes `.expert/.gitignore` is not created

mhanberg:
I think this is because the logger creates the directory and file but the gitignore isn't added to later, so if you use like the --version command it doesn't get created.

Draggu:
It was reported to me on a clear install (vsc extension) with just opening a project. Unfortunately, I don't have logs from this run

katafrakt:
I think I've seen it yesterday. I opened one of my old projects, on which Expert was never run. Quite quickly, `.expert` directory appeared in the list of unstaged files, but also it disappeared after a while, because `.gitignore` was added. Perhaps we could change the order to add `.gitignore` immediately after create a directory, but I don't think it is a persistent issue.

katafrakt:
I believe this is fixed with the linked PR

PR #520, opened 2026-03-18
fix: issues with creating `.expert` directory too eagerly

katafrakt:
This addresses things from #504. As @mhanberg said, currently running `expert --version` was indeed creating `.expert` dir with empty `expert.log`. This was because of the setup in `runtime.exs`, which I removed. Instead, we explicitly attach the logger handler in new `ProjectLogFile` module when expert is started with `--stdio` or `--port`.

The log lever in `runtime.exs` is set to `:none` to avoid polluting stdout.

It also moves `Project.ensure_workspace(project)` way up in the `Bootstrap.init`. This is the function that creates `.gitignore` and we want to ensure this is almost always created, regardless of a potential failure of subsequent operations in bootstrap.

### DIFF
--- a/apps/expert/test/expert/logging/project_log_file_test.exs
+++ b/apps/expert/test/expert/logging/project_log_file_test.exs
@@ -5,14 +5,10 @@ defmodule Expert.Logging.ProjectLogFileTest do
 
   require Logger
 
-  @handler_name :expert_project_log
-
   setup do
     on_exit(fn ->
-      case :logger.remove_handler(@handler_name) do
-        :ok -> :ok
-        {:error, {:not_found, @handler_name}} -> :ok
-      end
+      ProjectLogFile.handler_name()
+      |> :logger.remove_handler()
     end)
   end
 
--- a/apps/expert/lib/expert/logging/project_log_file.ex
+++ b/apps/expert/lib/expert/logging/project_log_file.ex
@@ -1,9 +1,12 @@
 defmodule Expert.Logging.ProjectLogFile do
   @moduledoc false
 
-  @handler_name :expert_project_log
   @max_no_bytes 10_485_760
   @max_no_files 3
+
+  def handler_name do
+    :expert_project_log
+  end
 
   def attach(root_path \\ File.cwd!()) when is_binary(root_path) do
     with :ok <- ensure_workspace(root_path) do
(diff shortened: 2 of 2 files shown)


apps/expert/test/expert/logging/project_log_file_test.exs
```diff
@@ -0,0 +1,41 @@
...
+
+  require Logger
+
+  @handler_name :expert_project_log
+
+  setup do
+    temp_root =
+      System.tmp_dir!()
+      |> Path.join("expert-log-test-#{System.unique_integer([:positive])}")
+      |> tap(&File.mkdir_p!/1)
```
mhanberg: you can create a tmp dir with the `@tag :tmp_dir` tag, which puts a `:tmp_dir` field in the context.

apps/expert/test/expert/logging/project_log_file_test.exs
```diff
@@ -0,0 +1,41 @@
+defmodule Expert.Logging.ProjectLogFileTest do
+  use ExUnit.Case, async: false
+
+  alias Expert.Logging.ProjectLogFile
+
+  require Logger
+
+  @handler_name :expert_project_log
```
mhanberg: i'd make the handler name accessible in a public function in the module and then use it here.

apps/forge/test/forge/project_test.exs
```diff
@@ -245,4 +245,26 @@ defmodule Forge.ProjectTest do
       assert Atom.to_string(node_name) =~ sanitized_name
     end
   end
+
+  describe "ensure_workspace/1" do
+    test "creates .gitignore when the workspace directory already exists" do
+      temp_root =
+        System.tmp_dir!()
+        |> Path.join("expert-project-test-#{System.unique_integer([:positive])}")
+        |> tap(&File.mkdir_p!/1)
```
mhanberg: see previous comment about tmp_dir

apps/expert/test/expert/logging/project_log_file_test.exs
katafrakt: Right, great point. Changed.

apps/expert/test/expert/logging/project_log_file_test.exs
katafrakt: Done

Issue #340, opened 2026-01-27
Project engine does not initialize when project Elixir version is newer than Expert build


nicflower:
## Environment
- Editor: Neovim
- Expert version: master (built from source)
- Expert build Elixir: Elixir 1.16.3 (OTP 25)
- Project Elixir (tested):
    - ❌ Elixir 1.18.4 + OTP 28 (fails)
    - ❌ Elixir 1.18.x + OTP 27 (fails)
    - ✅ Elixir 1.16.x + OTP 25 (works correctly)
- elixir installation managed via asdf
- OS: Linux

## Description

`project.log` is never created when using Expert with newer Elixir versions (engine appears to start but no project logs), attempting to call an Lsp function causes this log message (from `:LspLog`):
```
[WARN][2026-01-27 09:24:22] /usr/local/share/nvim/runtime/lua/vim/lsp/log.lua:151	"Received request textDocument/hover before engine was initialized. Ignoring."
```
In the root of the project i correctly see an `.expert` folder containing the `expert.log` file, but all the other files are missing.

Attached you can find the `expert.log` file. Here I don't see particular errors besides these:
```
09:10:18.327 [error] ** (XPGenLSP.InvalidResponse) Invalid response for request shutdown.

Response: {:ok, nil}
Errors: "expected either null or a %XPGenLSP.ErrorResponse{}"

```

but digging around issues does not seem like a blocking error

[expert.log](https://github.com/user-attachments/files/24878289/expert.log)

mhanberg:
How long did you wait before determining it was not working? It does need to compile the "engine" on first launch for a specific elixir/erlang pair.

Also, we don't test on OTP 25, so there is potentially it had an issue with clustering to an engine node (the server is a node, and starts an "engine" node for your project) that is OTP 26 or higher.

Could you try compiling expert with elixir = "1.17.3" erlang = "27.3.4.1" ? Those are the versions that we build the nightly release with.

spicychickensauce:
I have the same problem with the latest 0.1.0-rc.2 in Zed and in VSCode. It just stays forever stuck at "Starting project node".
There are no useful logs, just multiple `Received request textDocument/codeAction before engine for app was initialized. Ignoring.`

Any ideas on how we could help debug this?

katafrakt:
@spicychickensauce when you say "no useful logs" do you mean `expert.log` files or "Server Logs" inside Zed?

[...]

PR #527, opened 2026-03-21
fix: hang at startup if .bashrc contains exec call

spicychickensauce:
See: https://github.com/elixir-lang/expert/issues/340

In my .bashrc I've got this at the end:

```
# If not running interactively, don't do anything
[[ $- != *i* ]] && return
exec fish
```
This is one of the recommended ways on how to set up fish on the arch wiki.
But, with this .bashrc, expert hangs indefinitely at startup.

The problem with the original code is the use of the `-i` flag.
I don't know the original intent behind using that flag here, so instead of simply removing it, I added a timeout + fallback without -i instead.

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -263,26 +263,7 @@ defmodule Expert.Port do
       {"PATH", System.get_env("PATH", @default_unix_path)}
     ]
 
-    shell_name = Path.basename(shell)
-
-    args =
-      case shell_name do
-        "fish" ->
-          cmd =
-            "cd #{directory}; printf \"#{@path_marker}:%s:#{@path_marker}\" (string join ':' $PATH)"
-
-          ["-l", "-c", cmd]
-
-        "nu" ->
-          cmd =
-            "cd #{directory}; print $\"#{@path_marker}:($env.PATH | str join \":\"):#{@path_marker}\""
-
-          ["-l", "-c", cmd]
-
-        _ ->
-          cmd = "cd #{directory} && printf \"#{@path_marker}:%s:#{@path_marker}\" \"$PATH\""
-          ["-i", "-l", "-c", cmd]
-      end
+    args = path_fetch_cmd_args(shell, directory)
 
     maybe_cmd_output =
       case cmd_with_timeout(shell, args, env: env, timeout: 1_000) do
@@ -305,10 +286,10 @@ defmodule Expert.Port do
       {:ok, {output, exit_code}} ->
         case Regex.run(~r/#{@path_marker}:(.*?):#{@path_marker}/s, output) do
           [_, clean_path] when exit_code == 0 ->
-            clean_path
+            {:ok, clean_path}
 
           _ ->
-            output |> String.trim() |> String.split("\n") |> List.last()
+            {:ok, output |> String.trim() |> String.split("\n") |> List.last()}
         end
 
       {:error, :timeout} ->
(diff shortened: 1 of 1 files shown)


spicychickensauce:
I fixed the credo warning by extracting the args construction into its own function, should be good now

katafrakt:
I think interactive flag might be necessary to support mise in "path mode", which relies on the prompt display to activate.

I also wonder if this is the best solution to the problem. Maybe instead there should be a flag passed to expert command, such as `expert --stdio --force-non-interactive`.

spicychickensauce:
I'm also using mise, and it works for me only without the -i flag...

I don't like if users need to configure the right behavior, then it will just not work out of the box for some and the setting will be hard to track down.

Also, a timeout is needed anyway, as without it the whole lsp just hangs, with no logs to even help identify the problem.

katafrakt:
> a timeout is needed anyway, as without it the whole lsp just hangs

I agree with that. But what bothers me is what happens after - removal of `-i` switch is very arbitrary. What if someone has a condition in `.bashrc` on `-l` flag for some reason?

> I'm also using mise, and it works for me only without the -i flag...

This is interesting. Maybe I misunderstood how the "path mode" works. How do you set up mise (for bash or for fish)? Can you paste first 4 lines of `mise doctor` output?
It seems from the problem description that your `SHELL` env variable is set to `bash`, not `fish`, when Expert starts. This is a real problem issue IMO.

> I don't like if users need to configure the right behavior

Calling scripts in an interactive shell is also a correct behaviour. Relying on `exec` in `.bashrc` will likely cause problems with any software that does that.

Maybe we need `EXPERT_SHELL` env var for cases like this, akin to `MISE_SHELL`, so user can override which shell is used by Expert?

spicychickensauce:
> What if someone has a condition in .bashrc on -l flag for some reason?

It will just run into a timeout in the worst case and fallback to the `filter_release_root_from_path()` function.

I think removing the `-i` flag is not arbitrary, as using the flag from scripts is already quite out there.

> How do you set up mise (for bash or for fish)?

I set up mise in both bash and fish via `mise activate`, so all dev tools always have the right paths, no matter which environment.
`mise doctor` output
```
version: 2026.3.8 linux-x64 (2026-03-12)
activated: yes
shims_on_path: no
self_update_available: no
```

> It seems from the problem description that your SHELL env variable is set to bash, not fish

Yes, my `$SHELL` is set to bash. I can't set this to fish because many other tools break if I do. The whole point of my setup is to stay fully compatible with bash for tools and scripts and only use fish interactively. This is the least invasive setup, [as recommended by the arch wiki](https://wiki.archlinux.org/title/Fish#Modify_.bashrc_to_drop_into_fish). This is the first time I've run into any issue with this setup.

CptnKirk:
Interactive shells also cause problems for zsh users if they or a plugin use zle.  That also results in a wedged process.  I should have checked here first before I submitted https://github.com/elixir-lang/expert/pull/537.  Happy to close mine if this is going to go through.

katafrakt:
FWIW I checked mise in path mode and it indeed seems to work fine for non-interactive shell, so my main worry about this change is gone.

apps/expert/lib/expert/port.ex
```diff
@@ -260,35 +263,69 @@ defmodule Expert.Port do
...
+        ["-l", "-c", cmd]
+
+      _ ->
+        cmd = "cd #{directory} && printf \"#{@path_marker}:%s:#{@path_marker}\" \"$PATH\""
+        ["-i", "-l", "-c", cmd]
+    end
+  end
 
-    {output, exit_code} = System.cmd(shell, args, env: env)
+  defp cmd_with_timeout(shell, args, env: env, timeout: timeout) do
```
doorgan: It's generally not a good pattern to pattern match on a keyword list as positional arguments
```suggestion
  defp cmd_with_timeout(shell, args, env, timeout) do
```

apps/expert/lib/expert/port.ex
```diff
@@ -260,35 +263,69 @@ defmodule Expert.Port do
...
+        {:ok, result} ->
+          {:ok, result}
+
+        {:error, :timeout} ->
+          if Enum.member?(args, "-i") do
+            # If the command contained the -i flag, try again without it.
+            # Some users have exec calls or blocking prompts in their .bashrc,
+            # so we would hang here without the timeout
+            args = Enum.reject(args, &(&1 == "-i"))
+            cmd_with_timeout(shell, args, env: env, timeout: 1_000)
```
doorgan:
```suggestion
            cmd_with_timeout(shell, args, env, 1_000)
```

apps/expert/lib/expert/port.ex
```diff
@@ -260,35 +263,69 @@ defmodule Expert.Port do
       {"PATH", System.get_env("PATH", @default_unix_path)}
     ]
 
-    shell_name = Path.basename(shell)
+    args = path_fetch_cmd_args(shell, directory)
+
+    maybe_cmd_output =
+      case cmd_with_timeout(shell, args, env: env, timeout: 1_000) do
```
doorgan:
```suggestion
      case cmd_with_timeout(shell, args, env, 1_000) do
```

doorgan (commented):
I tested this and it works on my setup with mise, I only have a few nitpicky comments and this is good to go

doorgan (approved):

doorgan:
Thank you! 💜 
@CptnKirk if this fixed the issues for you too, let's close #537

PR #537, opened 2026-03-31
fix(expert): default PATH lookup to noninteractive login shell

CptnKirk:
## Summary
- Change Unix PATH discovery to use a noninteractive login shell by default.
- Keep an opt-in escape hatch for the legacy interactive behavior via `EXPERT_PATH_SHELL_MODE=interactive`.
- Add regression coverage for default, interactive, and unknown shell-mode behavior.

## Why
This is a blocking fix for environments where interactive zsh startup triggers `zle`/plugin side effects. The previous default could wedge PATH discovery and prevent Expert from starting reliably.

A noninteractive login shell is the least surprising default for PATH discovery: we want environment setup, not interactive UI behavior. Interactive shell startup has side effects, and users who need that legacy behavior can still opt back in explicitly.

## Testing
- `just compile expert --warnings-as-errors`
- `just mix expert format --check-formatted`
- `just test expert --warnings-as-errors`

## Notes
Users who need the old interactive behavior can opt back in with `EXPERT_PATH_SHELL_MODE=interactive`.

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -278,7 +278,7 @@ defmodule Expert.Port do
 
         _ ->
           cmd = "cd #{directory} && printf \"#{@path_marker}:%s:#{@path_marker}\" \"$PATH\""
-          ["-i", "-l", "-c", cmd]
+          path_lookup_args(cmd)
       end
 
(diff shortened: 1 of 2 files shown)


CptnKirk:
This has been resolved by https://github.com/elixir-lang/expert/pull/527

PR #550, opened 2026-04-02
chore(just): add mix-local recipe

AlternateRT:
This is a small nicety to ensure that these components are installed when you attempt to run any of the build recepies


doorgan (approved):

doorgan:
I approved this before checking the comments in #557, the implementation here looks correct for what it intends to do but the sentiment there makes me think we don't actually want to merge this at the moment so I will close it. We can reopen this if we change opinion.

Issue #536, opened 2026-03-30
Index code from macro expansion


Draggu:
Currently indexer is not trying to perform macro expansion; we should change it. 

A simplified approach is to map everything found inside the expansion to call as a whole, without trying to guess which input node corresponds to which output node. 

In the long term, we can make it more advanced.

PR #559, opened 2026-04-08
feat: add macro expansion capability to Analysis and Indexer

### LINKED ISSUE #517: References not found if function is imported by use
Repro

```elixir
defmodule MyProject.B do
    def foo(), do: 1
end

defmodule MyProject.A do
  defmacro __using__(ast) do
    quote do
      import MyProject.B
      unquote(ast)
    end
  end
end

defmodule MyProject do
  use MyProject.A

  def test() do
    foo()
  end
end

```

run goto references on `foo` in `def foo(), do: 1`

Indexer is collecting `use`s but not resolving them. `foo()` call in `test` function is indexed as `MyProject.foo` instead of `MyProject.B.foo`.

Going the opposite way works only because we have `elixir_sense` fallback

Draggu:
Close #517 #536 

This PR uses Spitfire expand to get macro expansions in `Analysis`, then indexer can visit expanded asts too. There is assumption that all generataed code will be pointing to whole macro call. Making it more precise is complex and can be done on top of this PR


Draggu:
Seems like it won't work on 1.16 due to missing `Macro.Env` functions. I'm reluctant to copy-paste their polyfill though

katafrakt:
Regarding the polyfills, I _personally_ think that if it's only Elixir 1.16 that's affected and it's basically a new functionality, it would be ok to guard around the version and not offer this for Elixir 1.16 users. Not sure what others think.

doorgan:
I agree with @katafrakt

Issue #541, opened 2026-04-01
Production release logging is too verbose


emdvtc:
The generated log file(s) `~./.expert/expert.log` is way too big because it outputs debug logs, after one or two hours of usage I've accumulated about ~18MB of log files. I think it makes sense for production builds to only log info and above (or have it configurable). Here's the ls output

```
> ls -lah ./.expert/ | grep '.log'
-rw-r--r--@  1 emdv  staff   8.4M Apr  1 19:43 expert.log
-rw-r--r--@  1 emdv  staff    10M Apr  1 18:17 expert.log.0
-rw-r--r--@  1 emdv  staff   165K Apr  1 19:43 project.log
-rw-r--r--@  1 emdv  staff   979K Apr  1 19:27 project.log.0
```

Here's my erlang/elixir versions, with expert compiled on an M4 PRO

```
elixir 1.18.4-otp-28
erlang 28.3.1
```

I do see my IDE stuttering once in a while, or at least the "busy" icon spinning way more than usual, unsure whether it's because excessive logging or other underlying issue. Happy to do further investigations.

Thank you

katafrakt:
Can you submit the logs? Something is likely going on if they grow so fast.

emdvtc:
Had to redact some stuff out since it's a private project, here you go

[expert.log.0.redacted.txt](https://github.com/user-attachments/files/26457045/expert.log.0.redacted.txt)

katafrakt:
Thanks. From looking at the log:

1. There is a surprising amount of `debug] sent notification server -> client textDocument/publishDiagnostics`, perhaps because of a lot of compiler warnings in the project (?)
2. A lot of messages like:
```
12:00:59.138 instance_id=19D4D583D79 [debug] Node port message:     warning: [REDACTED].CommentController.init/1 is undefined (module [REDACTED].CommentController is not available or is yet to be defined). Did you mean:

          * [REDACTED].CommentController.init/1
```
This is reported in #296, no solution found

3. Also plenty of
```
12:00:59.136 instance_id=19D4D583D79 [debug] Node port message:      warning: a struct for [REDACTED]PG.Users.User is expected on struct update:

         %[REDACTED]PG.Users.User{user | password: password}

     but got type:

         dynamic()

     where "user" was given the type:

         # type: dynamic()
         # from: lib/[REDACTED]/account/create.ex:147:22
         {:ok, user} <- [REDACTED]PG.Accounts.create(Map.merge(attrs, %{password: password}))

[...]

PR #561, opened 2026-04-08
feat(expert): make log level for log files configurable

katafrakt:
This adds a command-line flag --log-level which affects the filtering of the log level in expert.log and project.log. It keeps `:debug` as default.

Addresses #541  (I don't want to close it automatically, as we may want to continue there for feedback)

First few logs on info level are kept always for debugging. This is when I set the log level to `error`:

```
17:07:10.045 instance_id=19D6DA20D51 [info] Expert v0.1.0 starting on stdio, argv=["--stdio", "--log-level", "error"]
17:07:10.046 instance_id=19D6DA20D51 [info] Log level set to error
```

### DIFF
--- a/apps/engine/lib/engine/bootstrap.ex
+++ b/apps/engine/lib/engine/bootstrap.ex
@@ -16,7 +16,8 @@ defmodule Engine.Bootstrap do
         document_store_entropy,
         app_configs,
         manager_node,
-        logger_global_metadata
+        logger_global_metadata,
+        log_level \\ :debug
       ) do
     :logger.update_primary_config(%{metadata: logger_global_metadata})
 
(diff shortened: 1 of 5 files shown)


katafrakt:
Currently this approach has too many problems to be ready for review. It affects editor log level too, among others.

emdvtc:
When reporting the verbose logging issue I've tried my hand at setting the log level and compiling the app. I found logging quite complicated, how come you don't use the standard `Logger` ?

doorgan:
@emdvtc we did until we unified elixir and LSP logs
we're not doing anything wild outside of that, just forwarding them to .log files, from both the main manager node and the engine nodes

katafrakt:
I like #563 more, I think this should be closed, because setting CLI flag is not very user-friendly.

PR #563, opened 2026-04-09
feat(expert): allow setting file log level via configuration

katafrakt:
This is separated from LSP logging configuration. It defines the log level for expert.log file via workspace configuration under key `fileLogLevel`. It only affects `expert.log` in this PR. `project.log` is handled by Engine, so it can be done as a followup (at the same time, the default level for `project.log` is `info`, not `debug`, so it's not so dire as `expert.log` in terms of growing file size.

Sending explicit `null` as `fileLogLevel` resets it to default value (`debug`).

Addresses #541

### DIFF
--- a/apps/expert/lib/expert/configuration.ex
+++ b/apps/expert/lib/expert/configuration.ex
@@ -11,21 +11,25 @@ defmodule Expert.Configuration do
   alias GenLSP.Structures
 
   @default_lsp_log_level :info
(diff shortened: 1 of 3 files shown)


doorgan (approved):
We should also update https://github.com/elixir-lang/expert/blob/main/pages/configuration.md

katafrakt:
Right. Updated.

PR #566, opened 2026-04-09
fix: rework workspace folders and project discovery

### LINKED ISSUE #542: Don't start an engine for dependencies
Currently using Go to definition on a dependency module or function causes the expert to start an engine for that dependency. This is unnecessary because the engine for the project already handles the indexing of dependency code

Questions for monorepos:
- Should this be the case for dependencies that are also a project in a monorepo or umbrella?

Let's say there's projects A and B in a monorepo, and A depends on B. If opening a file in B should start an engine, but then:

- What engine should handle the "Find references" request? If B, uses of A won't be reported. If A then those would be reported, but the more "correct" behavior would be for  B to handle it instead?

- If we decide to not start an engine for B, only for A, then this becomes tricker or wasteful depending on the order files are opened. If you open a file in B an engine gets started, then you open a file in A and another engine is started, both know about B, so we need to select one again?

### LINKED ISSUE #549: (FunctionClauseError) no function clause matching in XPExpert.ActiveProjects.active?/1
I am getting a lot of these.  Does expert support multiple project workspaces yet?

```
2026-04-03 09:50:11.833 [info] [Info  - 9:50:11 AM] Received request textDocument/codeLens before engine for  was initialized. Ignoring.
2026-04-03 09:50:11.850 [info] [Info  - 9:50:11 AM] Received request textDocument/codeAction before engine for  was initialized. Ignoring.
2026-04-03 09:50:11.924 [info] [Info  - 9:50:11 AM] Received request textDocument/documentSymbol before engine for  was initialized. Ignoring.
2026-04-03 09:50:11.926 [info] [Info  - 9:50:11 AM] Received request textDocument/documentSymbol before engine for  was initialized. Ignoring.
2026-04-03 09:50:13.869 [info] [Info  - 9:50:13 AM] Received request textDocument/codeAction before engine for  was initialized. Ignoring.
2026-04-03 09:50:15.876 [info] [Error - 9:50:15 AM] ** (FunctionClauseError) no function clause matching in XPExpert.ActiveProjects.active?/1
    (xp_expert 0.1.0-3fe1565) lib/expert/active_projects.ex:91: XPExpert.ActiveProjects.active?(nil)
    (xp_expert 0.1.0-3fe1565) lib/expert/state.ex:147: XPExpert.State.apply/2
    (xp_expert 0.1.0-3fe1565) lib/expert.ex:403: XPExpert.apply_to_state/2
    (xp_expert 0.1.0-3fe1565) lib/expert.ex:218: XPExpert.handle_notification/2
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:600: anonymous fn/2 in XPGenLSP.handle_notification_sync/3
    (xp_telemetry 1.3.0) /home/runner/work/expert/expert/apps/expert/deps/telemetry/src/telemetry.erl:324: :xp_telemetry.span/3
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:599: XPGenLSP.handle_notification_sync/3
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:502: XPGenLSP.loop/3
2026-04-03 09:50:15.876 [info] [Error - 9:50:15 AM] ** (FunctionClauseError) no function clause matching in XPExpert.ActiveProjects.active?/1
    (xp_expert 0.1.0-3fe1565) lib/expert/active_projects.ex:91: XPExpert.ActiveProjects.active?(nil)
    (xp_expert 0.1.0-3fe1565) lib/expert/state.ex:147: XPExpert.State.apply/2
    (xp_expert 0.1.0-3fe1565) lib/expert.ex:403: XPExpert.apply_to_state/2
    (xp_expert 0.1.0-3fe1565) lib/expert.ex:218: XPExpert.handle_notification/2
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:600: anonymous fn/2 in XPGenLSP.handle_notification_sync/3
    (xp_telemetry 1.3.0) /home/runner/work/expert/expert/apps/expert/deps/telemetry/src/telemetry.erl:324: :xp_telemetry.span/3
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:599: XPGenLSP.handle_notification_sync/3
    (xp_gen_lsp 0.11.3) lib/gen_lsp.ex:502: XPGenLSP.loop/3
```

### LINKED ISSUE #573: Workspace boundary set to root_uri (first folder only)
With this workspace file, only the first folder ever gets an engine.

> 
{
  "folders": [
    { "path": "anekol" },
    { "path": "anekol_phoenix" },
    { "path": "anekol_www" },
    { "path": "anekol_websites" },
  ]
}

In state.ex:
`elixirif event.root_uri do
  root_path = Document.Path.from_uri(event.root_uri)
  root_path |> Forge.Workspace.new() |> Forge.Workspace.set_workspace()
end`

In forge/project.ex, workspace_boundary_path/0 returns this single path, and boundary_reached? blocks traverse_path from walking up to find mix.exs in any sibling folder outside the first workspace folder.

So only the first folder ever gets an engine.

### LINKED ISSUE #574: workspace_folders from Initialize not used for multi-root
With this workspace file, only the first folder ever gets an engine.

{
"folders": [
{ "path": "anekol" },
{ "path": "anekol_phoenix" },
{ "path": "anekol_www" },
{ "path": "anekol_websites" },
]
}

In expert.ex the Initialize handler does register all workspace folders:
`projects = for %{uri: uri} <- workspace_folders, project = Project.new(uri), ...
ActiveProjects.set_projects(projects)`

But Project.new(uri) calls maybe_set_root_uri → Path.expand → Document.Path.to_uri, which is fine. The problem is workspace_boundary_path is already set to just anekol before these projects are even used, so when find_project is later called for files in sibling folders, boundary_reached? kills the traversal immediately.

doorgan:
Reworks workspace folders and project discovery

We are misinterpreting what workspace folders are, and we are not acknowledging "bare" projects properly, resulting in a lot of code that uses bespoke checks for mix projects, or we have situations where expectations can be broken(like a project being nil).

So far we have considered workspace folders to be "the root of a project", but that is not what the LSP spec says.
Workspace folders are the replacement for `rootUri`, which is now deprecated. `rootUri` is where the editor's workspace root is(in most cases, where the editor was started). Several editors support opening different folders within the same workspace, meaning the workspace would have several roots. Each of these workspace roots is what the LSP defines as a workspace folder. It has nothing to do with the contents of the folders, it's only a list of folders the editor has open, nothing more.

What is a language project is then up to the LSP implementations. For example, rust-analyzer uses `cargo.toml` as a signifier for a rust project, the Typescript lsp does something similar with package.json files, and so on. There may not be a project at all and the lsp would handle standalone files too. Workspace folders are, from what I could see in those lsp servers, mostly used as boundaries for "project discovery". The main point here is that it's up to the lsp implementation to determine all this if at all. The LSP is normally configured to be invoked with certain file extensions, so all the client does is:

- tell the lsp server about the workspace folders
- send document notifications and requests about files with the configured file extensions for the language 

So based on that, the goal of this PR is to:
- Store the workspace folders provided by the client and use them for project discovery. Remove the concept of "root uri" from expert internally, but fallback to it if workspace folders are not supported by the client
- Detach the idea of "workspace" with that of "project" since they are different concepts altogether
- Properly recognize bare projects as a thing that exists and force us to handle them rather than always default on "project is a mix project". There are features not possible to implement for bare projects, but a lot of them can be implemented, so the engines need to be aware of that
- Ensure elixir related documents always match against a project

I apologize for the large diff, however most of that is updating function signatures and adding or relocating tests. The files that need special attention during review are `apps/expert/lib/expert/document/lookup.ex` and `apps/expert/lib/expert/state.ex`, as those are where the main behavioral changes are.

Things I've tested:
- projects at the root of a workspace folder
- one or more projects inside the same workspace folder(ie monorepos like expert)
- projects nested inside other projects(specifically that forge fixtures start an independent engine) 
- umbrella apps(and that they start a single engine per umbrella)
- bare elixir files outside a mix project
- elixir projects in different workspace folders
- dependencies inside `deps` use the engine for the project that depends on it 

All those seem to work except for path dependencies outside a workspace folder. For example in this repo: https://github.com/doorgan/monorepo_test

If I start the editor inside `main` and then use go to definition on `Secondary`, it does find and open the file, but it will be considered a different project and start its own engine. I think one potential solution for this would be something like `cargo metadata`, but it seems like we would need to both evaluate the `mix.exs` file of the `main` app from the manager node at project discovery time, and do it in a sandboxed way(probably spawning a new node, which may be too much). I have decide to leave this as this since this change is big enough already.


katafrakt:
I went over the code. Really great work! I don't have any comments for that part. I will try to test it on some scenarios as well.

katafrakt (approved):
Ok, I ran some tests on few scenarios and editors. Everything seems in order.

Issue #507, opened 2026-03-13
Goto definition on std items is not working


mkaput:
https://github.com/user-attachments/assets/e9a185f4-629b-46c4-972d-34d6bf03e65c

```
Expert v0.1.0-rc.6-7a4e333
```

katafrakt:
This is very weird, because we can see that hover worked, and they use similar logic 🤔

katafrakt:
Actually, wait. Is it supposed to work? You'd need to have Elixir source files somewhere and Expert would have to know where. IIRC this was a separate setting on ElixirLS, but I don't think we have it in Expert.

mhanberg:
It doesn't work unless you compile elixir locally. Otherwise there is no source code to jump to. 

Most people download pre compiled elixir

mkaput:
Yup. I think it’d be best if Expert tries to autodownload sources and keep them in some kind of cache. This is what RustRover, Rust Analyzer (via src component in rustup) and CairoLS did.

alexandremcosta:
> It doesn't work unless you compile elixir locally. Otherwise there is no source code to jump to.
> 
> Most people download pre compiled elixir

@mhanberg I have elixir source in `~/.asdf/installs/elixir/1.19.5-otp-28/lib/`, but expert still can't find it. Is there I way I can point it there?

katafrakt:
@alexandremcosta unlike in ElixirLS, Expert does not have a configuration option to point it there.

However, looking at the ElixirLS code, I see this is actually implemented inside ElixirSense. It's just a matter of reading the configuration (presumably workspace setting) and setting the app env (`Application.put_env(:language_server, :elixir_src, stdlib_src_dir)`). If we are ok with that @mhanberg @doorgan it should be quite easy to do.

doorgan:
@katafrakt I think it's worth trying, we should also expose that as a configuration option and update the vscode extension accordingly

mkaput:
My follow-up to this would be to go an extra mile and just download Elixir's source code in the background from GitHub. I wouldn't want users to do this manually and put it in config.

katafrakt:
I'm going to close it, as we now have the ability to pass the path to Elixir source in configuration and we also have reasonably reliable autodetection of the source location.

PR #568, opened 2026-04-10
feat: allow setting Elixir source directory

katafrakt:
Addresses #507 (but without automatic source downloading)

Seems to work well for me with two caveats:
- There is just one value for the whole workspace. In theory a workspace can include multiple Elixir project with different version, but ElixirSense setting is global per Expert instance. Likely, this is not a huge problem, as the versions don't need to match perfectly for most lookups.
- For some reason for me it only worked with absolute paths. Probably ElixirSense implementation limitation.

### DIFF
--- a/apps/expert/lib/expert/configuration.ex
+++ b/apps/expert/lib/expert/configuration.ex
@@ -22,7 +22,7 @@ defmodule Expert.Configuration do
             workspace_symbols: %WorkspaceSymbols{},
             log_level: @default_lsp_log_level,
             file_log_level: @default_file_log_level,
-            elixir_src: nil
+            elixir_source_path: nil
 
   @type t :: %__MODULE__{
           support: support | nil,
@@ -31,7 +31,7 @@ defmodule Expert.Configuration do
           workspace_symbols: WorkspaceSymbols.t(),
           log_level: lsp_level(),
           file_log_level: file_level(),
-          elixir_src: String.t() | nil
+          elixir_source_path: String.t() | nil
         }
 
   @opaque support :: Support.t()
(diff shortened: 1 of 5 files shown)


pages/configuration.md
```diff
@@ -21,3 +22,4 @@ Expert supports the following configuration options.
 | `workspaceSymbols.minQueryLength` | integer | `2` | Minimum characters required before workspace symbol search returns results. Set to `0` to return all symbols with an empty query. |
 | `logLevel` | string | `"info"` | Minimum severity of log messages forwarded to the editor. Valid values: `"error"`, `"warning"`, `"info"`, `"log"`. |
 | `fileLogLevel` | string | `"debug"` | Minimum severity of log messages written to the log file (`.expert/expert.log`). Valid values: `"debug"`, `"info"`, `"warning"`, `"error"`, `null`. Sending `null` resets log level to default. |
+| `elixirSrc` | string | `null` | Path to a local Elixir source directory. When set, go-to-definition on Elixir standard library modules will navigate to source files in this directory instead of returning no result. Should be an absolute path. |
```
mkaput: > For some reason for me it only worked with absolute paths. Probably ElixirSense implementation limitation.

We can quickly check if this path is relative and if so, then make it absolute relevant to project working directory, aren't we?

apps/expert/lib/expert/configuration.ex
```diff
@@ -21,15 +21,17 @@ defmodule Expert.Configuration do
             additional_watched_extensions: nil,
             workspace_symbols: %WorkspaceSymbols{},
             log_level: @default_lsp_log_level,
-            file_log_level: @default_file_log_level
+            file_log_level: @default_file_log_level,
+            elixir_src: nil
```
mhanberg: Let's avoid using shortened names like src if we can. Especially for user facing properties.

katafrakt: Good point. Do you think it would be better to have `elixir_source` or `elixir_source_dir`/`elixir_source_directory`?

mhanberg: elixir_source_path or elixir_source_dir is probably good.

PR #569, opened 2026-04-11
ci: backports

mhanberg:
This sets up the ability to backport PRs to release branches in order to
aid in releasing bug fixes.

For example, we don't want to release 0.2.0 until everything on the
milestone is completed, but we still want to be able to release bug
fixes. What do we do if we make a new feature that bumps the version to
0.2?

Now when we publish a new minor release, a release branch is
automatically created. We can then apply a label (like `release:backport v0.1`)
to a pull request and it will automatically cherry pick that PR into a
PR that will be automerged into the release branch. If there are
conflicts or a test fails, we have a chance to fix it up and then we can
merge.

To enable automerge, we set a required status check on a new job called
that will be skipped if all other steps pass, or fail if another job
fails. This is possible because a skip counts as a pass for required
status checks.

Then, a release PR is opened against the release branch the same way it
is for the main branch and we can release that and it will create a
patch release on that release branch. For example currently we could
backport bug fixes or a feature to the v0.1 branch and then release
v0.1.1, without releasing all commits that are on main.

Backporting is done using a GitHub App for fine grained and short lived
access tokens and will show up as a bot user opening the pull request.
It'll leave a comment on the original PR with a link to the backported
PR, or a comment saying it failed to open the PR if something goes
wrong.

The label can be applied before the PR is merged or after.

This is a strategy used by a lot of big projects. Our exact workflow
implemented here is adapted from the Neovim project.

I have implemented this in a test repo at
https://github.com/mhanberg/backporting_demo. You can see the various
release branches and PRs that were backported, etc and the releases.

One downside (that is fixable I think in the future, but not super
important in my opinion) is that the CHANGELOG gets a little confusing.
The CHANGELOG on the main branch will only every have minor release in
it (0.1.0, 0.2.0, ..., 1.0.0 etc) and the release branches will have all
the patch releases for that minor release. But on the main branch, the
changelog will still include every single change that happened since the
previous minor release.

However, the releases page will have all of the changelogs for each
release. In the future we could potentially do some CHANGELOG merging
into the main branch, but I think the tradeoff is worth it for how we
want to release currently.

Release branches can also be useful if we say, stop supporting a given
Elixir version but a bug fix can still apply to an older version and
those who can't update can still get the fix.

**TODO**

There are still some tasks to complete before this can be merged.

- [x] Update branch protection rulesets. There needs to be required
  status checks and branch protection rules on the main and release
  branches.
- [x] Create GitHub App in the expert-lsp org and install it and
  configure it with all the right permissions, etc.
- [x] Manually create `v0.1` release branch
- [x] Manually create `release:backport v0.1` label
- [x] Satisfy/suppress zizmor comments appropriately (we have to use pull_request_target if we want the bot to be able to comment on the pull requests, so we need to suppress that one.)


doorgan (approved):
This is great, and the changes look in line with the backporting demo on your repo. I think the changelog file issue is a reasonable tradeoff and we can update it manually if needed before cutting maintenance releases.

I'm approving to unblock you merging at your convenience. I don't think I have permissions to do most of the remaining work except for fixing the zizmor comments

mhanberg:
Manually created backport PR https://github.com/expert-lsp/expert/pull/601

PR #570, opened 2026-04-11
feat: add hex intelligence

dbernheisel:
Adds hex package intelligence to `mix.exs` — package name completion, version completion, dep option suggestions, and hover documentation. Works across the public hex.pm registry and custom self-hosted repos like Oban with no configuration beyond what Mix already knows.

This is fairly self-contained with a few integrated points, so because of that, I felt a large PR would be ok for completeness. Happy to refactor/reorganize modules as you see fit. 

Generally, this is porting [hex-cmp](https://github.com/dbernheisel/hex-cmp)

## What it does

- **Package completion** — type `{:phoe` and get `phoenix`, `phoenix_live_view`, etc. with download counts and descriptions. Custom repo packages (like `oban_pro`) appear alongside hexpm results with their repo labeled.
- **Version completion** — type `{:phoenix, "~> 1.` and get filtered version candidates sorted newest-first. Retired versions are annotated with the retirement reason. Works for custom repos without needing `repo:` typed first.
- **Dep option completion** — type `{:phoenix, "~> 1.7", ` and get `only`, `runtime`, `override`, `repo`, etc.
- **Hover docs** — hover a package atom to see description, installed vs latest version, license, download count, and links. For custom repos, metadata is extracted from locally-cached tarballs.

## How it works

- **Engine.Deps** runs in the project's Mix context (where the Hex archive is loaded) and exposes repo config, dep versions, project files, and cached tarball paths over RPC.
- **Hex.Context** detects the cursor slot (name/version/opts) inside a `deps` function tuple using Sourceror's error-recovering parser, with a regex line fallback for cases where both parsers fail (unclosed strings). This assumes the conventional "deps" function within the project file (eg, `mix.exs`), and will not offer intelligence otherwise.
- **Hex.Api** routes through the hex.pm HTTP API via `:hex_core` dep for public packages and the repo protocol (`:hex_repo`) for custom repos. Custom repo responses are enriched with tarball metadata if they're already cached.
- **Hex.Cache** is a DETS-backed GenServer with 1-hour TTL and stale-on-error fallback so completions survive network failures and LSP restarts. This helps relieve the server from being hammered with repeated known info from the user. 

Testing

- [x] Verified with a custom repo, eg, oban.
- [x] Verified with public packages
- [ ] Verified with private hex repos.
- [ ] Verified with umbrella apps, but it should work since it uses the engine's current InProject helpers.
- [ ] Verified with projects with a non-standard project file, eg, not named `mix.exs`

image

image

image

image

image

image


apps/expert/lib/expert/provider/handlers/code_lens.ex
```diff
@@ -21,8 +21,6 @@ defmodule Expert.Provider.Handlers.CodeLens do
     projects = ActiveProjects.projects()
     project = Project.project_for_document(projects, document)
 
-    document = Document.Container.context_document(params, nil)
-
```
dbernheisel: not related, but a duplicated call (document is set on line 20 already)

andyl:
@dbernheisel - if this is accepted, will we still need hex-cmp ?

ps hex-cmp has been **very handy** thanks for this great tool!

dbernheisel:
Nope! It's more complete within Expert; supports custom repos, displays the installed version of the dep, and has access to the projects actual hex to know these things rather than reinventing what mix/hex do.

apps/expert/lib/expert/code_intelligence/hex.ex
```diff
@@ -0,0 +1,390 @@
...
+  def project_file?(%Project{} = project, %Document{path: path}) do
+    project_file?(project, path)
+  end
+
+  def project_file?(%Project{} = project, path) when is_binary(path) do
+    MapSet.member?(project_files(project), normalize_path(path))
+  end
+
+  defp project_files(%Project{} = project) do
+    key = {__MODULE__, :project_files, Project.name(project)}
```
katafrakt: I think it's better to use `project.root_uri` for cache keys, similar to #575

apps/engine/lib/engine/deps.ex
```diff
@@ -0,0 +1,172 @@
...
+  """
+  @spec dep_version(atom() | String.t()) :: {:ok, String.t()} | :error
+  def dep_version(app) when is_binary(app) do
+    dep_version(String.to_existing_atom(app))
+  rescue
+    ArgumentError -> :error
+  end
+
+  def dep_version(app) when is_atom(app) do
+    case Application.spec(app, :vsn) do
```
katafrakt: This only returns dependencies that are actually loaded in the engine node, which means it does not return anything for deps with `runtime: false` or `only: :dev` (because IIRC engine node is run in test env). I actually noticed that while testing, before reading the code.

I wonder if it would make sense to use something like `Mix.Dep.Lock.read`, either as an alternative or as fallback.

Here's a dev-only dep (no remote version info):

Image

and here's one with `[:dev, :test]`:

Image

apps/expert/lib/expert/code_intelligence/hex.ex
```diff
@@ -0,0 +1,390 @@
...
+
+      :error ->
+        {:error, :unknown_repo}
+    end
+  end
+
+  @doc false
+  @spec project_scope(Project.t() | nil) :: atom() | String.t()
+  def project_scope(nil), do: :__local__
+  def project_scope(%Project{} = project), do: Project.name(project)
```
katafrakt: These should be private perhaps?

And also wondering if `project.root_uri` would not be better choice for a cache key here.

apps/engine/test/engine/deps_test.exs
```diff
@@ -0,0 +1,36 @@
...
+    test "returns a string or nil (doesn't crash when Mix context is partial)" do
+      refute Deps.project_file(:app)
+      refute Deps.project_file(:umbrella)
+    end
+  end
+
+  describe "project_files/0" do
+    test "returns a list (possibly empty)" do
+      assert is_list(Deps.project_files())
+    end
```
katafrakt: I'm not sure this test is actually useful

apps/engine/test/engine/deps_test.exs
```diff
@@ -0,0 +1,36 @@
...
+      assert :error =
+               Deps.dep_version("definitely_not_a_real_atom_xyz_#{System.unique_integer()}")
+    end
+  end
+
+  describe "project_file/1" do
+    test "returns a string or nil (doesn't crash when Mix context is partial)" do
+      refute Deps.project_file(:app)
+      refute Deps.project_file(:umbrella)
+    end
```
katafrakt: The assertions and the test description here don't match. I think the biggest value is checking that it does not crash, so something like:


```suggestion
    test "doesn't crash when Mix context is partial" do
      Deps.project_file(:app)
      Deps.project_file(:umbrella)
    end
```

(not sure how others feel about test cases without assertions though)

katafrakt (commented):
I tested it and it works really well! Just left some comments, mostly about minor things.

apps/engine/test/engine/deps_test.exs
dbernheisel: you're right. I deleted it.

apps/expert/lib/expert/code_intelligence/hex.ex
dbernheisel: yes this is a better choice. I changed it to use `project.root_uri`

apps/engine/lib/engine/deps.ex
dbernheisel: good catch. I switched over to using the lock file instead, and falling back onto the Application.spec

katafrakt:
One additional thing I noticed is that the changes here don't play nice with `expert engine clean`. I'm getting this:

```
❯ ~/.local/bin/expert_darwin_arm64 engine clean --force
Kernel pid terminated (application_controller) ("{application_start_failure,xp_expert,{bad_return,{{'Elixir.XPExpert.Application',start,[normal,[]]},{'EXIT',{#{reason => enotdir,path => <<\"/Users/pawel.sw/Library/Caches/expert/hex.dets\">>,'__struct__' => 'Elixir.File.Error','__exception__' => true,action => <<\"list directory\">>},[{'Elixir.File','ls!',1,[{file,\"lib/file.ex\"},{line,2064}]},{'Elixir.XPExpert.Engine','-get_engine_dirs/0-fun-1-',3,[{file,\"lib/expert/engine.ex\"},{line,103}]},{'Elixir.Enum','-reduce/3-lists^foldl/2-0-',3,[{file,\"lib/enum.ex\"},{line,2520}]},{'Elixir.XPExpert.Engine',get_engine_dirs,0,[{file,\"lib/expert/engine.ex\"},{line,102}]},{'Elixir.XPExpert.Engine',clean_engines,1,[{file,\"lib/expert/engine.ex\"},{line,80}]},{'Elixir.XPExpert.Application',start,2,[{file,\"lib/expert/application.ex\"},{line,26}]},{application_master,start_it_old,4,[{file,\"application_master.erl\"},{line,299}]}]}}}}}")

Crash dump is being written to: erl_crash.dump...done
```

Looks like the code for `clean` only handles directories under `expert` dir. I think there's no reason to delete `hex.dets` on clean, so we need to just skip it. Not sure it needs to happen in this PR, which is already quite large. Just a prerequisite to merge this one.

mhanberg:
Could you go over this code again manually, there seems to be a lot of pointless wrapping and unwrapping of return values, unnecessary intermediate variables, and some basic things. 

I imagine the design can be improved and minimized with a closer inspection.

dbernheisel:
> Could you go over this code again manually, there seems to be a lot of pointless wrapping and unwrapping of return values, unnecessary intermediate variables, and some basic things. 
> 
> I imagine the design can be improved and minimized with a closer inspection. 

I did before I opened it and a couple times after, and I didn't see anything obvious. There are some "safe" rescues and catches that are annoying but they're at RPC boundaries that seem reasonable.

Feel free to modify it as you wish.

katafrakt:
@mhanberg I fixed some things code style wise. Do you think it's better now?

doorgan:
I'll review this. I misclicked and requested copilot's review instead of self-assigning me as a reviewer and now I can't find how to cancel it :/

apps/engine/lib/engine/deps.ex
```diff
@@ -0,0 +1,167 @@
...
+  def cached_tarball_path(_repo, _package, _version), do: :error
+
+  defp safe_call(module, fun, args) do
+    if Code.ensure_loaded?(module) and function_exported?(module, fun, length(args)) do
+      apply(module, fun, args)
+    end
+  rescue
+    _ -> nil
+  catch
+    _, _ -> nil
```
doorgan: This looks like LLM leftovers, it doesn't seem to be doing anything. `catch` is for a `throw` which is not an error mechanism in elixir.

apps/engine/test/engine/deps_test.exs
```diff
@@ -0,0 +1,23 @@
...
+  use ExUnit.Case, async: true
+
+  alias Engine.Deps
+
+  describe "dep_version/1" do
+    test "returns the version of a loaded OTP application (atom form)" do
+      assert {:ok, version} = Deps.dep_version(:elixir)
+      assert is_binary(version)
+      assert Regex.match?(~r/^\d+\.\d+/, version)
+    end
```
doorgan, line 11: I don't think this is a good test, it's testing that the `:elixir` application version matches the regex, but it would be more helpful to use a project fixture and test for the version of a dependency in that fixture. We already have a few fixtures we can use here, I think we use them mostly to test dependency completions(which this is related to)

apps/engine/test/engine/deps_test.exs
```diff
@@ -0,0 +1,23 @@
...
+
+    test "returns :error when the atom does not correspond to a loaded app" do
+      assert :error = Deps.dep_version(:this_app_is_not_loaded_at_all)
+    end
+  end
+
+  describe "project_files/0" do
+    test "doesn't crash when Mix context is partial" do
```
doorgan, line 19: It's not clear to me how this test is exercising the situation it describes

apps/expert/lib/expert/code_intelligence/hex/context.ex
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+  @spec detect(Analysis.t(), Position.t()) :: {:ok, t()} | :error
+  def detect(%Analysis{} = analysis, %Position{} = position) do
+    # Prefer Sourceror's permissive whole-file parse over the fragment-
+    # parsed `analysis.ast`. Forge's `reanalyze_to/2` pipeline truncates
+    # the document at the cursor and replaces the in-progress expression
+    # with a `:__cursor__` marker — useful for Elixir-scope awareness but
+    # destructive for tuple-in-list detection because it discards the real
+    # deps tuples. Sourceror's error-recovering parser keeps them.
```
doorgan:
```suggestion
    # The `analysis.ast` replaces incomplete expressions with a `:__cursor__`
    # marker, discarding the real deps tuples the user is writing when we
    # provide completions. Reparsing the document and locating the deps
    # tuple allows us to have the complete AST we need to suggest hex packages.
```

apps/expert/lib/expert/code_intelligence/hex/context.ex
```diff
@@ -0,0 +1,272 @@
...
+    end
+  end
+
+  # Fallback when Sourceror's parse found the `deps/0` function but no
+  # tuple literal covers the cursor position. This typically means the
+  # user is mid-typing a brand-new dep that Sourceror hasn't recovered as
+  # a tuple shape. If Forge's fragment-parsed analysis confirms a
+  # `:__cursor__` marker inside the deps body — the cleanest signal that
+  # the cursor is actually inside the list — extract a `:name` slot prefix
+  # from the raw line text and return that.
```
doorgan:
```suggestion
  # Fallback when we find the `deps/0` function but no tuple literal
  # covers the cursor position. This typically means the user
  # is mid-typing a brand-new dep that we haven't recovered as
  # a tuple shape.
```
The rest can be read from the implementation

apps/expert/lib/expert/code_intelligence/hex/context.ex
```diff
@@ -0,0 +1,272 @@
...
+
+        :error ->
+          case extract_prefix(analysis.document, position, :name) do
+            {:ok, prefix} when byte_size(prefix) > 0 ->
+              {:ok, %{slot: :name, prefix: prefix, package: nil, repo: "hexpm"}}
+
+            _ ->
+              :error
+          end
+      end
```
doorgan, line 80: I suspect this can be written as a `with`

apps/expert/lib/expert/code_intelligence/hex/context.ex
```diff
@@ -0,0 +1,272 @@
...
+      {pl, pc} >= {ol, oc} and {pl, pc} <= {cl, cc + 1}
+    else
+      {:error, :no_bounds} ->
+        {ol, oc} = {meta[:line], meta[:column]}
+        pl == ol and pc >= oc
+
+      _ ->
+        false
+    end
+  end
```
doorgan, line 129: We should avoid short var names in favor of clarity, same for other functions here. I'm not entirely sure if `pc` is meant to be `pos_character` or `pos_column` here.
```suggestion
  defp position_in?(meta, %Position{line: pos_line, character: pos_character}) do
    with {:ok, meta_line} <- Keyword.fetch(meta, :line),
         {:ok, meta_column} <- Keyword.fetch(meta, :column),
         {:ok, closing_line, closing_column} <- closing_bounds(meta) do
      {pos_line, pos_character} >= {meta_line, meta_column} and {pos_line, pos_character} <= {closing_line, closing_column + 1}
    else
      {:error, :no_bounds} ->
        {meta_line, meta_column} = {meta[:line], meta[:column]}
        pos_line == meta_line and pos_character >= meta_column

      _ ->
        false
    end
  end
```
This will probably cause a formatter warning if committed verbatim though

apps/expert/lib/expert/code_intelligence/hex/repo.ex
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  defp engine_get_repo(%Project{} = project, name) do
+    case EngineApi.call(project, Engine.Deps, :get_repo, [name]) do
+      {:ok, %{} = entry} -> {:ok, entry}
+      _ -> :error
+    end
+  rescue
+    _ -> :error
+  catch
```
doorgan:
```suggestion
```

apps/expert/lib/expert/code_intelligence/hex/repo.ex
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  defp do_resolve("hexpm", _project), do: {:ok, default()}
+
+  defp do_resolve("hexpm:" <> org, %Project{} = project) do
+    with {:ok, entry} <- engine_get_repo(project, "hexpm:" <> org),
+         {:ok, key} <- fetch_string(entry, :auth_key) do
+      {:ok, Map.merge(default(), %{api_organization: org, api_key: key})}
+    else
+      _ -> :error
```
doorgan, line 27: `engine_get_repo` and `fetch_string` both return `:error` so this is redundant
```suggestion
```

apps/expert/lib/expert/code_intelligence/completion.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Expert.CodeIntelligence.Completion do
   alias Expert.CodeIntelligence.Completion.Builder
   alias Expert.CodeIntelligence.Completion.Translatable
+  alias Expert.CodeIntelligence.Hex
+  alias Expert.CodeIntelligence.Hex.Context, as: HexContext
```
doorgan, line 5: We should avoid `as:` aliases

apps/expert/lib/expert/code_intelligence/hex.ex
```diff
@@ -0,0 +1,381 @@
+defmodule Expert.CodeIntelligence.Hex do
+  @moduledoc """
+  Orchestrates hex-aware completion and hover for `mix.exs` documents.
+
+  Uses `Hex.Context` to detect whether the cursor is inside `deps/0` and which
+  repo (default `"hexpm"`, an organization, or a self-hosted repo) the dep
+  references. Routes through `Hex.Cache` (DETS, stale-on-error) to `Hex.Api`,
+  which delegates to `:hex_api_package` from `:hex_core`.
```
doorgan: I don't think this is particularly useful, it's just describing implementation details
```suggestion
```

apps/expert/lib/expert/code_intelligence/hex.ex
```diff
@@ -0,0 +1,381 @@
...
+      list when is_list(list) ->
+        list |> MapSet.new(&normalize_path/1)
+
+      _ ->
+        MapSet.new()
+    end
+  rescue
+    _ -> MapSet.new()
+  catch
+    _, _ -> MapSet.new()
```
doorgan:
```suggestion
```

apps/expert/lib/expert/code_intelligence/hex.ex
```diff
@@ -0,0 +1,381 @@
...
+
+  def installed_version(%Project{} = project, package) when is_binary(package) do
+    case EngineApi.call(project, Engine.Deps, :dep_version, [package]) do
+      {:ok, version} when is_binary(version) -> version
+      _ -> nil
+    end
+  rescue
+    _ -> nil
+  catch
+    _, _ -> nil
```
doorgan:
```suggestion
```

doorgan: Actually most of these don't even need a `rescue`, this looks like the LLM being overly defensive for no real reason

apps/expert/lib/expert/code_intelligence/hex.ex
```diff
@@ -0,0 +1,381 @@
...
+      end
+    )
+    |> case do
+      {:ok, repos} -> repos
+      _ -> []
+    end
+  rescue
+    _ -> []
+  catch
+    _, _ -> []
```
doorgan: Same as above

apps/expert/lib/expert/code_intelligence/hex.ex
```diff
@@ -0,0 +1,381 @@
...
+    }
+  ]
+
+  @doc """
+  Returns `true` if `document` is a Mix project file (`mix.exs` by
+  convention, but anything returned by `Mix.Project.project_file/0`)
+  for `project` — including any umbrella child project files.
+  """
+  @spec project_file?(Project.t() | nil, Document.t() | String.t() | nil) :: boolean()
+  def project_file?(project, document_or_path)
```
doorgan, line 67: I'm not sure I understand why this is needed, Projects, already hold the path to the project's mix.exs file

apps/expert/test/expert/code_intelligence/hex/context_test.exs
```diff
@@ -0,0 +1,290 @@
...
+
+    assert {:ok, ctx} = detect(text)
+    assert ctx.slot == :version
+    assert ctx.package == "phoenix"
+    assert ctx.prefix == "~> 1."
+  end
+
+  test "detects :name slot in a realistic mid-edit mix.exs with a large deps list" do
+    # Reproduces the grove scenario: the user is mid-typing a new dep on
+    # the first line of an already-populated deps list. Sourceror's
```
doorgan: Generally there should not be so many mentions to Sourceror in comments, it seems the LLM saw parsing was done by Sourceror at some point and it decided every comment should obnoxiously mention it. It is also outdated because we use Spitfire now but still it's not a detail that should be leaking so much.

apps/expert/test/expert/code_intelligence/hex/repo_test.exs
```diff
@@ -0,0 +1,153 @@
...
+  alias Expert.EngineApi
+  alias Forge.Project
+
+  # All repo lookups flow through `Engine.Deps.get_repo/1` via an
+  # RPC to the project's engine node. Tests mock `EngineApi.call/4` to
+  # return whatever shape `Hex.Repo.get_repo/1` would produce for that
+  # repo — atom-keyed maps with `:url`, `:auth_key`, `:public_key`,
+  # etc. — so the assertions here verify the `:hex_core` config we
+  # assemble from that shape, not the file-parsing logic (which no
+  # longer exists in Expert's process).
```
doorgan: This is just LLM over-explanations
```suggestion
```

apps/expert/test/expert/code_intelligence/hex/repo_test.exs
```diff
@@ -0,0 +1,153 @@
...
+         %{project: project} do
+      mock_engine_repo("internal", %{
+        url: "https://hex.internal.example/repo",
+        auth_key: "tok-int-789",
+        public_key: "-----BEGIN PUBLIC KEY-----\nAAA\n-----END PUBLIC KEY-----\n"
+      })
+
+      assert {:ok, config} = Repo.resolve("internal", project: project)
+      # Self-hosted repos go through the repo protocol
+      # (`:hex_repo.get_package`), NOT the `/api/packages` endpoint.
```
doorgan: I don't care where it does NOT go
```suggestion
```

apps/expert/test/expert/code_intelligence/hex/repo_test.exs
```diff
@@ -0,0 +1,153 @@
...
+      assert config[:repo_url] == "https://hex.internal.example/repo"
+      assert config[:repo_name] == "internal"
+      assert config[:repo_key] == "tok-int-789"
+      assert config[:repo_verify] == true
+      assert config[:repo_public_key] =~ "BEGIN PUBLIC KEY"
+      # `api_url` is left at the hex.pm default — self-hosted repos
+      # don't speak the API protocol, so any code path that would hit
+      # it is a bug.
```
doorgan: ?

apps/expert/test/expert/code_intelligence/hex_test.exs
```diff
@@ -0,0 +1,681 @@
...
+  alias Forge.Project
+
+  setup do
+    path =
+      Path.join(
+        System.tmp_dir!(),
+        "hex_orchestrator_test_#{System.unique_integer([:positive])}.dets"
+      )
+
+    on_exit(fn -> File.rm(path) end)
```
doorgan: ?
Isn't this what `@tag :tmp_dir` is for?

apps/expert/test/expert/code_intelligence/hex_test.exs
```diff
@@ -0,0 +1,681 @@
...
+        """
+        |> candidates()
+        |> Enum.map(& &1.version)
+
+      assert versions == ["2.19.2", "2.18.0"]
+    end
+
+    test "returns the full release list when the prefix has no trailing version chars" do
+      # `"~>|` (cursor just past the operator, no digit yet) should still
+      # produce candidates — blink.cmp would hide them client-side if we
```
doorgan: I assume this refers to https://github.com/saghen/blink.cmp and is related to this being a port of a nvim extension

apps/expert/test/expert/code_intelligence/hex_test.exs
```diff
@@ -0,0 +1,681 @@
...
+
+      patch(EngineApi, :call, fn _, Engine.Deps, :project_files, [] ->
+        raise "engine down"
+      end)
+
+      refute Hex.project_file?(project, "/tmp/anywhere/mix.exs")
+    end
+
+    test "does not memoize an empty result — retries on subsequent calls" do
+      # Critical regression: if the engine is temporarily unavailable
```
doorgan: There can't be a regression for a brand new feature, that's when a change already landed

apps/expert/test/expert/code_intelligence/hex_test.exs
```diff
@@ -0,0 +1,681 @@
...
+      refute Hex.installed_version(project, "phoenix")
+    end
+  end
+
+  describe "candidates_for_context/2" do
+    # Direct contract tests so the per-slot dispatch doesn't rely on
+    # `Context.detect/2` being called first. `Completion.complete/4`
+    # uses this entry point to avoid re-parsing the document twice per
+    # completion request — `Context.detect` already parsed it once.
+
```
doorgan:
```suggestion
```

doorgan (commented):
There's still quite a bit of llm noise(most of my comments are about this) but generally the implementation makes sense, and from what I tested so far it works
I get the feeling the parsing implementation could be simpler but it is good enough for now

apps/engine/lib/engine/deps.ex
s3cur3: Ah, this will [catch process exits](https://elixir.hexdocs.pm/main/try-catch-and-rescue.html#exits), whereas `rescue` cannot.

doorgan: True but it's not what can happen in any of the cases in this PR where `catch` is used. Here for instance the code can raise but not cause the process to exit. In the other instances(mostly around `EngineApi` calls) the process exiting is not something we want to hide.
In my experience with LLMs(even the frontier ones) they tend to be overly defensive and add `catch` clauses where they're not needed and they're like "you're right I overreacted" when you push back

katafrakt: It's true that LLMs tend to write overly defensive code, but in defense of this one - if I see a helper called `safe_call` I assume it wraps the call in a way that won't crash. But maybe it's overly-defensive on my side.

I this case, for example, `safe_call` wraps `Hex.Repo.get_repo` which internally calls `Agent.get` which can cause an exit if the agent is not started. However, I agree that this is a situation that is totally unexpected and perhaps better to surface it.

apps/expert/lib/expert/code_intelligence/hex/context.ex
katafrakt: I generally removed most of test comments. You are right that there's too much prose duplicating what code does.

apps/expert/lib/expert/code_intelligence/completion.ex
katafrakt: in this case I'd say it's defensible, because we also have `CompletionContext` in the scope. And `HexContext` vs `CompletionContext` is clearer than `Context` vs `CompletionContext`. But if we are feeling strongly, we can remove it or rename to `Expert.CodeIntelligence.HexContext`.

apps/engine/test/engine/deps_test.exs
katafrakt: I think this is the iteration after [my previous comment](https://github.com/expert-lsp/expert/pull/570#discussion_r3130276576). TBH I'm not sure if tests in this file are in general useful in current state.

doorgan: Yeah I"m more inclined to remove them if they're not really that useful

apps/expert/lib/expert/code_intelligence/completion.ex
doorgan: `Hex.Context` does the job, I generally dislike `as:` aliases but I don't consider it a nit that should block a PR

apps/expert/lib/expert/code_intelligence/hex/context.ex
katafrakt: It can, but I'm not sure it's better, because the "happy path" is actually not all happy one

```elixir
defp cursor_fallback(%Analysis{} = analysis, %Position{} = position) do
    with true <- cursor_inside_deps?(analysis),
         :error <- version_from_line(analysis.document, position),
         {:ok, prefix} when byte_size(prefix) > 0 <-
           extract_prefix(analysis.document, position, :name) do
      {:ok, %{slot: :name, prefix: prefix, package: nil, repo: "hexpm"}}
    else
      {:ok, package, prefix} ->
        {:ok, %{slot: :version, prefix: prefix, package: package, repo: "hexpm"}}

      _ ->
        :error
    end
  end
```

doorgan: Got it, let's leave it as it is now

doorgan (approved):
@katafrakt I think we can merge this as-is and tackle refactors and other code improvements in other PRs, this one is already big and works as advertised

katafrakt:
I'm running some manual tests now and I see few things erroring out (I think they worked before, so chances are I broke it). When I fix them, I will merge and then we can followup with the code cleanup in another PR.

PR #571, opened 2026-04-11
fix(engine): respect formatter line length in Refactorex actions

gpact:
Fix Refactorex code actions to respect project formatter `line_length` when rewriting code (for example, `Remove pipe`).

`Refactorex` was calling `Sourceror.to_string/2` without forwarding `line_length` from formatter config. As a result, code actions formatted output using Sourceror’s default `line_length` of `98`, even when the project configured a different value.


doorgan (approved):
Thank you! 💜

PR #572, opened 2026-04-13
fix(release): ensure start_expert works if symlinked

AlternateRT:
With this, it should now be possible to symlink `start_expert` anywhere (e.g. `~/.local/bin/expert`).

### DIFF
--- a/apps/expert/rel/overlays/bin/start_expert
+++ b/apps/expert/rel/overlays/bin/start_expert
@@ -1,3 +1,16 @@
 #!/bin/sh
 
-exec "$(dirname -- "$0")"/plain eval "System.no_halt(true); Application.ensure_all_started(:xp_expert)" "$@"
+readlink_f () {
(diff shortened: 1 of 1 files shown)


doorgan:
Thank you! Do we also need to update `start_expert.bat` for windows?

AlternateRT:
I'm not entirely sure. It would also be impossible for me to test any update to that file as I do not use windows.

But from what I gathered by looking at the `.bat` files that ship with ElixirLS and Elixir v1.19 (`language_server.bat` and `elixir.bat` respectively), it might not be needed; upon inspecting those files, neither contained a mechanism to resolve a symlink.

I can't say this for certain, but I think they lack that mechanism because symlinking is a less common practice in Windows-land.

doorgan (approved):
We can add this for windows on a different pr if needed
Thank you! 💜

katafrakt:
I can test Windows at some point (I have Expert set up on one), but I agree this should not block this PR.

PR #578, opened 2026-04-15
fix: improve release overlays

AlternateRT:
With this, `start_expert` should only be included when building a plain release - afaict from my testing, it does not seem like this was being used by burrito releases

This also ensures that `.sh` and `.bat` executables are only included when building plain releases from Linux/macOS and Windows respectively


mhanberg:
Please change to a fix (instead of chore)

PR #579, opened 2026-04-15
fix(engine): fix interpreting words inside strings as identifiers

Draggu:
Previously Expert interpreted words inside string (not inside interpolation) leading to bugs like this
image

### DIFF
--- a/apps/engine/lib/engine/code_intelligence/entity.ex
+++ b/apps/engine/lib/engine/code_intelligence/entity.ex
@@ -1,6 +1,7 @@
 defmodule Engine.CodeIntelligence.Entity do
   alias Forge.Ast
   alias Forge.Ast.Analysis
+  alias Forge.Ast.Detection
   alias Forge.Document
   alias Forge.Document.Position
   alias Forge.Document.Range
@@ -35,6 +36,7 @@ defmodule Engine.CodeIntelligence.Entity do
       |> Engine.CodeIntelligence.Heex.maybe_normalize(position)
 
     with :ok <- check_commented(analysis, position),
+         :ok <- check_in_string(analysis, position),
          {:ok, surround_context} <- Ast.surround_context(analysis, position),
          {:ok, resolved, {begin_pos, end_pos}} <-
            resolve(surround_context, analysis, position) do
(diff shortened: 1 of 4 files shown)


apps/engine/lib/engine/code_intelligence/entity.ex
```diff
@@ -71,6 +73,14 @@ defmodule Engine.CodeIntelligence.Entity do
     end
   end
 
+  defp check_in_string(%Analysis{} = analysis, %Position{} = position) do
+    if Detection.String.detected?(analysis, position) do
+      {:error, :in_string}
```
mkaput: why can't this return `:error` like `check_commented` does?

apps/engine/lib/engine/code_intelligence/definition.ex
```diff
@@ -19,6 +19,9 @@ defmodule Engine.CodeIntelligence.Definition do
     with {:ok, _, analysis} <- Document.Store.fetch(document.uri, :analysis),
          {:ok, entity, _range} <- Entity.resolve(analysis, position) do
       fetch_definition(entity, analysis, position)
+    else
+      {:error, :in_string} -> {:ok, nil}
+      other -> other
```
mkaput: This (and change in apps/expert/lib/expert/provider/handlers/hover.ex) looks like very unidiomatic Elixir. Coercing error values in with clauses should be done in particular clauses (in this example, within `{:ok, entity, _range} <- Entity.resolve(analysis, position)` expr). TBH I don't know why `Entity.resolve` can't return `{:error, :not_found}` like it does when within a comment?

mkaput (changes requested):

apps/engine/lib/engine/code_intelligence/entity.ex
Draggu: Changed it a bit, `ElixirSense` contains the same bug. We don't want a fallback to `ElixirSense` in this case.

apps/engine/lib/engine/code_intelligence/definition.ex
Draggu: Fixed

doorgan (approved):
Thank you! 💜

PR #584, opened 2026-04-20
fix: bring back remote shell

doorgan:
Adds a new `connectionDetails` command with the required details to launch a remote shell, or connect in some other way to the running expert manager

Also adds documentation on how to use it in neovim(my personal setup). There is a companion PR to add the shell to the VS Code extension too: https://github.com/expert-lsp/vscode-expert/pull/42

I decided to implement this as a command and let the editor handle it because we can no longer rely on `epmd -names` to discover running expert instances. Only the editor has access to that, and rather than logging the manager node name and "epmd" port and asking the user to pass that to a remote shell script, I figured the most reliable way is to handle those details to the client, and let the user configure the editor to make use of them.

This also includes `:wx` to the release applications so `:observer` can be used, and makes changes such that we use the `Forge.Workspace` instead of a random project for the manager node(this was needed, since the net kernel must be started when the command is run, but no project is provided to it so the old approach wasn't viable).


doorgan:
I have an issue on my windows box that prevents me from building anything because it can't connect to hex, so I can't test it there right now

apps/expert/priv/remote_shell.sh
```diff
@@ -0,0 +1,17 @@
...
+epmd_ebin_path="$4"
+cookie="${5:-expert}"
+
+export EXPERT_PARENT_PORT="$port"
+
+exec iex \
+  --erl "-pa ${epmd_ebin_path} -start_epmd false -epmd_module ${epmd_module} -connect_all false" \
+  --name "expert-debug-$$@127.0.0.1" \
+  --cookie "$cookie" \
+  --remsh "$node_name"
```
katafrakt, line 19: I had an issue with this, causing debug shell to completely not work for me on MacOS, because of the space in the `empd_ebin_path`. Maybe I was doing something wrong, but I wasn't able to correctly escape it, so I went with a suggestion from [erl docs](https://www.erlang.org/doc/apps/erts/erl_cmd.html) to use `ERL_LIBS` instead and now it works.


```suggestion
cookie="${5:-expert}"
lib_dir=$(dirname "$(dirname "$epmd_ebin_path")")

export EXPERT_PARENT_PORT="$port"
export ERL_LIBS="$lib_dir${ERL_LIBS:+:$ERL_LIBS}"

exec iex \
  --erl "-start_epmd false -epmd_module ${epmd_module} -connect_all false" \
  --name "expert-debug-$$@127.0.0.1" \
  --cookie "$cookie" \
  --remsh "$node_name"
```

apps/expert/lib/expert/provider/handlers/commands.ex
```diff
@@ -41,6 +42,23 @@ defmodule Expert.Provider.Handlers.Commands do
           Logger.info("Reindex #{project_names}")
           reindex_all(projects)
 
+        @connection_details_name ->
+          {:ok, _} = Expert.Clustering.start_net_kernel()
+
+          epmd_module = Forge.EPMD
+          ebin_path = Path.dirname(to_string(:code.which(epmd_module)))
```
katafrakt: Perhaps we could check if this returned a binary here? Because it could return `:non_existing` in theory and this would lead to strange errors on `Path.dirname`.

apps/expert/lib/expert/provider/handlers/commands.ex
```diff
@@ -41,6 +42,23 @@ defmodule Expert.Provider.Handlers.Commands do
...
+          script_ext = if Forge.OS.windows?(), do: ".bat", else: ".sh"
+
+          %{
+            "nodeName" => to_string(Node.self()),
+            "port" => Forge.EPMD.dist_port(),
+            "cookie" => to_string(Node.get_cookie()),
+            "epmdModule" => Atom.to_string(epmd_module),
+            "epmdEbinPath" => ebin_path,
+            "debugScriptPath" => Path.join(priv_dir, "debug_shell#{script_ext}")
+          }
```
katafrakt: I was thinking that we could have a `command` entry here (or maybe log it on a debug level) with all these values concatenated in a correct order, so one could just grab it and paste into the terminal.

One reason for that is because in VSCode when spawning the terminal failed, I wasn't able to see the error for it anywhere. It could also maybe help other editors integrate.

doorgan: Good idea, I'll add that

mhanberg:
Reminder to update name to remote shell to correlate with the change in the extension.

doorgan:
Done

katafrakt (approved):
Works great now. I only tested on MacOS though. I can try to test on Windows too, but perhaps this should not block the merge.

PR #577, opened 2026-04-15
chore(just): use plain release with install recipe

AlternateRT:
Switches the `just install` recipe to building a plain release as opposed to a burrito one and gives the installation docs a much needed update.

Closes #576

PR #614, opened 2026-04-24
chore(just): apply formatting changes

AlternateRT:
Applies the `justfile` formatting changes that were removed from #577

### DIFF
--- a/.github/workflows/ci-justfile.yaml
+++ b/.github/workflows/ci-justfile.yaml
@@ -32,4 +32,4 @@ jobs:
         uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4.0.0
 
       - name: Run formatter
-        run: just --fmt --check --indentation="  "
+        run: just --fmt --check
--- a/.github/workflows/ci-justfile.yaml
+++ b/.github/workflows/ci-justfile.yaml
@@ -1,0 +1,35 @@
+name: Justfile CI
+
+on:
+  pull_request:
+    paths:
+      - "justfile"
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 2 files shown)


mhanberg:
Is there a just formatter or are these just your personal preferences?

AlternateRT:
> Is there a just formatter or are these just your personal preferences?

Yes, there is a formatter, [`just --fmt`](https://just.systems/man/en/formatting-and-dumping-justfiles.html). More specifically, I used:

```sh
JUST_INDENTATION="  " just --fmt
```

Because by default, the formatter will apply 4-spaced indentation. But if you want, we could stick to the defaults.

---

**EDIT:**
It's also possible to use:
```sh
just --fmt --indentation="  "
```

mhanberg:
Please include a CI step to run the formatter

.github/workflows/ci-justfile.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Set up just
+        uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4.0.0
+
+      - name: Run formatter
+        run: just --fmt --check --indentation="  "
```
mhanberg: Unless there is a config file for just, probably want a recipe called `just-fmt` that formats with the indentation options. 

Or, be happy with the default indentation (idc either way but I was also fine with what it was before.)

AlternateRT: Easier to just use the default

PR #634, opened 2026-04-25
perf(indexer): cache application and available modules

doorgan:
Caches the result of `Application.get_application` and `:code.all_available`

Indexing the `apps/expert` project with `main` takes over 20 seconds on my machine(sometimes as much as 28 seconds)
baseline

I profiled the indexer with `:tprof` and the reports suggested these two functions were a bottleneck(the total time is cumulative, it's distributed over the tasks created by `async_chunks` in the indexer):

```
Module                                       Function/Arity                Calls     Total           Per Call µs       %
application_controller                       get_application_module/1      438643    19.72 s         44.96             5.55
code                                         -all_available/2-F/2-1-/2     11907594  2.08 s          0.17              0.58
```

This PR introduces a cache for both. The cache also clears after the indexer runs because those values rarely change while the indexer run, but they may change between runs(like if you work in a monorepo and move files from one project to the other).

After that, the expert app indexes in a little over 10 seconds for me(there's quite some variation though, it oscilates between 10 and 14 seconds, it's still way less than 20 seconds though)
app cache

### DIFF
--- a/apps/engine/lib/engine/application.ex
+++ b/apps/engine/lib/engine/application.ex
@@ -10,6 +10,7 @@ defmodule Engine.Application do
     children =
       if Engine.project_node?() do
         [
+          Engine.ApplicationCache,
           Engine.Api.Proxy,
           Engine.Commands.Reindex,
           Engine.Module.Loader,
--- a/apps/engine/lib/engine/application_cache.ex
+++ b/apps/engine/lib/engine/application_cache.ex
@@ -1,0 +1,67 @@
+defmodule Engine.ApplicationCache do
+  @moduledoc false
+
+  use GenServer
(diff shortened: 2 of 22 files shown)


Terbium-135:
Working _very nicely_ here (indexing other project than expert):

| Branch | Time |
|----------|----------:|
| Main   | **183 sec**  |
| Cached   | **53 sec**   |

katafrakt (commented):
I'm in a bit of shock and awe, as it seems to reduce the time to do an initial indexing on my work project from about 200s to about 40s.

katafrakt (approved):

mhanberg:
swag machine

PR #641, opened 2026-04-27
perf(indexer): only use relevant extractors when indexing deps

katafrakt:
When indexing dependencies, we are only interested in the definitions. Currently we are doing a full extraction and then filtering out non-definition entries. After this change we are running more focused set of extractors - only the ones that can actually produce definitions.

Tried to benchmark this in separation and got these results (indexing Expert's deps):

```
Name                                    ips        average  deviation         median         99th %
new: deps extractors + filter          0.70         1.42 s     ±8.59%         1.38 s         1.68 s
old: all extractors + filter           0.49         2.04 s     ±1.39%         2.04 s         2.09 s

Comparison: 
new: deps extractors + filter          0.70
old: all extractors + filter           0.49 - 1.44x slower +0.63 s

Memory usage statistics:

Name                                  average  deviation         median         99th %
new: deps extractors + filter         0.97 GB     ±0.02%        0.97 GB        0.97 GB
old: all extractors + filter          1.09 GB     ±0.00%        1.09 GB        1.09 GB

Comparison: 
new: deps extractors + filter         0.97 GB
old: all extractors + filter          1.09 GB - 1.12x memory usage +0.121 GB
```

### DIFF
--- a/apps/engine/lib/engine/search/indexer.ex
+++ b/apps/engine/lib/engine/search/indexer.ex
@@ -2,6 +2,7 @@ defmodule Engine.Search.Indexer do
   alias Engine.ApplicationCache
   alias Engine.Progress
   alias Engine.Search.Indexer
+  alias Engine.Search.Indexer.Extractors
   alias Forge.Identifier
   alias Forge.ProcessCache
   alias Forge.Project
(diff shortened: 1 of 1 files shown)


doorgan:
I have a change locally that is basically the same and I see similar improvements, so this is nice
There is something going on with Github again and CI is failing but the change itself looks good

doorgan (approved):

PR #642, opened 2026-04-27
fix(refactorex): use field-level comparison for cursor detection

arctarus:
## Problem

Line-based Refactorex code actions (e.g. "Introduce pipe", "Remove IO.inspect",
"Sort nested aliases", "Rewrite function using keyword syntax") were silently
returning no results for some editors, most notably Helix.

The root cause was in `Engine.CodeAction.Handlers.Refactorex.line_or_selection/2`.
To decide whether an incoming LSP range represents a cursor position (no selection)
or an actual selection, the function used full structural equality on the `Position`
struct:

```elixir
# Before
defp line_or_selection(_, %{start: start, end: start}), do: {:ok, start.line}
```

The `Position` struct carries metadata fields beyond `line` and `character`
(namely `starting_index`, `context_line`, `document_line_count`). When an editor
constructs the `start` and `end` positions through different code paths, these
metadata fields can differ even though both positions logically point to the same
cursor location. As a result, the pattern match failed and the handler fell through
to the selection path, which attempted to parse a zero-width text fragment as an
AST — producing no actions.

## Fix

Replace the structural equality match with a field-level match that pins only
`line` and `character`:

```elixir
# After
defp line_or_selection(_, %{start: %{line: line, character: char}, end: %{line: line, character: char}}),
  do: {:ok, line}
```

This is immune to any differences in Position metadata and correctly handles the
LSP semantics: a range where start and end share the same line and character is a
cursor position, not a selection.

## Tests added

- **Regression test** — constructs two `Position` structs with the same
  `line`/`character` but a deliberately different `starting_index`, asserts they
  are structurally unequal (`refute start_pos == end_pos`), and verifies that
  `Refactorex.actions/3` still returns results via the cursor path.

- **Line-based refactoring coverage** — one "offers" test and one "applies" test
  for each previously broken feature:
  - Introduce pipe
  - Remove IO.inspect
  - Sort nested aliases
  - Rewrite function using keyword syntax / Rewrite keyword function using regular syntax
  - Rewrite if/else using keyword syntax / Rewrite if/else using regular syntax

## Files changed

| File | Change |
|------|--------|
| `apps/engine/lib/engine/code_action/handlers/refactorex.ex` | Fix `line_or_selection` pattern match |
| `apps/engine/test/engine/code_action/handlers/refactorex_test.exs` | Add regression + line-based refactoring tests |

### DIFF
--- a/apps/engine/test/engine/code_action/handlers/refactorex_test.exs
+++ b/apps/engine/test/engine/code_action/handlers/refactorex_test.exs
@@ -144,7 +144,7 @@ defmodule Engine.CodeAction.Handlers.RefactorexTest do
 
       document = Document.new("file:///file.ex", code, 0)
 
-      start_pos = Position.new(document, 1, 5)
+      %Position{} = start_pos = Position.new(document, 1, 5)
 
       end_pos = %Position{start_pos | starting_index: 0}
 
--- a/apps/engine/test/engine/code_action/handlers/refactorex_test.exs
+++ b/apps/engine/test/engine/code_action/handlers/refactorex_test.exs
@@ -146,14 +146,7 @@ defmodule Engine.CodeAction.Handlers.RefactorexTest do
 
       start_pos = Position.new(document, 1, 5)
 
-      end_pos = %Position{
-        line: start_pos.line,
-        character: start_pos.character,
-        valid?: start_pos.valid?,
-        context_line: start_pos.context_line,
-        document_line_count: start_pos.document_line_count,
-        starting_index: 0
-      }
+      end_pos = %Position{start_pos | starting_index: 0}
 
       assert start_pos.line == end_pos.line
       assert start_pos.character == end_pos.character
@@ -163,9 +156,7 @@ defmodule Engine.CodeAction.Handlers.RefactorexTest do
 
       actions = Refactorex.actions(document, range, [])
 
-      assert Enum.any?(actions, &(&1.title == "Underscore variables not used")),
-             "Expected 'Underscore variables not used' in actions #{inspect(Enum.map(actions, & &1.title))} — " <>
-               "line_or_selection fell through to the selection path because start != end despite same line/character"
+      assert Enum.any?(actions, &(&1.title == "Underscore variables not used"))
     end
   end
 end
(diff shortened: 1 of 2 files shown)


doorgan (changes requested):
The change in `refactorex.ex` is good, but the tests are full of AI generated noise. Can you please clean these up and limit them to test *only* what this PR intends to fix? In particular we want a test that verifies that for a Position struct with different metadata still resolves to the right line/selection. Thank you!

arctarus:
@doorgan thanks for your review!

I removed all not related tests and clarified the description of the tests checking that two Position structs with different metadata resolves to the right line/selection.

Hope now everything is fine :smile:

apps/engine/test/engine/code_action/handlers/refactorex_test.exs
```diff
@@ -132,4 +134,38 @@ defmodule Engine.CodeAction.Handlers.RefactorexTest do
...
+      start_pos = Position.new(document, 1, 5)
+
+      end_pos = %Position{
+        line: start_pos.line,
+        character: start_pos.character,
+        valid?: start_pos.valid?,
+        context_line: start_pos.context_line,
+        document_line_count: start_pos.document_line_count,
+        starting_index: 0
+      }
```
doorgan:
```suggestion
      end_pos = %Position{start_pos | starting_index: 0}
```

apps/engine/test/engine/code_action/handlers/refactorex_test.exs
```diff
@@ -132,4 +134,38 @@ defmodule Engine.CodeAction.Handlers.RefactorexTest do
...
+      assert start_pos.character == end_pos.character
+      refute start_pos == end_pos
+
+      range = Range.new(start_pos, end_pos)
+
+      actions = Refactorex.actions(document, range, [])
+
+      assert Enum.any?(actions, &(&1.title == "Underscore variables not used")),
+             "Expected 'Underscore variables not used' in actions #{inspect(Enum.map(actions, & &1.title))} — " <>
+               "line_or_selection fell through to the selection path because start != end despite same line/character"
```
doorgan: The assert messages are not useful
```suggestion
      assert Enum.any?(actions, &(&1.title == "Underscore variables not used"))
```

doorgan (approved):
Thanks!

PR #643, opened 2026-04-27
ci: change to taiki-e/install-action for just and fix erlang patch version issue on setup-beam

doorgan:
The `setup-just` action is failing to get `casey/just`, and the setup-beam action no longer returns the otp version we used to pin in their query to the hex builds

Moving to a different action to install just, and bumping the patch version of otp seems to fix both

### DIFF
--- a/.github/matrix.json
+++ b/.github/matrix.json
@@ -128,13 +128,13 @@
     },
     {
       "os": "ubuntu-latest",
-      "otp": "27.3.4.1",
+      "otp": "27.3.4.10",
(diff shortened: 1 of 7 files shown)


mhanberg:
I think it might be a coincidence that it works. GitHub is currently having issues and it says actions and packages are degraded.

doorgan:
Github inflicts me great pain on a weekly basis

doorgan:
It seems things are working again in the other branches, so I'll close this

Issue #146, opened 2025-09-24
Folding Range


mhanberg:
## Description

Port this feature from ElixirLS

https://github.com/elixir-lsp/elixir-ls/tree/master/apps/language_server/lib/language_server/providers/folding_range

oaksprout:
Would you want this one done? — The Elixir language server's folding-range provider is missing folds for several documented constructs, including with/for/fn blocks, delimiter pairs, indentation blocks, and charlist heredocs, even though comments, do/end blocks, and string or sigil literals already fold. If yes, we'll do it and open a PR; if not, say so (or just ignore this) and we won't touch it.

How it works: an AI agent does the work against your write-up. Before anything reaches you, the change is rebuilt and tested from a clean copy of your project, using your own test suite, and the results are kept in a record you can check yourself. A human reviews it, submits it, and answers the review. You keep full control — nothing lands unless you merge it.

We're testing whether work offered this way is useful to maintainers — a blunt "no" is as helpful as a yes.

sophiathedev:
I am working on a focused implementation for the remaining anonymous function folding case, adding `fn ... end` folding ranges and regression coverage without expanding this PR to the other unchecked constructs.

PR #36, opened 2025-07-10
feat: folding ranges

doorgan:
Closes #146

PR #649, opened 2026-04-28
feat(expert): support folding ranges

katafrakt:
While attempting to address https://github.com/expert-lsp/vscode-expert/issues/45 I realized that while we can fix the particular issue that was reported, we can't really have reliable folding based on what VSCode offers (which is basically regexes to determine start of the fold and end of the fold). For example, this will always be problematic:

```elixir
def foo do
  """
  abc
  """
  :foo
end
```

In this example there's no way to tell if `"""` is opening or closing, using just regexes.

That lead me to investigate [foldingRange request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange) in LSP spec. I realized that it's much more fitting for Elixir, probably, and decided to try to implement it.

It works pretty well in my tests and VSCode automatically picks this up when the server advertises its capabilities of defining the ranges. I haven't checked other editors yet though.

### DIFF
--- a/apps/expert/lib/expert.ex
+++ b/apps/expert/lib/expert.ex
@@ -536,6 +536,9 @@ defmodule Expert do
       %Requests.TextDocumentDocumentSymbol{} ->
         {:ok, Handlers.DocumentSymbols}
 
+      %Requests.TextDocumentFoldingRange{} ->
+        {:ok, Handlers.CodeFolding}
+
       %GenLSP.Requests.WorkspaceSymbol{} ->
         {:ok, Handlers.WorkspaceSymbol}
 
--- a/apps/expert/lib/expert/provider/handlers/code_folding.ex
+++ b/apps/expert/lib/expert/provider/handlers/code_folding.ex
@@ -1,0 +1,125 @@
+defmodule Expert.Provider.Handlers.CodeFolding do
+  @behaviour Expert.Provider.Handler
+
+  alias Expert.Document.Context
+  alias Forge.Ast
+  alias Forge.Document
+  alias GenLSP.Requests
+  alias GenLSP.Structures
+
+  @impl Expert.Provider.Handler
+  def handle(
+        %Requests.TextDocumentFoldingRange{params: %Structures.FoldingRangeParams{}},
+        %Context{} = context
+      ) do
+    %Context{document: document} = context
+    {:ok, folding_ranges(document)}
+  end
+
+  defp folding_ranges(%Document{} = document) do
+    case Ast.from(document) do
+      {:ok, ast, _comments} ->
+        ranges_from(ast)
+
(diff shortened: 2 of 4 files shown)


doorgan:
I'll give this a proper review tomorrow, but there's also https://github.com/expert-lsp/expert/pull/36 which is a draft I started last year to see how hard it'd be to port this feature from ElixirLS, you might want to take a look to see if we're covering all the cases

RedCMD:
shouldn't `do`/`end` already have the correct folding because they're brackets?

katafrakt:
@doorgan I did not see the other PR, sorry! From the brief analysis, I think the scope of the other one is much bigger. I was mostly after what the extension currently does. So my solution does not fold, for example:

-  Many consecutive commented out lines – I had it started, but I figured it's not a proper "comment block", so I removed it
- Branches on `case`/`cond` statements
- Pipe expressions
- Probably more

So if you like, we can instead try to get your PR to mergeable state. I'm not sure about this licensing thing mentioned there though.

@RedCMD not sure I understand. For now the server does not provide any folding guidance so the editors might do it on their own, either from an Expert extension (VSCode) or by other means, like tree sitter.

RedCMD:
my bad
I always thought VSCode would apply folding ranges by default to brackets
but it does not seem to be the case

might make a report on it

katafrakt:
@RedCMD from my tests yesterday it looked like

- if the language server provides folding ranges, VSCode uses them (when I just added advertising the capability, always returning empty list, I had no folds in VSCode)
- if the server does not provide, it uses [definitions](https://github.com/expert-lsp/vscode-expert/blob/main/syntaxes/elixir.json) from the extension, especially `foldingStartMarker` and `foldingStopMarker` from lines 5-6
- perhaps if this is not available, it does what you wrote

Not an expert here, just what I observed while investigating the issue with folding.

RedCMD:
would be these [definitions](https://github.com/expert-lsp/vscode-expert/blob/5b4f0b2f128c7ece913584c899013d3e419b04e0/elixir-language-configuration.json#L162-L166)

`editor.foldingStrategy`:
1. `auto` - Use a language-specific folding strategy if available, else the indentation-based one.
2. `indentation` - Use the indentation-based folding strategy.

doorgan (approved):
Works well for what it implements and code looks good

We can add the missing features from #36 in a separate PR. The old one exists because I was testing how hard would it be to port ElixirLS features to Expert(not that hard tbh) but I deprioritized to follow the milestones roadmap. Feel free to take over or reimplement it though.

The licensing comments in that PR are because I mostly copy-pasted ElixirLS implementation, and that warrants attribution

mhanberg:
Please update the tracking issue with any further enhancements that need to be added before closing.

katafrakt:
Updated #146

mhanberg:
Please update the website with this new feature

PR #651, opened 2026-04-29
perf: use spitfire ranges

doorgan:
Using `:tprof` against the indexer running on `apps/expert`, it shows `Access.get` and related are the main offenders:

```
Module                                       Function/Arity                Calls     Total           Per Call µs       %
1   Access                                       get/3                         266851794 43.56 s         0.16              9.0
2   lists                                        keyfind/3                     285151073 25.75 s         0.09              5.32
3   Access                                       get/2                         266851794 20.82 s         0.08              4.3
4   Forge.Document.Range                         contains?/2                   128818907 19.51 s         0.15              4.03
5   Enum                                         filter_list/2                 130959217 19.28 s         0.15              3.99
6   Keyword                                      -take/2-lists^filter/1-1-/2   100833583 18.09 s         0.18              3.74
7   Keyword                                      -take/2-fun-0-/2              75463142  16.02 s         0.21              3.31
8   Sourceror                                    compare_positions/2           24492485  14.55 s         0.59              3.01
9   Sourceror                                    get_node_end_position/2       18221337  11.69 s         0.64              2.42
10  Forge.Ast.Analysis                           -scopes_at/2-fun-1-/2         129889062 11.09 s         0.09              2.29
```

This is due to the sheer amounts of `node[:line]`, `node[:column]` etc that happen during indexing, mostly related to calculating the ranges for nodes via `Sourceror.get_range/1` and similar

I have a Spitfire branch that modifies the tokenizer and parser such that the ranges are produced as the code gets parsed, which removes two things:
- the extra ast traversals for precise range calculation
- a ton of `Access.get` related calls

This PR uses that Spitfire branch and updates only the indexer code to real with those ranges. I'm also trying to make the change as invisible to `Ast` callers as possible, since the `elixir` and `spitfire` parsers should still be interchangeable at this moment.

After this change, this is what I get in the trace report:

```
Module                                       Function/Arity                Calls     Total           Per Call µs       %
1   Access                                       get/3                         172785696 28.45 s         0.16              6.28
2   Forge.Document.Range                         contains?/2                   137324875 21.93 s         0.16              4.84
3   Enum                                         filter_list/2                 139510431 21.47 s         0.15              4.74
4   lists                                        keyfind/3                     204193388 20.81 s         0.1               4.59
5   Keyword                                      -take/2-lists^filter/1-1-/2   80791754  16.17 s         0.2               3.57
6   Access                                       get/2                         172785696 14.28 s         0.08              3.15
7   Keyword                                      -take/2-fun-0-/2              64588261  13.62 s         0.21              3.01
8   Forge.Ast.Analysis                           -scopes_at/2-fun-1-/2         138417653 12.23 s         0.09              2.7
9   lists                                        member/2                      69503211  10.2 s          0.15              2.25
10  Sourceror                                    compare_positions/2           15678073  10.01 s         0.64              2.21
```

Access.get dropped by about 20 seconds(cumulative across all tasks).

There's 2 important things to note though:
- This is using profiler instrumentation, which has a noticeable performance hit. The numbers seen here are exaggerated by that, but they are consistent between changes
- When I check the times reported to neovim on an actual run in my editor, I still see the indexer takes between 10 and 14 seconds to index from scratch. I started this PR before #634 and with or without changes in that branch I get similar indexer times, which makes me think it can probably have a similar impact in very large codebases but I don't have one to verify. After timing other stages of the indexer I found that of those 10-14 seconds, roughly 9-12 are spent persisting new values to disk or storing in ETS, so until I improve that I'm not sure I'm personally able to measure how truly impactful this is in large codebases.


doorgan:
Doesn't provide noticeable benefits on a real world run

PR #652, opened 2026-04-30
fix: don't index build paths

doorgan:
This drastically improves indexing times for expert itself for me(it normally takes 10-15 seconds)
Screenshot 2026-04-30 at 3 57 23 AM

I found two problems that thix fixes:
- `_build` directories were being indexed instead of source files only because we only excluded the `.expert` builds, not the user builds too. The `MIX_BUILD_PATH` set in `bootstrap.ex` took priority over anything else when determining the build path, which was wrong for the indexer in particular
- path dependencies weren't properly indexed with the fix for the above point, so workspace symbols wouldn't provide complete results

It's important to note that the reason this has such an impact for expect itself is because we were indexing the `priv/engine_source` directory with all its dependencies inside the `_build` directories and we weren't using them for anything. I do not expect this to have a big impact on other projects.

### DIFF
--- a/apps/engine/lib/engine/mix.ex
+++ b/apps/engine/lib/engine/mix.ex
@@ -6,61 +6,111 @@ defmodule Engine.Mix do
   end
 
   def in_project(fun) do
-    if Engine.project_node?() do
-      in_project(Engine.get_project(), fun)
-    else
-      {:error, :not_project_node}
+    case Engine.get_project() do
+      %Project{} = project ->
+        in_project(project, fun)
+
+      _ ->
+        {:error, :not_project_node}
     end
   end
 
-  def in_project(%Project{} = project, fun) do
-    # Locking on the build make sure we don't get a conflict on the mix.exs being
-    # already defined
+  def in_project(%Project{kind: :bare}, fun) do
+    run_and_normalize(fn -> fun.(nil) end)
+  end
 
-    old_cwd = File.cwd!()
-    project_root = Project.root_path(project)
-    build_path = Project.versioned_build_path(project)
-    app = Project.atom_name(project)
-    file = Project.mix_exs_path(project)
-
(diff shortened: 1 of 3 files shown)


katafrakt:
I'm curious, because I tried that before and didn't see any improvements. In my project, `_build` does not contains any `.ex` or `.exs` files, so it was effectively excluded anyway. Is that different in this case?

edit: ah, okay, makes sense for Expert 👍

doorgan:
Yeah, the issue is that it's indexing the engine_source packaged in the expert builds, hence I don't expect this to help projects other than expert

katafrakt:
@doorgan I was trying to test it today (after rebasing main). I open the root Expert dir in the editor, but I think it still indexes the build, as I'm seeing logs like this:

```
11:23:12.363 [warning] Timeout updating index for path: /Users/pawel.sw/dev/expert/apps/expert/_build/prod/rel/expert/lib/xp_expert-0.1.0/priv/engine_source/apps/engine/deps/gen_lsp/lib/gen_lsp/protocol/structures/rename_options.ex
```

I wonder if I'm missing something to see same improvements as you do.

doorgan:
@katafrakt do you get better results if you delete your index and try this branch again? I suspect the engine_source files are already in your index, so the indexer is trying to revisit them when reindexing, but I'm not 100% sure. I pushed a fix for this hypothesis but I'm not able to reproduce

katafrakt:
yes, I think removing old index was the key here, thanks!

katafrakt (approved):

katafrakt:
Do we want to backport this in the first place, if it mostly affects just us (Expert devs)? I can do it, but there's quite a few conflicts to resolve.

doorgan:
Not really necessary, but I think it's make it easier to backport changes that change the index if it doesn't deviate too much from main. I can look at the conflicts

Issue #477, opened 2026-03-02
[Zed Editor] XPEngine.Build crashes while typing and stops working until workspace reload


dabaer:
* Server: expert (id 0)
* Version: 0.1.0-rc.5-44a43e2
* Editor: Zed 0.225.10
* Ecto: 3.13.5
* Elixir 1.19.5-otp-28
* Erlang 28.3.1

When typing out Ecto schema definitions, I find Expert will often crash in the middle of typing a field (specifically right after I type the first letter of the field data type atom). I get a diagnostic message like this:

<img width="535" height="100" alt="Image" src="https://github.com/user-attachments/assets/eb375e29-854b-444e-90d8-1b2c15e4ab06" />

And in the LSP logs this error occurs:

```elixir
handled notification client -> server textDocument/didChange in 962µs
Node port message: 
14:08:55.914 [error] GenServer XPEngine.Build terminating
** (stop) exited in: GenServer.call(Mix.ProjectStack, {:update_state, #Function<18.71857876/1 in Mix.ProjectStack.post_config/1>}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir 1.19.5) lib/gen_server.ex:1135: GenServer.call/3
    (xp_engine 0.1.0-rc.5-44a43e2) lib/engine/mix.ex:20: anonymous fn/3 in XPEngine.Mix.in_project/2
    (kernel 10.5) global.erl:703: :global.trans/4
    (stdlib 7.2) timer.erl:599: :timer.tc/2
    (xp_engine 0.1.0-rc.5-44a43e2) lib/engine/build/state.ex:153: anonymous fn/3 in XPEngine.Build.State.compile_file/2
    (kernel 10.5) global.erl:703: :global.trans/4
    (xp_engine 0.1.0-rc.5-44a43e2) lib/engine/build/state.ex:146: XPEngine.Build.State.compile_file/2
    (stdlib 7.2) maps.erl:894: :maps.fold_1/4
Last message: :timeout
[...]

PR #654, opened 2026-04-30
fix(engine): add project stack lock

katafrakt:
My attempt to address #477 - adds additional lock around places that mutate `ProjectStack`, so they cannot step on each others feet. I'm not 100% sure about this, as I don't have a reproduction example for the bug reported. But the idea is based on Jose's comment in the issue.


doorgan (approved):
I can't reproduce either that issue. The changes look consistent to what I understand from José's comment on a related issue, so let's try and see if this helps

PR #659, opened 2026-05-05
feat(engine): expand engine builder error output

alfieV:
builds on #188
This PR expands the last line of the output into the last X lines of the failing build output (I guesstimated 50 lines should cover most cases, may need to be changed). 
I feel like it makes engine errors easier to understand and to debug


doorgan (approved):
Thank you!

PR #670, opened 2026-05-07
perf: use beam metadata for dependencies

doorgan:
BEAM metadata provide all the information we need to index definitions, but not references. We only index deps definitions, so it makes sense to index from BEAM files for them.

This reduces a lot of work by skipping most of the work Extractors do for source files, only requiring to scan a few lines of code from the source file to refine the definition range(this is a bit hacky, but it seems to work well for most code, metaprogramming would be where we need to be more careful).

It also adds a new `Manifest` for incremental indexing. We were previously relying on the `Entry.path` to get the file path to compare mtime against, but `.beam` file paths are not stored in the Entry structs(nor should they be) which required me to rethink how we do incremental indexing. I added a Manifest(and a ManifestStore module to handle its lifecycle) which holds a mapping of file -> source files + file stats and all the metadata needed to determine if a file should be reindexed, and which entries are stale and need to be removed.

Expert indexes in ~1.8 seconds for me without this change, and goes down to ~800ms with this change.

Another reason this change is interesting is that if we decide to use compiler tracers we will need a beam extractor(since compiler tracers provide us with the compile code's debug_info), and this sets the foundation for that.

This does not replace indexing of source files for the user project though, for two reasons:
- I didn't see a significant boost in performance, mostly because we still need to properly parse the original files to index references, that is where the bulk of the work in the index seems to be spent. I think using this for the project beams would be better implemented in a PR that actually makes use of compiler tracers, or that attempts to improve the indexer accuracy on generated code
- We still require parsing source files for `.exs` files like configuration and test, those don't produce .beam artifacts and therefore can't be indexed with this approach

### DIFF
--- a/apps/engine/lib/engine/commands/reindex.ex
+++ b/apps/engine/lib/engine/commands/reindex.ex
@@ -147,17 +147,24 @@ defmodule Engine.Commands.Reindex do
 
     {elapsed_us, result} =
       :timer.tc(fn ->
-        with {:ok, entries} <- Search.Indexer.create_index(project) do
-          Search.Store.replace(entries)
-        end
+        Search.Store.rebuild_index(project)
       end)
 
     Engine.broadcast(
-      project_reindexed(project: project, elapsed_ms: round(elapsed_us / 1000), status: :success)
+      project_reindexed(
+        project: project,
+        elapsed_ms: round(elapsed_us / 1000),
+        status: reindex_status(result)
+      )
     )
 
     result
   end
+
+  defp reindex_status(:ok), do: :success
+  defp reindex_status({:ok, _}), do: :success
+  defp reindex_status({:error, reason}), do: {:error, reason}
+  defp reindex_status(other), do: {:error, other}
 
   defp schedule_gc do
     Process.send_after(self(), :gc, :timer.seconds(5))
--- a/apps/engine/lib/engine/search/indexer/paths.ex
+++ b/apps/engine/lib/engine/search/indexer/paths.ex
@@ -1,0 +1,285 @@
+defmodule Engine.Search.Indexer.Paths do
+  alias Forge.Project
+
+  @indexable_extensions "*.{ex,exs}"
+
+  defstruct source_paths: [], beam_paths: []
+
+  @type t :: %__MODULE__{
+          source_paths: [Path.t()],
+          beam_paths: [Path.t()]
+        }
+
+  def for_project(%Project{} = project) do
+    %__MODULE__{
+      source_paths: source_paths(project),
+      beam_paths: beam_paths(project)
+    }
+  end
+
+  def indexable_files(%Project{} = project) do
+    source_paths(project)
+  end
+
+  defp source_paths(%Project{} = project) do
+    source_roots = source_index_roots(project)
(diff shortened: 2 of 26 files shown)


apps/engine/lib/engine/search/indexer/paths.ex
```diff
@@ -0,0 +1,285 @@
...
+    end)
+    |> Enum.flat_map(&beam_files/1)
+  end
+
+  defp beam_paths(%Project{}), do: []
+
+  defp dependency_app_names(%Project{} = project) do
+    project
+    |> mix_dependency_app_names()
+    |> MapSet.union(configured_dependency_app_names(project))
```
katafrakt, line 75: Wouldn't that effectively eliminate dependencies with `app: false`? I'm not sure how big of an edge case this is, but feels like a regression from previous approach.

apps/engine/lib/engine/commands/reindex.ex
```diff
@@ -147,18 +147,25 @@ defmodule Engine.Commands.Reindex do
 
     {elapsed_us, result} =
       :timer.tc(fn ->
-        with {:ok, entries} <- Search.Indexer.create_index(project) do
-          Search.Store.replace(entries)
-        end
+        Search.Store.refresh_index(project)
```
katafrakt: Will this actually force reindex when nothing is changed in the filesystem?

katafrakt (commented):
I mostly tested it from the usage site. Seems nice for our large codebase. Good one. I left two comments of things I noticed.

Additional worry is that this is quite large change that would make backporting potentailly hard (unless we backport this too?)

doorgan:
> Additional worry is that this is quite large change that would make backporting potentailly hard (unless we backport this too?)

This is true, though at the same time I'm not sure this belongs to 0.1, there's also the SQLite backend PR and upcoming compiler tracer PR that also significantly change the indexer and those don't belong to 0.1 either but to 0.2.
I'm not entirely sure about how we'll want to handle this

apps/engine/lib/engine/commands/reindex.ex
doorgan: It did not, fixed!

apps/engine/lib/engine/search/indexer/paths.ex
doorgan: Good catch, fixed

katafrakt:
I agree this does not belong in 0.1. One solution would be to try to get through with all the large changes you mentioned (+ perhaps Hex intelligence) and try to cut out 0.2 RC soon. However, I don't know if this would meet the goal of NextLS feature-completeness.

katafrakt (approved):

doorgan:
@katafrakt it would meet the main goals of 0.2, the rest of the milestone are bugfixes, so that strikes me as a good solution

katafrakt:
Given we're close, I think we can just merge it and try to push through with other changes for 0.2 RC, right?

doorgan:
Yes
I'll go ahead and merge this, after that I need to update the sqlite pr and finish my in-flight work for compiler tracer indexing
With that and the main bug fixes addressed, we can release a 0.2.0-rc(or just 0.2.0)

PR #673, opened 2026-05-09
feat(rename): implement textDocument/rename via Refactorex

### LINKED ISSUE #265: Audit Refactorex features
## Description

In my experience, I can't seem to trigger many of the [refactorex](https://github.com/gp-pereira/refactorex) features that we included into expert.

We need to audit all the features to confirm they work, and document them in the docs (https://github.com/expert-lsp/expert-lsp.org).

arctarus:
## Summary

Implements `textDocument/rename` LSP support by wiring Refactorex's rename API end-to-end through the Expert/Engine stack. Closes #265 (partial — rename only; prepareRename, kinds hierarchy, and kinds advertisement are deferred).

Supported rename targets (all powered by Refactorex):
- **Variables** — renames a local variable and all its references in scope
- **Functions** — renames a function definition and all its call sites within the same module
- **Constants** — renames a module attribute and all its references
- **Guards** — renames a guard and all its usages (Refactorex extracts the implementation into a new private guard)

Returns `null` (LSP-valid) when the cursor is not on a renameable symbol.

### Engine (`apps/engine`)
- **New** `Engine.Rename` — derives an AST selection from the cursor `Position` using `Code.Fragment.surround_context/2`, checks rename availability via `Refactorex.Refactor.rename_available?/2`, performs the rename via `Refactorex.Refactor.rename/3`, then diffs the result into `Forge.Document.Changes`
- `Engine.CodeMod.Format` — extracts a new `source_to_changes/2` function (accepts either a source string or an AST) to consolidate the format → stringify → diff → `Changes` pipeline; `Engine.CodeAction.Handlers.Refactorex` is updated to delegate to it, removing its private copy
- `Engine` facade — adds `defdelegate rename/3`

### Expert (`apps/expert`)
- **New** `Expert.Provider.Handlers.Rename` — LSP handler that unpacks `RenameParams`, calls `EngineApi.rename/4`, and wraps the result in a `WorkspaceEdit`
- `Expert.EngineApi` — adds `rename/4` RPC wrapper
- `Expert.fetch_handler/1` — routes `TextDocumentRename` to `Handlers.Rename`
- `Expert.State` — advertises `rename_provider: true` in `ServerCapabilities`

### Tests
- Engine unit tests (`Engine.RenameTest`) covering all 4 renamer kinds (variable, function, constant, guard) plus a negative case
- Expert integration test (`Expert.Provider.Handlers.RenameTest`) booting the full supervision stack, asserting a `WorkspaceEdit` is returned and verifying the renamed identifier appears in the final document text

## Design notes

**Selection derivation.** Refactorex renamers all declare `works_on: :selection`, so passing a line integer (as the existing `line_or_selection` code-action path does) would return no match. Instead, we use `Code.Fragment.surround_context/2` (1-based line/column) to find the identifier bounds at the cursor, slice the document fragment, and parse it with `Sourceror.parse_string/2` preserving `line:`/`column:` metadata. Refactorex's `AST.equal?/2` then matches the selection by atom name against nodes in the full-document zipper.

**Diffing.** `Refactor.rename/3` returns a complete new source string. We apply a shared format-then-diff pipeline (`Engine.CodeMod.Format.source_to_changes/2`) to produce minimal, formatter-stable `TextEdit`s — the same pipeline used by the code-action handler.

**Position indexing.** `Forge.Document.Position` is already 1-based (the LSP→Forge conversion adds `lines.starting_index = 1`). `Code.Fragment.surround_context/2` also expects 1-based input, so no extra conversion is needed inside `Engine.Rename`.

## Not included

- `textDocument/prepareRename` — deferred; would return the identifier range before the client shows the rename prompt
- Multi-file / cross-module renames — Refactorex operates on a single document; `RenameFunction` only updates callers within the same `defmodule`


katafrakt:
I wonder if we want to do this via Refactorex in the end. For context, I have my own version of reaming support at https://github.com/katafrakt/expert/commit/e2a6e071f9a6fd01a043cc65f908a199a7848155 which is done without it. We released it internally at Jump and I haven't received any complaints. It also supports `prepareRename` and renaming across multiple files using Indexer and references.

What I think is that we mainly need a strategic decision about Refactorex - if we want to use it as much as possible, as little as possible or maybe on a case-by-case basis.

arctarus:
@katafrakt thanks for the context!

Given that you already have a rename implementation with `prepareRename` and cross-file support through the Indexer/references, I’m fine with closing this PR. That approach sounds like a better long-term fit.

I’d be very thankful if you could release that rename support in the public version of Expert, though. It would be a great improvement for users.

Also, is #265 still relevant after this discussion? If there are still Refactorex features or other refactorings that need auditing, wiring, testing, or documentation, I’d be happy to help.

apps/engine/lib/engine/rename.ex
```diff
@@ -0,0 +1,47 @@
...
+  alias Refactorex.Refactor
+
+  require Logger
+
+  @spec rename(Document.t(), Position.t(), String.t()) ::
+          {:ok, Forge.Document.Changes.t()} | {:ok, nil} | {:error, term()}
+  def rename(%Document{} = doc, %Position{} = position, new_name) when is_binary(new_name) do
+    source = Document.to_string(doc)
+
+    with {:ok, ast} <- Sourceror.parse_string(source),
```
doorgan, line 15: Direct calls to `Sourceror.parse_string` should be avoided in favor of functions in the `Forge.Ast` module

doorgan (commented):
This works, however renaming public functions will cause compilation errors if it's used in other files

@katafrakt I checked your branch and I think yours is a more sound approach by relying on the indexer. The approach in this PR is local, causing the issue I mentioned above. Feel free to open a PR for it!

@arctarus yes, mostly for local code actions that don't require cross-file support. I also opened #678 because the Refactorex author seems to be unavailable lately and there's bugs related to it that we need to fix, so porting it and/or its features would be welcome

arctarus:
Thanks, @doorgan! I’m going to close this PR and will try to contribute to the new issue.

PR #682, opened 2026-05-14
feat: configure toolchain paths

### LINKED ISSUE #389: Configurable elixir and Erlang
## Problem

We have a lot of issues finding the appropriate elixir or Erlang version to use for building the engine and running the users code. 

## Solution

Allow the user to bypass this by configuring them explicitly. 

These values should be paths to the executables. 

The schema of this configuration can be bike shed during the pull request phase.

## Notes

We may want to introduce a config file like .expertrc.json.

They could have workspace based sub configs in the top level config file as well as inherit config files in the root of a workspace folder. 

```
root/
  .expertrc.json
  /services
    foo/
      mix.exs
      lib/
      .expertrc.json
    bar/
      mix.exs
      lib/
      .expertrc.json
```

doorgan:
Close #389 

Adds two new configuration options to allow users to specify their paths to the elixir and erlang executables, as an escape hatch if our discovery logic doesn't work for their setup: `elixirExecutablePath` and `erlangExecutablePath`

There are 3 commits in this PR, I recommend reviewing them one at a time. The first commit is the main one that implements the feature. The other two are refactors I thought necessary now that we have several config options, and to cleanup the rot that was building up in `Expert.Port`

There's also a few changes to the env vars we scrub, I was getting crashes when pointing to a different toolchain and scrubbing those extra vars solved the issue.

NOTE: This does not support configuring the toolchain per-project, I'm not yet sure of a clean way to implement that because projects are dynamically discovered, and requiring a file that needs to be added to `.gitignore` does not strike me as the right solution at this moment.

### DIFF
--- a/apps/expert/test/expert_test.exs
+++ b/apps/expert/test/expert_test.exs
@@ -19,6 +19,7 @@ defmodule Expert.ExpertTest do
     # These tests call `Expert.handle_info/2` directly (bypassing `Expert.Application`),
     # so we must start the stores that runtime boot normally provides.
     start_supervised!({Forge.Document.Store, derive: [analysis: &Forge.Ast.analyze/1]})
+    start_supervised!({Task.Supervisor, name: :expert_task_queue})
     start_supervised!({Expert.Project.Store, []})
 
     # window/logMessage comes from Logger via WindowLogHandler,
@@ -29,6 +30,7 @@ defmodule Expert.ExpertTest do
     :logger.update_handler_config(:default, :level, :none)
 
     on_exit(fn ->
+      :persistent_term.erase(Expert.Configuration)
       Forge.Workspace.set_workspace(nil)
       Logger.configure(level: :none)
       :logger.update_handler_config(:default, :level, :all)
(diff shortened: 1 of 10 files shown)


apps/forge/test/forge/os_test.exs
```diff
@@ -0,0 +1,26 @@
...
+
+    test "returns :unix for non-win32 systems" do
+      patch(Forge.OS, :type, {:unix, :darwin})
+
+      assert Forge.OS.os_family() == :unix
+    end
+  end
+
+  describe "os_type/0" do
+    test "eturns the first element of :os.type/0" do
```
katafrakt:
```suggestion
    test "returns the first element of :os.type/0" do
```

apps/expert/test/expert_test.exs
```diff
@@ -183,6 +185,55 @@ defmodule Expert.ExpertTest do
     assert config.workspace_symbols.min_query_length == 2
   end
 
+  test "restarts pending projects when runtime executable configuration changes" do
+    project = Fixtures.project()
+    test_pid = self()
```
katafrakt, line 190: do we need to capture this here? The other test does not do it, so maybe for the sake of consistency these two tests could look the same

katafrakt (approved):
Looks good from the code, although I haven't really tested this extensively.

apps/expert/test/expert_test.exs
doorgan: This is needed, but good catch on the other test, that's the one where the issue is. That's fixed now.

PR #683, opened 2026-05-16
feat: allow multiple index writers with SQLite

### LINKED ISSUE #353: SQLite
## Description

Next LS uses SQLite as the persistence layer for the symbol index. This is a feature we want to explore bringing over to Expert to see if it has advantages still over what was brought from lexical.

doorgan:
Closes #353 

Replaces the indexer's ETS backend with SQLite. The ETS backend did not have good concurrent writes support, and we lost that completely after we moved to EPMDless deployments because project nodes could no longer discover each other to assign a leader writer.

This PR replaces the ETS backend with an SQLite one which solves this issue. It is not a faster backend, the benchmarks I ran(included in this PR) range from 1.5x to 3x slower, however when I try out the features it doesn't seem like that slowdown manifests noticeably during normal use, or even while indexing. YMMV very if you have a large codebase(which I don't have available to test against). This also sits on top of several optimizations I've been working on to be able to afford this switch, so overall this should be faster than v0.1 anyways.

I had tried to implement this several months ago to no avail due to namespacing: we don't have a clean way to namespace NIFs which causes crashes at runtime, and at the time the solution seemed to be to fork Exqlite and have our own CI tooling to release our own namespaced Exqlite, which I ultimately decided against due to the sheer complexity of that approach.

This PR instead makes two changes, based on how NextLS works: instead of having the engine nodes own their indexer storage, we move the storage layer to the manager node(where we can exclude apps from namespacing) and make the indexer in the engine nodes send the data back and forth with the manager to store or retrieve indexer information. Then we can swap the ETS backend to SQLite with relatively minimal effort.

This PR *does not* delete the ETS backend code even if it becomes unused. I think it's worth keeping around for a little while to be able to easily compare how they perform and to be able to benchmark them side by side. The ETS code can be deleted in a later PR when we're sure we are not going back. I personally dislike deleting "dead code" and having to revert commits to test older approaches during a transition period.


apps/expert/lib/expert/search/store/backends/sqlite.ex
```diff
@@ -0,0 +1,851 @@
...
+  end
+
+  def do_replace_all(%State{} = state, entries) when is_list(entries) do
+    transaction(state, fn ->
+      with :ok <- exec(state, "DELETE FROM entries"),
+           :ok <- exec(state, "DELETE FROM structures") do
+        insert_entries(state, entries)
+      end
+    end)
+  end
```
katafrakt, line 266: I would flag is as potentially problematic. Here's what I see on disk:

```
drwxr-xr-x@ 5 pawel.sw  staff   160B Jun 25 14:49 ./
drwxr-xr-x@ 3 pawel.sw  staff    96B Jun 25 14:49 ../
-rw-r--r--@ 1 pawel.sw  staff   2.3G Jun 25 15:04 source.index.sqlite3
-rw-r--r--@ 1 pawel.sw  staff   4.7M Jun 25 14:59 source.index.sqlite3-shm
-rw-r--r--@ 1 pawel.sw  staff   2.3G Jun 25 14:59 source.index.sqlite3-wal
```

The WAL file is so large, presumably because of this "write all" transaction. As far as I understand, SQLite does not truncate WAL automatically, so we would need to truncate it manually and set `PRAGMA journal_size_limit`.

Other option would be to not do a single transaction, but insert in batches and avoid WAL getting this large in the first place. The risk seems to be that if something crashes between batches, we'd end up with an incomplete index - but I believe this can be addressed in some other way.

apps/expert/lib/expert/search/store/backends/sqlite.ex
```diff
@@ -0,0 +1,851 @@
...
+
+  defp create_entries_table(%State{} = state) do
+    exec(state, """
+    CREATE TABLE IF NOT EXISTS entries (
+      id INTEGER,
+      path TEXT NOT NULL,
+      subject TEXT NOT NULL,
+      type BLOB NOT NULL,
+      subtype BLOB NOT NULL,
+      block_id BLOB NOT NULL,
```
katafrakt: Any reason to not use integer for `block_id` and text for `subtype`?

apps/expert/lib/expert/search/store/backends/sqlite.ex
```diff
@@ -0,0 +1,851 @@
...
+      structure BLOB NOT NULL
+    )
+    """)
+  end
+
+  defp create_indexes(%State{} = state) do
+    with :ok <-
+           exec(
+             state,
+             "CREATE INDEX IF NOT EXISTS entries_subject_idx ON entries (subject, type, subtype, path)"
```
katafrakt: I think `path` is not used in queries but maybe I missed something

katafrakt (commented):
This looks really good. I left some comment, I think most important one is about WAL size.

And I'm not sure this should be a `fix` type, sounds like a `feat` (new SQLite backend) to me.

apps/expert/lib/expert/search/store/backends/sqlite.ex
doorgan: Index entries are deleted by path, there's also a few queries to find entry siblings by path

apps/expert/lib/expert/search/store/backends/sqlite.ex
doorgan: I decided to remove the WAL altogether, reading its docs I realized that it prevents concurrent writes which is the entire point of this PR. I've not seen issues doing so, and next-ls doesn't use WAL either

apps/expert/lib/expert/search/store/backends/sqlite.ex
doorgan: Updated in [d1a9f74](https://github.com/expert-lsp/expert/pull/683/commits/d1a9f745bb19c1ca49ab886f5817cdd8bb3fac3d)

apps/expert/lib/expert/search/store/backends/sqlite.ex
katafrakt: Yes, but I think the queries you mentioned are already covered by other indexes, for example deletion by path by `entries_path_idx`. But anyway, I see it's removed.

katafrakt (approved):
Re-tested after the changes. DB file is not 940M (down from 2.3G) and no WAL. Some queries seem quite slow, I've seen some timeouts. But I think it's fine to merge this and do additional performance fixes afterwards (if you disagree, I can dive deeper into these issues).

doorgan:
The timeouts worry me. Your index is about 10 times larger that the largest one I have, if you have some time I'd appreciate if you could investigate those before we merge. We should now be able to EXPLAIN ANALYZE the slow queries, hopefully it's not too cumbersome.

katafrakt:
Sure I'll prioritize that for next few days.

katafrakt:
Here's my branch with some improvements: https://github.com/expert-lsp/expert/commits/katafrakt/sqlite-perf/

* 37b2fd93a78bb940dd6da794a97fda1d4346ba48 – the major cause of slowness was using `backend.reduce` and iterating over all entries on the application side like with ETS. Instead, it changes implementation to use more focused queries and drops the `reduce` completely.
* 4cfb2fa30a7da4c54236df2d63545f7f771d0b26, fbbf225666528f2cbd63bdf4c00e1606d37c463c – add more indices for common querying patterns
* 8df1a45f92112da2825baed126876f0bbf5e4ddc – for fuzzy search, only fetch the columns that are needed, specifically skip `entry` BLOB
* 2dda55e737a1d3d18162aa3a389327394952d8a7 – this extracts the `entry` BLOB content to a separate table, joined on demand, similar to how Postgres' extended storage works. Now when a large table scan is needed, the BLOBs are omitted, which gives surprisingly good performance improvements

In my tests these are generally enough, I usually only see one slow query (2s) per Engine start and the rest is very fast.

doorgan:
@katafrakt thank you so much for taking a look and coming up with the improvements! I'll merge this then, and we can open a PR with your changes

mhanberg:
image

PR #695, opened 2026-05-25
fix(expert): use document from request as fallback when document is not in the store

katafrakt:
Manual backport of #693

### DIFF
--- a/apps/expert/lib/expert.ex
+++ b/apps/expert/lib/expert.ex
@@ -113,9 +113,9 @@ defmodule Expert do
 
   def handle_request(request, lsp) do
     with {:ok, handler} <- fetch_handler(request),
-         {:ok, request} <- Convert.to_native(request),
-         :ok <- check_engine_initialized(request),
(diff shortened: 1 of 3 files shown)


apps/expert/lib/expert.ex
```diff
@@ -113,9 +113,9 @@ defmodule Expert do
 
   def handle_request(request, lsp) do
     with {:ok, handler} <- fetch_handler(request),
-         {:ok, request} <- Convert.to_native(request),
```
katafrakt, line 116: It seems in v1 we had first `Convert.to_native`, then `check_engine_initialized`, but in main the other way around. Not sure when/why this happened.

katafrakt:
This is a bit non-trivial backport because of differences between v0.1 and main, so if you could have a look when you have time @doorgan I'd be grateful for a second pair of eyes.

doorgan (approved):
Looks good and worked in my tests!

PR #698, opened 2026-05-28
fix(engine): increase build timeout duration, add batch uri reindex with debounce

### LINKED ISSUE #672: Debounce project compilation
Some users report that Expert recompiles and reindexes the project too often. In particular, it seems coding agents can launch an LSP(including expert), and when making tons of edits they save a lot of files in quick succession, apparently causing a lot of CPU load.

I think debouncing the project compilation + indexing every few seconds can help with that use case.

guetnga:
- Increased `Engine.Build` compilation timeout from 50ms to 1000ms (do not know which value will fit)

- Slightly reworked `Engine.Commands.Reindex`: now reindex adds URIs into `pending_uris` MapSet and process them after 1000ms debounce 

Closes #672

### DIFF
--- a/apps/engine/lib/engine/commands/reindex.ex
+++ b/apps/engine/lib/engine/commands/reindex.ex
@@ -19,16 +19,20 @@ defmodule Engine.Commands.Reindex do
 
     require Logger
 
-    @debounce_interval_millis 1000
+    @default_debounce_interval_millis 1000
 
     defstruct reindex_fun: nil,
               index_task: nil,
               pending_updates: %{},
               pending_uris: MapSet.new(),
-              debounce_timer: nil
+              debounce_timer: nil,
+              debounce_interval_millis: @default_debounce_interval_millis
 
-    def new(reindex_fun) do
-      %__MODULE__{reindex_fun: reindex_fun}
+    def new(reindex_fun, debounce_interval_millis \\ @default_debounce_interval_millis) do
+      %__MODULE__{
+        reindex_fun: reindex_fun,
+        debounce_interval_millis: debounce_interval_millis
+      }
     end
 
     def set_task(%__MODULE__{} = state, {_, _} = task) do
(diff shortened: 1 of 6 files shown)


apps/engine/lib/engine/commands/reindex.ex
```diff
@@ -35,26 +45,48 @@ defmodule Engine.Commands.Reindex do
...
 
-        _ ->
-          :ok
+      if state.debounce_timer do
+        Process.cancel_timer(state.debounce_timer)
       end
 
-      state
+      timer =
+        Process.send_after(self(), :flush_pending, state.debounce_interval_millis)
```
doorgan: The message should be `{:flush_pending, timer_ref}`, where `timer_ref` is created with `make_ref/0` and should be stored in the server state, then matched against in `handle_info`
Otherwise, if this timer becomes stale due to a new reindex running, it won't make this run while there is another reindex running

apps/engine/test/engine/commands/reindex_test.exs
```diff
@@ -15,7 +15,7 @@ defmodule Engine.Commands.ReindexTest do
       Process.sleep(20)
     end
 
-    start_supervised!({Reindex, reindex_fun: reindex_fun})
+    start_supervised!({Reindex, reindex_fun: reindex_fun, debounce_interval_millis: 0})
```
doorgan: Can we add some tests that set a non-zero debounce interval, and check that the debounce is working? 100 or 500ms should work

apps/engine/lib/engine/commands/reindex.ex
```diff
@@ -35,26 +45,48 @@ defmodule Engine.Commands.Reindex do
...
 
-    def reindex_uri(%__MODULE__{} = state, uri) do
-      case entries_for_uri(uri) do
-        {:ok, path, entries} ->
-          put_in(state.pending_updates[path], entries)
+    def flush_pending_uris(%__MODULE__{index_task: nil} = state) do
+      Enum.each(state.pending_uris, fn uri ->
+        case entries_for_uri(uri) do
+          {:ok, path, entries} ->
+            Search.Store.update(path, entries)
```
doorgan: Probably cleaner as a comprehension (not tested)
```suggestion
      for uri <- state.pending_uris,
        {:ok, path, entries} <- entries_for_uri(uri) do
          Search.Store.update(path, entries)
      end
```

doorgan (commented):
Thank you! I left you some comments, there's also some [failing tests](https://github.com/expert-lsp/expert/actions/runs/26752403752/job/78900446236?pr=698) that need fixing

guetnga:
Thanks for the feedback! I’ll push the fixes shortly.

Is it possible to run all tests locally, so I could be sure that CI passes?

doorgan:
@lsdrfrx yes you can run `just test` to run all tests, or `just test engine`, `just test expert` to run tests for a specific app
You need https://github.com/casey/just to run those

doorgan (approved):

PR #699, opened 2026-05-28
fix(expert): remove release root entries from subprocess PATH

jollyjerr:
Hello expert team 👋🏼 this PR fixes an issue I have been encountering [after this pr](https://github.com/expert-lsp/expert/pull/682) using this setup:

```
ide: neovim v0.12.0
expert install method: build from source with `just release`
language version manager: mise
```

When launching neovim in a project directory, expert will crash with:

```
LSP[expert] [my-project] Failed to start engine my-project::24676988: {:error, "Failed to determine Elixir/OTP runtime for project: Runtime terminating during boot ({'cannot get bootfile','/Users/me/cod
e/opensource/expert/apps/expert/_build/prod/rel/plain/bin/start.boot'})\r\n\r\nCrash dump is being written to: erl_crash.dump...done (status 1)"}
```

I believe this is happening because we are passing `ROOTDIR=""` to `erlexec` [here](https://github.com/expert-lsp/expert/blob/fb358158ba7d2b2f33abbaf4f15a52cbca004f86/apps/expert/lib/expert/port.ex#L366) so we are skipping [this logic](https://github.com/erlang/otp/blame/a294e52685847f5108a07507152d6ff0c540dea6/erts/etc/common/erlexec.c#L1725)

<details>
  <summary>Example of bug</summary>

```sh
ERLEXEC=~/.local/share/mise/installs/erlang/26.2.5.15/erts-14.2.5.11/bin/erlexec
BINDIR=~/.local/share/mise/installs/erlang/26.2.5.15/erts-14.2.5.11/bin

env -i HOME="$HOME" PATH="$PATH" BINDIR="$BINDIR" "$ERLEXEC" -noshell -eval 'halt().' 2>&1; echo "exit: $?"

# prints exit: 0

env -i HOME="$HOME" PATH="$PATH" ROOTDIR="" BINDIR="$BINDIR" "$ERLEXEC" -noshell -eval 'halt().' 2>&1; echo "exit: $?"

# prints
# Runtime terminating during boot ({'cannot get bootfile','/bin/start.boot'})
# Crash dump is being written to: erl_crash.dump...done
# exit: 1
```

</details>

By omitting `@scrubbed_env_vars` entirely, I was able to get expert working again with my setup.

I believe this will also be the fix for https://github.com/expert-lsp/expert/issues/637

Issue #637, opened 2026-04-26
Unable to start, start.boot missing


jacobat:
Trying to run expert lsp with neovim I get the following error in the lsp.log:

```
Engine build failed for my_app::7759237
Build script exited with status: 1
Crash dump is being written to: erl_crash.dump...done
```

One of the first lines of the `erl_crash.dump` says:

```
Slogan: Runtime terminating during boot ({'cannot get bootfile','/home/linuxbrew/.linuxbrew/Cellar/expert/0.1.2/libexec/expert/bin/start.boot'})
```

And surely, that file does not expert.

I've installed expert using homebrew under WSL2. Mise is used to manage Erlang/Elixir if that makes a difference.

I've tried copying `start.boot` from `/home/linuxbrew/.linuxbrew/Cellar/expert/0.1.2/libexec/expert/releases/0.1.2/start.boot` which leads to the error:
```
Slogan: Runtime terminating during boot ('cannot expand $RELEASE_LIB in bootfile')
```

Neovim version 0.12.2
Expert 0.1.2

mhanberg:
CC @AlternateRT  seems like a homebrew issue not working on linux

AlternateRT:
I don't think this issue is specific to Homebrew. I was able to reproduce this on macOS with Zed using a plain install of v0.1.2 built from source (**no Homebrew** involved). I got the exact same error:

```
[expert] Engine build failed for expert::10703338.
Build script exited with status: 1
Crash dump is being written to: erl_crash.dump...done
```

The `erl_crash.dump` (located at the root of the expert project) had this in the first 8 lines:

```
=erl_crash_dump:0.5
Fri May  1 00:06:31 2026
Slogan: Runtime terminating during boot ({'cannot get bootfile','/Users/notimportant/.local/libexec/expert/bin/start.boot'})
System version: Erlang/OTP 27 [erts-15.2.7] [source] [64-bit] [smp:18:18] [ds:18:18:10] [async-threads:1] [jit]
Taints: 
Atoms: 3323
Calling Thread: scheduler:3
=scheduler:1
```

However, I was only able to reproduce it **once**. After I restarted my editor, everything seemed to go smoothly.

Meanwhile, I have not been able to reproduce this with a Homebrew-managed install of Expert.

mhanberg:
I figured there might be something going on since you introduced the libexec directory, was thinking maybe something with the symlink or copying stuff over wasn't working. 

[...]

PR #703, opened 2026-06-01
fix(expert): always scrub env vars

katafrakt:
While checking #699 I found that if we use `fallback_executable` flow to find Elixir, we are not scrubbing the env, leading to potential env leaking (and perhaps to #637).

I moved the scrubbing down, to `open_executable`, to not have to remember to do that every time we add a executable finding way.

### DIFF
--- a/apps/expert/lib/expert/port.ex
+++ b/apps/expert/lib/expert/port.ex
@@ -363,10 +363,6 @@ defmodule Expert.Port do
     {key, path}
   end
 
-  defp sanitize_system_env_var({key, _value}, _path) when key in @scrubbed_env_vars do
-    {key, ""}
-  end
-
   defp sanitize_system_env_var({key, value}, _path) do
     {key, value}
   end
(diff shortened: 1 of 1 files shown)


apps/expert/lib/expert/port.ex
```diff
@@ -384,15 +380,25 @@ defmodule Expert.Port do
...
+      opts
+      |> Keyword.update(:env, scrub_env([]), &scrub_env/1)
+      |> Keyword.update!(:env, &ensure_charlists/1)
 
     open_port(Forge.OS.os_type(), executable, opts)
   end
 
+  @doc false
+  def scrub_env(env) do
+    already_set = env |> MapSet.new(&elem(&1, 0))
```
doorgan:
```suggestion
    already_set = MapSet.new(env, &elem(&1, 0))
```

apps/expert/lib/expert/port.ex
```diff
@@ -384,15 +380,25 @@ defmodule Expert.Port do
 
   defp open_executable(executable, opts) do
     opts =
-      if Keyword.has_key?(opts, :env) do
-        Keyword.update!(opts, :env, &ensure_charlists/1)
-      else
-        opts
-      end
+      opts
+      |> Keyword.update(:env, scrub_env([]), &scrub_env/1)
```
doorgan, line 384: Third arg here can just be an empty list, no? Or do we want to keep the `scrub_env` call in case we add default envs later on?

```suggestion
      |> Keyword.update(:env, [], &scrub_env/1)
```

doorgan (approved):
Works on my end

apps/expert/lib/expert/port.ex
katafrakt: I think we don't want an empty list here ever. If there is no element with `:env` key, we want a default list or "nullified" env vars, i.e. `ERL_AFLAGS=""` etc. `scrub_env([])` will do that.

apps/expert/lib/expert/port.ex
katafrakt: done

PR #705, opened 2026-06-01
feat: use compiler tracers for project indexing

### LINKED ISSUE #356: Compiler Tracers
## Description

Next LS uses compiler tracers to build its symbol index, whereas Expert (from Lexical) uses code traversal with some macro expansion (essentially pseudo compilation) instead.

We should explore the differences and advantages of both and see if we should introduce some compiler tracing.

### LINKED ISSUE #536: Index code from macro expansion

doorgan:
The main idea of this PR is that we can use Compiler Tracers for all compiled files like NextLS does. This allows us to index definitions and references *as the project compiles* rather than as a separate step. As a consequence, the indexer no longer goes over .ex files, as those are compilable, and indexes all .beam files instead if a compiler tracer missed them.

Compiler tracers provide all the information we need: aliases, alias references, remote/imported functions/macros, local functions/macros, struct expansion, and definitions(via the module's compiled bytecode), which in principle should allow us to skip indexing from source code, or in the worst case scenario reduce the source analysis to range refinement, which is faster than full AST analysis.

For non-compiled files like .exs scripts and tests, we still run the source code based indexer so we don't lose support for those.

At a high level this works as follows:
- File is saved, compilation is scheduled
- The compiler emits tracer events, we pick them up and update both the index and the indexer manifest
- The indexer runs, skips the .beams from files already analyzed via compiler tracer events, indexes the rest of the stale files if any

If the index is missing but the project was already compiled, we will index from disk files instead: from .beam files for definitions since those are generally way more complete, and from the .ex sources for references.

---
This introduces two different compiler tracers(just like NextLS). This is because mix compilation happens in two main stages:
1. Deps compilation via `deps.loadpaths`
2. Actual compilation of the user code

This happens both in Expert and in regular `mix compile` runs, as `deps.loadpaths` is called internally by `mix compile`.
We need to both run a compiler tracer for dependencies, AND to discriminate between dependency code and user code to decide wether or not to index references. Having a DependencyTracer that is configured only during `deps.loadpaths` and a ProjectTracer that is configured afterwards allows us to do both things.

There is another quirk with `deps.loadpaths` as of elixir 1.20 where the compiler options will not get propagated to the compiler workers if `MIX_OS_DEPS_COMPILE_PARTITION_COUNT` is `>1`, fixed in elixir's main branch with commit `https://github.com/elixir-lang/elixir/commit/6f18bf70c07b118dc434b3fd478b420071bc487f` and not yet released in a stable version. To workaround this issue, that env var is forced to `1` for the duration of the compilation.

---
Later on for 0.4 we will want to investigate using the compiler locks to share the expert build with the user builds and save repeated work. That would be partially incompatible with compiler tracers because a user triggered build will not send the compiler tracer events to expert. In that case we want to also have the ability to index what's missing after compilation, and this PR should set the foundations for that.

I'd like to know how this perform in real codebases too, as I see a noticeable slowdown in cold compilation(mostly the `deps.loadpaths` step), but I don't know how impactful this is on large codebases.


doorgan:
One thing I just now realize I might want to explore here is to always index .beam files, and use tracers for references only. Scanning references is the most costly step of indexing from files and it can't be done with .beam files alone, but definitions are retrievable from .beam files.
This change slows down compilation mostly because buffering the module bytecode during compilation increased memory usage way too much in my tests, so we produce the index entries in the tracer which slows down compilation a lot.
But reference events are cheap to buffer, and the module bytecode is available in .beam files, so indexing definitions from disk and references from tracers might get us the best from both worlds.

katafrakt:
@doorgan not sure if I tested the right thing - I removed `.expert/build` and started the server from main, then removed `.expert/build` again and ran server from this branch. The second run was actually faster - 422s vs 379s.

doorgan:
@katafrakt I think the second run might have been faster because it didn't need to index the test files too, it only ran the compiler tracer

apps/engine/lib/engine/build/document/compilers/quoted.ex
```diff
@@ -131,6 +131,49 @@ defmodule Engine.Build.Document.Compilers.Quoted do
...
+  defp source_path(_path), do: nil
+
+  defp normalize_path(nil), do: nil
+
+  defp normalize_path(path) when is_binary(path) do
+    path = Path.expand(path)
+
+    if Forge.OS.windows?() do
+      String.downcase(path)
+    else
```
katafrakt: I assume this is to address Windows filesystem being case-insensitive, but MacOS is very often also case-insensitive. I'm wondering how important this step is. Perhaps we could just always downcase?

apps/engine/lib/engine/build/project.ex
```diff
@@ -140,24 +210,87 @@ defmodule Engine.Build.Project do
...
+    end)
+  end
+
+  defp with_dependency_compile_partitions_serialized(fun) when is_function(fun, 0) do
+    # As of Elixir 1.20, the compiler options are not propagated to the deps
+    # compiler partition workers, so we need to force the partitions to 1
+    # to ensure the DependencyTracer runs for every trace event.
+    # See https://github.com/elixir-lang/elixir/issues/15457
+    original = System.fetch_env(@dependency_compile_partition_env)
+    System.put_env(@dependency_compile_partition_env, "1")
```
katafrakt, line 246: This is the part that slows down code compilation you mentioned? Seeing that the linked issue is resolved, is it viable to "backport" is as a Future or we don't want that?

apps/engine/lib/engine/compilation/dependency_tracer.ex
```diff
@@ -0,0 +1,71 @@
...
+      beam_path when is_binary(beam_path) -> TraceBuffer.add_beam_path(file, beam_path)
+      _ -> :ok
+    end
+  end
+
+  defp beam_path(module) when is_atom(module) do
+    if Engine.Mix.loaded?() do
+      Path.join(Mix.Project.compile_path(), "#{Atom.to_string(module)}.beam")
+    end
+  rescue
```
katafrakt: What can fail here that we need to `rescue`? It seems to be a very straightforward operation.

katafrakt (approved):
I went over the code. Admittedly, I did not understand all of it, but overall looks ok to me. Let me know if there are any specific aspects I should test on a large project to asses the safety of this.

apps/engine/lib/engine/compilation/dependency_tracer.ex
doorgan: It's a leftover from debugging, I'll remove it

apps/engine/lib/engine/build/project.ex
doorgan: That's one reason yes. The other is that I decided to extract definitions from the module's bytcode inside the compiler tracer; the more work a tracer does the slower compilation becomes, so I fear this might cause a noticeable slowdown on large projects. For example the expert app compiles(with dependencies) in 18 seconds on a terminal, but 30 seconds with the expert tracers

apps/engine/lib/engine/build/document/compilers/quoted.ex
doorgan: It's to compare files are the same, I changed to something simpler

katafrakt:
I checked the compilation in terminal vs in Expert with these changes, as suggested. It's 361s in terminal vs 431s in Expert, so it is significantly slower, but not as much as with compiling Expert itself.

doorgan:
Still testing this, I found that references are incorrect, for .ex files they point to the top of the file instead of the line they occur at. I'll fix it.

doorgan:
I didn't catch any other outstanding issues, so I'll merge this now

PR #707, opened 2026-06-03
fix: only index sibling beams when origin source file changes

doorgan:
I found an issue where we would incorretly delete index entries created from .beam files, and also over-index files that did not change.

If you have a module like this:

```elixir
# parent.ex
defmodule Parent do
end
```

which produces `Elixir.Parent.beam` and update it like this:

```elixir
# parent.ex
defmodule Parent do
  defmodule Child do
  end
end
```

it would now also produce `Elixir.Parent.Child.beam`.

This should produce two entries, one for `Parent` definition and one for `Parent.Child` definition. When the indexer saw the latter, it would consider `parent.ex` to be stale, delete all entries for it in the index, and perform a full reindex of the .beam files, which is wasteful.

The fix in this PR is to check if entries from a .beam file shares source file with other .beam files, and only reindex those.

### DIFF
--- a/apps/engine/lib/engine/search/indexer/manifest.ex
+++ b/apps/engine/lib/engine/search/indexer/manifest.ex
@@ -237,7 +237,8 @@ defmodule Engine.Search.Indexer.Manifest do
     beam_paths_to_index =
       beam_paths_to_index
       |> include_known_beams_for_dirty_outputs(manifest, beam_paths, dirty_outputs)
-      |> include_new_beams(new_beam_paths, beam_paths)
+      |> Enum.concat(new_beam_paths)
+      |> Enum.uniq()
 
     %Plan{
       source_paths_to_index: Enum.uniq(source_paths_to_index ++ new_source_paths),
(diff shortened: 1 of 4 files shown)


apps/engine/test/engine/search/indexer/manifest_test.exs
```diff
@@ -0,0 +1,39 @@
...
+  alias Engine.Search.Indexer.Manifest
+  alias Engine.Search.Indexer.Manifest.Entry
+  alias Engine.Search.Indexer.Paths
+
+  @moduletag :tmp_dir
+
+  describe "plan/2" do
+    test "does not fan out from one new beam to all known beams", %{tmp_dir: tmp_dir} do
+      [known_beam_1, known_beam_2, known_beam_3, new_beam_path] = beam_paths(tmp_dir, 4)
+      beam_paths = [known_beam_1, known_beam_2, known_beam_3, new_beam_path]
```
katafrakt:
```suggestion
      beam_paths = [known_beam_1, known_beam_2, known_beam_3, new_beam_path] = beam_paths(tmp_dir, 4)
```

I think we can collapse it

apps/engine/lib/engine/search/indexer/manifest.ex
```diff
@@ -352,10 +352,10 @@ defmodule Engine.Search.Indexer.Manifest do
     Enum.uniq(beam_paths_to_index ++ dirty_beam_paths)
   end
 
-  defp include_new_beams(beam_paths_to_index, [], _beam_paths), do: Enum.uniq(beam_paths_to_index)
+  defp include_new_beams(beam_paths_to_index, []), do: Enum.uniq(beam_paths_to_index)
 
-  defp include_new_beams(_beam_paths_to_index, _new_beam_paths, beam_paths) do
-    MapSet.to_list(beam_paths)
+  defp include_new_beams(beam_paths_to_index, new_beam_paths) do
+    Enum.uniq(beam_paths_to_index ++ new_beam_paths)
```
katafrakt: We could now just have this one implementation. Not sure how meaningful optimization is to save this appending of an empty list.

katafrakt (approved):

PR #708, opened 2026-06-03
chore: bump to elixir 1.20 and otp 29

doorgan:
Elixir 1.20 was released today
We build the engine using the user's toolchain so we can bump elixir and erlang and it should not cause issues

It caught a lot of unused requires, type violations, and a few places where it had very subtle bugs though I'm not sure I've seen them manifested at runtime


.github/matrix.json
```diff
@@ -129,13 +147,13 @@
     {
       "os": "ubuntu-latest",
       "otp": "27.3.4.1",
-      "elixir": "1.18.4",
+      "elixir": "1.20.0",
       "project": "expert"
     },
     {
       "os": "windows-2022",
       "otp": "27.3.4.1",
```
katafrakt: BTW is this OTP version special in some way?

.github/matrix.json
katafrakt, line 1: I don't see OTP 29 added here TBH

.github/matrix.json
doorgan: No, it's the one that existed when lexical became expert

.github/matrix.json
doorgan: I updated the .tool-versions before flagging for review and forgot to update the ci matrix too, it's updated now

doorgan:
This pr made me realize that the flaky windows ci is not always solved by increased timeouts as I saw a run that failed the tests after a 30 seconds timeout then passed on rerun, it's very weird

apps/engine/lib/engine/build/project.ex
```diff
@@ -89,7 +89,7 @@ defmodule Engine.Build.Project do
     end
   end
 
-  defp maybe_load_modules do
+  def maybe_load_modules do
```
mhanberg, line 92: ?

apps/expert/lib/expert.ex
```diff
@@ -228,22 +228,8 @@ defmodule Expert do
   end
 
   def handle_notification(notification, lsp) do
-    with {:ok, handler} <- fetch_handler(notification),
```
mhanberg, line 231: Why was this removed?

apps/engine/lib/engine/build/project.ex
doorgan: I made it public to mock it with Patch in tests. I now realize we can also use `Patch.expose` in test instead
This function was added in #604 to improve load times in elixir >=1.19 but slows down tests significantly, so I decided to mock this to always return `:ok` since its return is always ignored anyways

apps/expert/lib/expert.ex
doorgan: `fetch_handler` does not define any handler for notifications and would always return `{:error, {:unhandled, request_module}}`. The type system now flags this so I inlined the error.

doorgan: To be more clear, this is the error, it's a about the context being `nil` for a handler that expects a context struct, but none of the handlers returned by `fetch_handler` is a notification handler so it doesn't make sense to have this block of code anyways

```
     warning: incompatible types given to Expert.Provider.Handlers.FindReferences.handle/2:

         handler.handle(notification, nil)
         #=> invoked as Expert.Provider.Handlers.FindReferences.handle/2

     given types:

         (
           dynamic(%GenLSP.Requests.TextDocumentReferences{
             params: %GenLSP.Structures.ReferenceParams{context: %{..., include_declaration: term()}}
           }),
           nil
         )

     but expected one of:

         (
           %GenLSP.Requests.TextDocumentReferences{
             params: %GenLSP.Structures.ReferenceParams{context: %{..., include_declaration: term()}}
           },
           %Expert.Document.Context{}
         )

     where "handler" was given the type:

         # type: dynamic(
           Expert.Provider.Handlers.CodeAction or Expert.Provider.Handlers.CodeFolding or
             Expert.Provider.Handlers.CodeLens or Expert.Provider.Handlers.Commands or
             Expert.Provider.Handlers.Completion or Expert.Provider.Handlers.DocumentSymbols or
             Expert.Provider.Handlers.FindReferences or Expert.Provider.Handlers.Formatting or
             Expert.Provider.Handlers.GoToDefinition or Expert.Provider.Handlers.Hover or
             Expert.Provider.Handlers.WorkspaceSymbol
         )
         # from: lib/expert.ex:231:25
         {:ok, handler} <- fetch_handler(notification)

     where "notification" was given the type:

         # type: dynamic(%{..., params: term()})
         # from: lib/expert.ex:232:30
         {:ok, notification} <- Expert.Protocol.Convert.to_native(notification)

     type warning found at:
     │
 233 │          {:ok, _response} <- handler.handle(notification, nil) do
     │                                      ~
     │
     └─ lib/expert.ex:233:38: Expert.handle_notification/2
```
Some handlers like `Expert.Provider.Handlers.Commands` do support `nil`

katafrakt:
Re: Windows:
Previous attempt to solve was disabling Microsoft Defender in the CI images and it often helps, but it looks like still sometimes something else is clogging the BEAM on Windows there.

katafrakt (approved):

Issue #727, opened 2026-06-15
Diagnostics error when using `Config.config_target/0`


LostKobrakai:
Reproduction:

- `mix new repro_target`
- Create `config/config.exs`
- Fill with 
```elixir
import Config

IO.inspect(config_target())
```

Diagnostics show this for `config_target()`
```
no :target key was given to this configuration file
```

while `mix compile` works just fine:

```sh
❯❯ echo $MIX_ENV

❯❯ echo $MIX_TARGET

❯❯ mix compile
:host
Compiling 1 file (.ex)
Generated repro_target app
```

katafrakt:
I think that was fixed in the linked PR

PR #731, opened 2026-06-18
fix(engine): set target when evaluating config diagnostics

SAY-5:
Diagnostics for config files are evaluated with `Config.Reader.eval!/3` but only `:env` is passed, so any config that calls `config_target/0` raises "no :target key was given to this configuration file" even though `mix compile` reports `:host` and compiles fine (#727).

Passing `target: Mix.target()` mirrors how mix resolves the target (defaulting to `:host`) and matches the existing `Mix.target()` usage in the indexer. Added a regression test covering a config file that uses `config_target/0`.


doorgan (approved):
The fix works, thank you! 💜

PR #736, opened 2026-06-19
fix(expert): extend list of editors having VSCode completion bug

katafrakt:
There is a bug in VSCode https://github.com/microsoft/vscode/issues/155738 that when an empty list is received by completion request, it never asks again for current word (they [claim](https://github.com/Microsoft/vscode/issues/13735) it's intentional). We had a workaround for that inherited from Lexical, but it only works for VSCode itself, not for any of its forks.

This is still a hack, just includes more editors. I believe the proper fix would be in our extension: either send in initialization options something like `"imBuggyVscode": true`, or even somehow intercept the empty completion responses and do something to trigger completion after next char (I haven't really explored the latter).


doorgan (approved):

doorgan:
> I believe the proper fix would be in our extension

Same and we can do it, but we should also contemplate that editors usually don't conform to spec so having the fix in expert itself is not a bad compromise

PR #740, opened 2026-06-23
chore(expert): use colon in filter text for Hex completion

katafrakt:
VSCode filters out the results returned by completion according to filter text. If it does not contain the colon in the beginning, all results get filtered out and instead of Hex packages, VSCode presents fuzzy-matched irrelevant atoms from the project.

This works differently in Emacs (and I assume in Vim, which was the primary target for the author of Hex intelligence feature), so it was missed during the review phase.

## Before
Screenshot 2026-06-23 at 10 57 39

## After
Screenshot 2026-06-23 at 11 27 59

Not sure if this should be a "fix" PR. It is a fix, but of an unreleased feature, so maybe it could be "chore" to have only one entry in the changelog? Anyway, this is why I'm not exactly a fan of autogenerated changelogs ;)

### DIFF
--- a/apps/expert/lib/expert/code_intelligence/completion/translations/hex_package.ex
+++ b/apps/expert/lib/expert/code_intelligence/completion/translations/hex_package.ex
@@ -15,17 +15,20 @@ defmodule Expert.CodeIntelligence.Completion.Translations.HexPackage do
         description: repo_label(package.repo)
       }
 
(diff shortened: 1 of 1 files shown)


doorgan (approved):

doorgan:
Maybe `chore` is alright? I agree this is not a changelog worthy fix

Issue #744, opened 2026-06-25
Expert errors on folding a phoenix render function


Terbium-135:
On clicking fold in VSCode at a `def render` function expert immediately stops:

```
2026-06-25 22:39:28.620 [info] Logger - error: 
2026-06-25 22:41:39.848 [info] [Error - 10:41:39 PM] Client Expert: connection to server is erroring.
Message header must separate key and value using ':'
{removed_failing_filter,logger_translator}
Shutting down server.
2026-06-25 22:41:41.848 [info] [Error - 10:41:41 PM] Stopping server timed out
2026-06-25 22:41:41.850 [info] [Error - 10:41:41 PM] Stopping server failed
2026-06-25 22:41:41.850 [info] Error: Stopping the server timed out
	at /home/elixir-dev/.vscode-server-insiders/extensions/expertlsp.expert-0.5.0/out/main.js:41:8889
```

This is a screenshot of what VSCode is showing after folding (and expert crashed)

<img width="331" height="130" alt="Image" src="https://github.com/user-attachments/assets/7cb8b71b-f6f2-4877-bb81-3131ef00c07b" />

Expert has been compiled with:
```
➜  expert git:(main) ✗ cat .tool-versions
zig    0.15.2
elixir 1.20.2-otp-29
erlang 29.0.2
```

Expert commit `de9e092`

katafrakt:
This is strange. On folding nothing is sent to the server. However I see VSCode sending documentSymbols right after folding, not sure what for.

The error looks like accidental write of non-JSON text to stdout, which the server tries to interpret as a JSON-RPC message. This happens when there are some debug statements left in the code, but I don't think we have these.

One thing that's also visible here is that the folding does not work correctly here - it should collapse the whole function body, but it leaves the last line of it. I'll look into this.

katafrakt:
Actually, it's probably not a stray `dbg` or `puts`, but actually a log crash - in such case Erlang indeed prints to stdio no matter what. So now we have to find why it fails. @Terbium-135 do you think there is anything that could help in the log lines preceding the ones you pasting? Could you also paste the VSCode extension output before the crash, so we see what was sent to server?

Terbium-135:
I'll get back to you asap.
Right now I am afk

katafrakt:
I cannot reproduce the bad folding of render too. In my case it's folding correctly in many scenarios I tried. Perhaps there is something specific inside this `render` function?

[...]

PR #746, opened 2026-06-29
chore(expert): serve foldingRange request even if the engine is initializing

katafrakt:
The reason for bad folding in #744 is that Expert does not serve `foldingRanges` requests until the engine is ready. This causes VSCode to fall back into the extensions regex-based folding, which is buggy (and likely non-fixable). However, `foldingRanges` request only needs an AST analysis, not the whole engine, so we can server it, even if the server is still starting.

There is still very brief moment when VSCode uses extension folding, then sends request and waits for response - during which the folding is wrong - but it gets fixed in under a second usually. Not sure if anything can be done about it. Perhaps we could disable extension-side folding completely.

### DIFF
--- a/apps/expert/lib/expert.ex
+++ b/apps/expert/lib/expert.ex
@@ -5,6 +5,7 @@ defmodule Expert do
   alias Expert.Project.Store
   alias Expert.Protocol.Convert
   alias Expert.Protocol.Id
+  alias Expert.Provider.Handler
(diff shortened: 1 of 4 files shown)


doorgan (approved):
If we already provide folding range here, it'd probably be good to remove the regex fallback from the extension

katafrakt:
We should wait for 0.2 release for that, but I agree.

PR #750, opened 2026-06-30
feat: add configuration to autofetch dependencies

### LINKED ISSUE #516: Automatically fetch dependencies on server start
Hi!

I am using `expert` inside Zed. We are using Renovate in the company, so almost every time we fetch master, there are some dependency updates.

This makes expert error with:

```
The Expert engine failed with errors on dependencies for project [project-name-here]. Would you like to fetch them?
```

Would it be possible to add a configuration flag where I can say "fetch dependencies automatically"?

doorgan:
Close #516 

Adds a new `autoFetchDependencies` configuration that skips the dependency fetching prompt and just runs `mix deps.get` for you.

If the autofetch fails, you get the retry prompt as usual.

### DIFF
--- a/apps/expert/lib/expert.ex
+++ b/apps/expert/lib/expert.ex
@@ -276,34 +276,41 @@ defmodule Expert do
   defp maybe_prompt_deps_fetch(lsp, project) do
     state = assigns(lsp).state
 
-    supports_show_message = Expert.Configuration.client_support(:show_message)
+    supports_show_message = show_message_supported?()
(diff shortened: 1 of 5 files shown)


doorgan:
Wondering if we should set this to true as default, I think that's what you want most of the time

katafrakt (approved):
Change looks good. Not sure about enabling by default fro security point of view, but perhaps ok (we do "worse" things already).

doorgan:
I changed the default to `true`, I'm not entirely sure the security concerns win over usability in this particular scenario but I do understand them

PR #752, opened 2026-07-01
chore: improve performance of SQLite indexer

katafrakt:
Apply performance improvements for SQLite indexer found in a large project. Details: https://github.com/expert-lsp/expert/pull/683#issuecomment-4855923028

- add missing indices
- skip `backend.reduce` completely
- select only needed columns in few places
- extract `entry_blobs` table to keep most heavy data for an entry - joined on demand now

### DIFF
--- a/apps/expert/lib/expert/search/store/backends/sqlite.ex
+++ b/apps/expert/lib/expert/search/store/backends/sqlite.ex
@@ -53,8 +53,8 @@ defmodule Expert.Search.Store.Backends.Sqlite do
   def destroy_all(%Project{} = project), do: project |> root_path() |> File.rm_rf!()
 
   @impl Backend
-  def reduce(%Project{} = project, acc, reducer_fun) do
-    GenServer.call(name(project), {:reduce, acc, reducer_fun}, :infinity)
+  def path_to_ids(%Project{} = project) do
+    GenServer.call(name(project), :path_to_ids, :infinity)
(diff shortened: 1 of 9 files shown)


apps/expert/lib/expert/search/store/backends/sqlite.ex
```diff
@@ -244,11 +252,26 @@ defmodule Expert.Search.Store.Backends.Sqlite do
     end)
   end
 
-  def do_reduce(%State{} = state, acc, reducer_fun) do
-    case query(state, "SELECT entry FROM entries ORDER BY id") do
+  def do_path_to_ids(%State{} = state) do
+    case query(state, "SELECT path, MAX(id) FROM entries WHERE id IS NOT NULL GROUP BY path") do
```
katafrakt, line 256: This could also be slow-ish in my tests before[^1], but I think extraction of `entry_blobs` made it fast again.

[^1]: Whenever I say "slow", it's 5+ seconds for my project on my machine

doorgan (approved):
💜

katafrakt:
@mhanberg I did `chore` instead of `perf`, similar to other PRs recently, to not duplicate changelog entries about a new feature. Do you think it should be `perf`?

mhanberg:
Sorry, I thought it was an oversight. 

I'll change it back since it was intentional.

PR #759, opened 2026-07-06
feat(engine): find definition for functions defined by macro

katafrakt:
This is especially useful with Ecto and jumping to functions like `MyApp.Repo.all`. Currently it takes the user to `use Ecto.Repo` inside `my_app/repo.ex`, but after this change the user actually lands in Ecto source.

The real changes are made in ElixirSense, this just updates the dependency and unskips the test verifying this behaviour.

* https://github.com/elixir-lsp/elixir_sense/pull/333
* https://github.com/elixir-lsp/elixir_sense/pull/339


katafrakt:
Looks like it broke some tests for completion. Probably ElixirSense changed something about it. I'll investigate.

katafrakt:
I solved it by pinning ElixirSense to an older commit, one that includes macro handling fixes, but does not include toxic2 changes.

@lukaszsamson want me to report the findings as issues in ElixirSense repo?

lukaszsamson:
Please do @katafrakt

doorgan (approved):
Comes from free by using ElixirSense so I'm good with this change.
FWIW #705 adds the same functionality by inspecting the bytecode, but this can be merged in the meantime before I finally merge that one

PR #760, opened 2026-07-07
fix: correctly pass Elixir source path config

katafrakt:
Before merging #568 I did a rename of the configuration key and renamed too much. As a result, ElixirSense was receiving `elixir_source_path` setting when is expected `elixir_src`.

This fixes that, also unskipping a test verifying jump to definition to Elixir source and fixes a bug with Zipper when a guard is present.

### DIFF
--- a/apps/expert/test/engine/code_intelligence/definition_test.exs
+++ b/apps/expert/test/engine/code_intelligence/definition_test.exs
@@ -67,6 +67,17 @@ defmodule Expert.Engine.CodeIntelligence.DefinitionTest do
     EngineApi.register_listener(project, self(), [:all])
     EngineApi.schedule_compile(project, true)
 
+    # point ElixirSense at the project node's Elixir source
+    elixir_src =
+      project
+      |> EngineApi.call(:code, :lib_dir, [:elixir])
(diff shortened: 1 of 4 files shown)


apps/expert/test/engine/code_intelligence/definition_test.exs
```diff
@@ -67,6 +67,17 @@ defmodule Expert.Engine.CodeIntelligence.DefinitionTest do
     EngineApi.register_listener(project, self(), [:all])
     EngineApi.schedule_compile(project, true)
 
+    # point ElixirSense at the project node's Elixir source
+    elixir_src =
+      project
+      |> EngineApi.call(:code, :lib_dir, [:elixir])
+      |> to_string()
+      |> Path.join("../..")
+      |> Path.expand()
```
katafrakt, line 76: I think we could perhaps use a similar trick to find the source directory when the config is not passed, making the experience more OOTB. I would have to test it separately though.

doorgan: If it works I'm all in for it!

doorgan (approved):

PR #761, opened 2026-07-07
feat: defer code action execution

Moosieus:
This pull request implements support for [`codeAction/resolve` requests](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#codeAction_resolve).

Prior to this pull request, Expert did not support `codeAction/resolve`, which meant all possible code actions (including refactors) were evaluated eagerly and sent to the client, even when a user did not expressly request them. This led to an outsized number of errors and excess processing that would effectively be discarded. Language server clients frequently make [`textDocument/codeAction` requests](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_codeAction) as users are editing, and the resolve request exists exactly to offload things accordingly.

## How it works

The server now advertises `resolveProvider` based on client capability. When a client declares it can resolve the edit property (`textDocument.codeAction.resolveSupport.properties` contains `"edit"`), Expert responds to `textDocument/codeAction` with a list of actions that carry only their title, kind, and an opaque data payload — not their edits. When the user actually invokes one, the client sends `codeAction/resolve` with that payload, and only then does Expert compute the edit for that single action.

Clients that don't support resolve fall back to the previous eager behavior where edits are computed inline, so no client regresses.

## Ancillary: RemovePipe crash fix

This addresses a specific crash I ran into, which begot all the above. Included here since it's what surfaced the problem, but it stands on its own.

## Implementation notes

- `Forge.CodeAction` owns the deferred payload's schema via `to_refactor_data/4` and `from_refactor_data/1`, so the wire format lives in one place rather than being spelled out at each producer and consumer.
- The code-action Handler behaviour is unified on `actions/4`, removing per-handler reflection over an optional callback.
- Formatter configuration is resolved once per request and shared across a request's refactorings, rather than re-read per action.

## Not in scope / follow-ups

- **Request -> document discovery:** Routing `codeAction/resolve` (whose document URI lives in its data payload rather than a textDocument field) currently adds small special-case clauses to the request-dispatch predicates. If more special cases accrete, the dispatch mechanism can be reconsidered.

### DIFF
--- a/apps/engine/lib/engine/code_action.ex
+++ b/apps/engine/lib/engine/code_action.ex
@@ -1,4 +1,11 @@
 defmodule Engine.CodeAction do
+  @moduledoc """
+  Handles `textDocument/codeAction` requests, which language server clients frequently send while
+  users are editing. Frequently computing all possible code actions can be expensive due to the
+  hot-path nature, so Language Server Protocol provides a mechanism to deouple the list of
+  possible actions from executing them. See `Expert.Configuration.client_resolves_code_action_edits?/1`.
+  """
+
   alias Engine.CodeAction.Handlers
   alias Forge.CodeAction.Diagnostic
   alias Forge.Document
@@ -15,21 +22,42 @@ defmodule Engine.CodeAction do
     Handlers.CreateUndefinedFunction
   ]
 
+  @doc """
+  Fans the request out to every handler whose advertised kinds and trigger kind match, and concats their actions.
+
+  ## Options
+
+    * `defer_edits?` - when `true`, handlers supporting deferral return actions without edits,
+      carrying a `data` payload for a later `codeAction/resolve` request (see `resolve_refactor/3`).
+      Handlers without deferral support ignore the option and return their edits inline. Defaults
+      to `false`.
+  """
   @spec for_range(
           Document.t(),
           Range.t(),
           [Diagnostic.t()],
           [Forge.CodeAction.code_action_kind()] | :all,
-          Forge.CodeAction.trigger_kind()
+          Forge.CodeAction.trigger_kind(),
+          keyword()
         ) :: [Forge.CodeAction.t()]
-  def for_range(%Document{} = doc, %Range{} = range, diagnostics, kinds, trigger_kind) do
+  def for_range(%Document{} = doc, %Range{} = range, diagnostics, kinds, trigger_kind, opts \\ []) do
     Enum.flat_map(@handlers, fn handler ->
       if handle_kinds?(handler, kinds) and handle_trigger_kind?(handler, trigger_kind) do
-        handler.actions(doc, range, diagnostics)
+        handler.actions(doc, range, diagnostics, opts)
       else
         []
       end
     end)
+  end
+
+  @doc """
+  Resolves the edits for a deferred refactor action, namely those listed by `Engine.CodeAction.Handlers.Refactorex`.
+  """
+  @spec resolve_refactor(Document.t(), Range.t(), String.t()) ::
+          {:ok, Forge.Document.Changes.t()} | :error
+
+  def resolve_refactor(%Document{} = doc, %Range{} = range, module_name) do
+    Handlers.Refactorex.resolve(doc, range, module_name)
   end
 
   defp handle_kinds?(_handler, :all), do: true
--- a/apps/engine/lib/engine/code_action.ex
+++ b/apps/engine/lib/engine/code_action.ex
@@ -1,9 +1,8 @@
 defmodule Engine.CodeAction do
   @moduledoc """
-  Handles `textDocument/codeAction` requests, which language server clients frequently send while
-  users are editing. Frequently computing all possible code actions can be expensive due to the
-  hot-path nature, so Language Server Protocol provides a mechanism to deouple the list of
-  possible actions from executing them. See `Expert.Configuration.client_resolves_code_action_edits?/1`.
+  Handles `textDocument/codeAction` requests.
+
+  Language clients frequently emit these while users are editing.
   """
 
   alias Engine.CodeAction.Handlers
--- a/apps/engine/lib/engine/code_action/handler.ex
+++ b/apps/engine/lib/engine/code_action/handler.ex
@@ -1,10 +1,32 @@
 defmodule Engine.CodeAction.Handler do
+  @moduledoc """
+  Behaviour for individual code action definitions, invoked in `Engine.CodeAction.for_range/6`.
+  """
+
   alias Forge.CodeAction
   alias Forge.CodeAction.Diagnostic
   alias Forge.Document
   alias Forge.Document.Range
 
-  @callback actions(Document.t(), Range.t(), [Diagnostic.t()]) :: [CodeAction.t()]
+  @doc """
+  Returns the handler's actions for the given document and range.
+
+  The diagnostics come from the request's context: those the client knows to overlap the range,
+  which diagnostic-driven handlers match against to offer fixes.
+
+  `opts` carries the request options (see the "Options" section of `Engine.CodeAction.for_range/6`).
+  Code actions that support deferred edits to `codeAction/resolve` honor `defer_edits?`
+  there; the rest ignore `opts` and always return their edits inline.
+  """
+  @callback actions(Document.t(), Range.t(), [Diagnostic.t()], keyword()) :: [CodeAction.t()]
+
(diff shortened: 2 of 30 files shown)


apps/engine/lib/engine/code_action/handlers/refactorex.ex
```diff
@@ -9,32 +9,100 @@ defmodule Engine.CodeAction.Handlers.Refactorex do
...
       _ -> []
     end
   end
 
+  @doc """
+  Computes the edits for a single refactoring previously listed with `defer_edits?: true`, as
+  part of the `codeAction/resolve` flow.
+
+  Availability is re-checked against the current document. Rewrite crashes intentionally surface
+  to the caller: they fail only this resolve request, and silently dropping them would hide bugs.
```
doorgan: I'm not sure this comment is useful, if there is no `rescue` we can always assume crashes will surface

apps/engine/lib/engine/code_action/handlers/refactorex.ex
```diff
@@ -9,32 +9,100 @@ defmodule Engine.CodeAction.Handlers.Refactorex do
...
+      )
+
+      []
+  catch
+    kind, reason ->
+      Logger.warning(
+        "Refactoring #{inspect(refactoring.module)} failed (#{kind}): #{inspect(reason)}"
+      )
+
+      []
```
doorgan: This is not useful. We'd only get thrown values upon process crashes and this is not the place to handle that.

apps/engine/lib/engine/code_action/handlers/refactorex.ex
```diff
@@ -9,32 +9,100 @@ defmodule Engine.CodeAction.Handlers.Refactorex do
...
+        ]
+
+      :error ->
+        []
+    end
+  rescue
+    error ->
+      Logger.warning(
+        "Refactoring #{inspect(refactoring.module)} failed: #{Exception.message(error)}"
+      )
```
doorgan: If we're logging the exception we should also get the stacktrace

apps/engine/lib/engine/code_action/handlers/refactorex.ex
```diff
@@ -47,19 +115,20 @@ defmodule Engine.CodeAction.Handlers.Refactorex do
     |> Sourceror.parse_string(line: start.line, column: start.character)
   end
 
-  defp ast_to_changes(doc, ast) do
+  defp sourceror_opts(doc) do
```
doorgan, line 108: This seems wrong, the function is renamed to `sourceror_opts` but it does not return opts, it returns changes

apps/engine/lib/engine/code_action/handler.ex
```diff
@@ -1,10 +1,32 @@
...
-  @callback actions(Document.t(), Range.t(), [Diagnostic.t()]) :: [CodeAction.t()]
+  @doc """
+  Returns the handler's actions for the given document and range.
+
+  The diagnostics come from the request's context: those the client knows to overlap the range,
+  which diagnostic-driven handlers match against to offer fixes.
+
+  `opts` carries the request options (see the "Options" section of `Engine.CodeAction.for_range/6`).
+  Code actions that support deferred edits to `codeAction/resolve` honor `defer_edits?`
+  there; the rest ignore `opts` and always return their edits inline.
```
doorgan: The docs update here are appreciated but slightly out of scope. It also reads too specific of the internals of the callers of this behavior, I don't think that explanation belongs here.

apps/engine/lib/engine/code_action.ex
```diff
@@ -1,4 +1,11 @@
 defmodule Engine.CodeAction do
+  @moduledoc """
+  Handles `textDocument/codeAction` requests, which language server clients frequently send while
+  users are editing. Frequently computing all possible code actions can be expensive due to the
+  hot-path nature, so Language Server Protocol provides a mechanism to deouple the list of
+  possible actions from executing them. See `Expert.Configuration.client_resolves_code_action_edits?/1`.
+  """
+
```
doorgan, line 7: I apologize if I'm wrongly assuming here, but this reads like the llm regurgitating the prompt or what it read at some point in the session, I don't think this explanation belongs here.

apps/engine/test/engine/code_action/handlers/refactorex_test.exs
```diff
@@ -135,6 +135,63 @@ defmodule Engine.CodeAction.Handlers.RefactorexTest do
...
+                 "module" => "Elixir." <> _,
+                 "uri" => "file:///file.ex",
+                 "version" => 7,
+                 "range" => %{"start" => %{"line" => _, "character" => _}, "end" => _}
+               } = action.data
+      end
+
+      for eager <- eager_actions do
+        deferred = Enum.find(deferred_actions, &(&1.title == eager.title))
+        assert deferred, "no deferred action for #{eager.title}"
```
doorgan, line 168: We generally do not use messages when asserts fail but I'm fine leaving it here

apps/expert/lib/expert/document/lookup.ex
```diff
@@ -282,5 +282,10 @@ defmodule Expert.Document.Lookup do
   defp extract_uri(%{text_document: %{uri: uri}}) when is_binary(uri), do: uri
   defp extract_uri(%{uri: uri}) when is_binary(uri), do: uri
   defp extract_uri(%{params: params}) when is_map(params), do: extract_uri(params)
+  # These code actions carry a document uri in their data payload rather than a text_document
+  # field (see the note in `Expert.document_request?/1`).
+  defp extract_uri(%{data: %{"provider" => "refactor", "uri" => uri}}) when is_binary(uri),
```
doorgan: I think this can be simplified to support other cases we may be missing
```suggestion
  defp extract_uri(%{data: %{"uri" => uri}}) when is_binary(uri),
```

apps/expert/lib/expert/provider/handlers/code_action_resolve.ex
```diff
@@ -0,0 +1,134 @@
...
+  @impl Expert.Provider.Handler
+  def handle(%CodeActionResolve{params: %CodeAction{} = action}, context) do
+    case resolve(action, context) do
+      {:ok, %CodeAction{} = resolved} -> {:ok, resolved}
+      {:error, reason} -> {:ok, error_response(reason)}
+    end
+  end
+
+  # One of our deferred refactor actions, resolved against the current document.
+  # The payload's field schema is owned by Forge.CodeAction.from_refactor_data/1.
```
doorgan: There's a lot of comments is this kind in the PR, I already flagged some, I think they should be removed.
They only re-state what is already implied either by the code or by its context, and are not useful.
Correct me if I'm wrong but it feels like most of this is the LLM explaining to you in the code itself what it just did, but they are not useful comments and should not be committed.

doorgan (commented):
Generally speaking the code looks good, I tested this and it seems to work as advertised
I have several comments, mostly around noisy comments which are probably byproduct/leftover of development.
The only comments I have some real concerns are about error handling, we should not be dropping stacktraces and we should not guard against process failures in code actions code.

apps/engine/lib/engine/code_action/handlers/refactorex.ex
Moosieus: I think this is just strange diff keying, `ast_to_changes/3` is defined below.

apps/engine/lib/engine/code_action.ex
Moosieus: This one was genuinely me. I wanted to call out the mechanism for people looking to jump in and make contributions. I've truncated it to this:
```
Handles `textDocument/codeAction` requests.

Language clients frequently emit these while users are editing.
```

I'll maintain it's prudent to call this out as somewhat of a hot path.

apps/expert/lib/expert/provider/handlers/code_action_resolve.ex
Moosieus: You're on point, Opus 4.8 specifically. I'd already trimmed out a lot of comments for the same reasons.

apps/engine/lib/engine/code_action/handlers/refactorex.ex
doorgan: Forever mad at github

doorgan (approved):
Thank you!

PR #764, opened 2026-07-07
feat: use compiler tracers for project indexing

doorgan:
This is a re-publishing of #705, which was reverted due to issues we found after merging. You can read that PR for the original description of how this works.

---

Adds the following fixes on top of the previous attempt:

- Uses `GenServer.cast` instead of `GenServer.call` in the `TraceBuffer`, this should make compilation faster by not serializing the work done with the tracer events. This was the original intent but I missed it in the original implementation, as the timing didn't noticeable change when testing it against expert itself.
- Stops tracing dependency compilation. The dependency tracer forced us to limit the dependency worker threads to 1, slowing down its compilation significantly. We already index dependency by reading the .beam files, which in my tests is way faster than a dependency tracer and a single worker thread, so I removed dependency tracing entirely
- fixes module attribute references. compiler tracers don't produce events for them, so I fixed that. indexing those would have required to read source files during compilation to extract them(very slow) or to do source indexing only for them(slow too). since references for a module attribute can only appear in the file they're defined in, I opted to move their resolution to the moment the user requests them, and extract them from the ast of the currently open document.


doorgan:
I think the issues are fixed now, in particular find references which was reported as not working:

For a module
Screenshot 2026-07-15 at 2 49 40 PM

For a public function
Screenshot 2026-07-15 at 2 50 29 PM

For private function
Screenshot 2026-07-15 at 2 51 20 PM

For module attributes
Screenshot 2026-07-15 at 2 50 53 PM

For variables
Screenshot 2026-07-15 at 2 51 50 PM

For erlang modules
Screenshot 2026-07-15 at 2 53 02 PM

For guards
Screenshot 2026-07-15 at 2 54 22 PM

For macro generated functions
Screenshot 2026-07-15 at 2 57 46 PM

For all the above, go to definition also works.

One reported issue was that references for modules and functions were not working. This new PR does not have commits that specifically fix those, so I'm not really sure what happened in that case and I can't reproduce the problem as you see above. I'd appreciate more eyes testing this!

apps/engine/lib/engine/build/project.ex
```diff
@@ -164,5 +250,27 @@ defmodule Engine.Build.Project do
...
+  defp log_progress(token, message) do
+    Progress.log_info(message)
+    Progress.report(token, message: message)
   end
+
+  defp timed(fun) when is_function(fun, 0) do
+    start = System.monotonic_time(:millisecond)
+    result = fun.()
+    {System.monotonic_time(:millisecond) - start, result}
+  end
```
scohen, line 272: isn't this just :timer.tc but with millis? Why not just use `timed_log` ?

scohen:
Running this branch, I get "mix deps.loadpaths" as progress during deps compilation, not the actual files being compiled.

scohen:
again, there's a dramatic slowdown during compilation of our largest file. It's still taking ages.

doorgan:
Closing in favor of #799

PR #766, opened 2026-07-08
chore: do bulk inserts for the `replace_all` callback of SQLite

doorgan:
Measured against the index for the expert app itself, it brings down `replace_all` time from ~4000ms down to ~550ms

This is a perf commit, marked as chore because it's for an unreleased feature. This adds an explicit column for entry_id to not rely on table metadata like rowid, hence the schema version bump. Even if 0.2 is not yet released, users of nightly build would need to know that they have to delete the indexes folder when updating if we don't bump it.

### DIFF
--- a/apps/expert/lib/expert/search/store/backends/sqlite.ex
+++ b/apps/expert/lib/expert/search/store/backends/sqlite.ex
@@ -5,14 +5,19 @@ defmodule Expert.Search.Store.Backends.Sqlite do
 
   alias Expert.EngineApi
   alias Expert.Search.Store.Backend
-  alias Exqlite.Sqlite3
   alias Forge.Project
   alias Forge.Search.Indexer.Entry
 
   require Entry
 
-  @schema_version 1
+  @schema_version 2
   @database_file "source.index.sqlite3"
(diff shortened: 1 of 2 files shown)


apps/expert/lib/expert/search/store/backends/sqlite.ex
```diff
@@ -5,14 +5,19 @@ defmodule Expert.Search.Store.Backends.Sqlite do
...
   alias Expert.EngineApi
   alias Expert.Search.Store.Backend
-  alias Exqlite.Sqlite3
   alias Forge.Project
   alias Forge.Search.Indexer.Entry
 
   require Entry
 
-  @schema_version 1
+  @schema_version 2
```
katafrakt, line 13: do we bump the schema, given it's still unreleased feature?

doorgan: Yes, for nightly users. Otherwise we have to instruct anyone giving it a try to delete the .expert/indexes folder
It's an internal counter so I figured the actual number doesn't matter, it's just a tool to force recreating the index

scohen (approved):

PR #768, opened 2026-07-09
fix(engine): preload deps and don't prune code paths

scohen:
I found a doozy of a bug when compiling a work project. For reasons, one of our erlang libraries is a fork of a well-known http client, but its application name isn't the same as the library's app name.

Another dep does checks for the client module being present, and if it's present, then reads the app config's vsn with the library's name. The problem is there is no app config, since the app name doesn't match the library name. This causes the build to fail, but only in expert.

The reason is that expert has previously loaded the app name when compiling other deps, and because module pruning was disabled, that old dep carried through and caused the compile issue. This does not happen on the command line, because when the project stack changes, the dep's modules are unloaded.

Expert originally disabled module pruning because it needs to have its own deps' modules available, but this can be accomplished by passing an argument to the project-level builds instead of always doing it when we need to be in a mix context.

### DIFF
--- a/apps/engine/lib/engine.ex
+++ b/apps/engine/lib/engine.ex
@@ -76,14 +76,33 @@ defmodule Engine do
   def ensure_apps_started(token \\ Progress.noop_token()) do
     apps_to_start = [:elixir, :runtime_tools | @allowed_apps]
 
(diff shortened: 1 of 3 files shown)


doorgan (approved):
Tested and it works

PR #769, opened 2026-07-09
perf: don't gate indexing on compliation completing

scohen:
When reindex was run at the same time as compilation, it caused something in the BEAM vm to soak up CPU. This was a permanent condition that would affect the project node until it was restarted.

Since we swapped out the ETS based index backend for a SQLite one, we should be able to reindex and compile at the same time. This will dramatically reduce the amount of time the project node takes to boot.

This change makes it so indexing and compilation occur at the same time, and we fix up the index after compilation completes

### DIFF
--- a/apps/engine/lib/engine.ex
+++ b/apps/engine/lib/engine.ex
@@ -8,6 +8,7 @@ defmodule Engine do
   alias Engine.Api.Proxy
   alias Engine.CodeAction
   alias Engine.CodeIntelligence
+  alias Engine.Commands
   alias Engine.Progress
   alias Forge.Project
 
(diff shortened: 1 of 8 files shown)


doorgan:
Maybe I'm missing something, I deleted .expert to force an initial compilation and indexing. Compilation starts, but indexing does not. When I use the "reindex" command I get a log for `project reindex requested` but it still waits for the compilation to finish

scohen:
@doorgan are you sure? When I run this with a clean build and index, I see progress for compilation and indexing happening at the same time, in the status bar, I see `Compiling foo.ex | Indexing 86%` Are you not seeing that?

doorgan:
Yes, here is a recording. After deleting .expert, it does a full compile, finishes, then kicks the indexing:

https://github.com/user-attachments/assets/613a81f9-aeb5-4ece-bbb9-57a25053a751

If I request reindexing while compilation happens it's the same, it waits for compilation to finish


https://github.com/user-attachments/assets/80ed2dd9-78e9-418d-9d18-80b34a8c7032

katafrakt:
I tested this too. On Expert repo, when open at root, I have seen something that looked like indexing parallel to compilation (but maybe it was for different apps), but on two other projects it was very clearly indexing after compilation finished.

scohen:
What's happening is that the new indexer needs the builds paths from mix, and that reintroduces the mix lock. Since the build starts before the indexer, the indexer needs to wait.

PR #774, opened 2026-07-09
fix: improve struct reference completion handling errors in struct completions

Moosieus:
* Add `reference_length/1` as an authority on what counts as a struct reference.
* Have `StructReference.detected?/2` and the engine's completion-length calculation both consume it so they agree.
* This stops `length/1` from being called on non-module `{:dot, ...}` payloads, and correctly admits `%__MODULE__.Foo`, `%__MODULE__.`, and `%@attr` — which the previous detection accepted but the length calc crashed on.

### DIFF
--- a/apps/engine/lib/engine/completion.ex
+++ b/apps/engine/lib/engine/completion.ex
@@ -134,18 +134,10 @@ defmodule Engine.Completion do
   end
 
   defp fetch_struct_completion_length(env) do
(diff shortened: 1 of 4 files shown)


doorgan (approved):

doorgan:
Thanks!
This needs rebasing, then we can merge it

PR #778, opened 2026-07-09
fix(nix): pass elixir_sense grammar and deps to expert in nix

### LINKED ISSUE #777: Project builds breaks with elixir_sense error in nix
Building a project with the `main` branch version of expert-lsp while installed through nix fails with the following:

```
10:48:59.865 instance_id=19F47C8540C [debug] Engine build output: Compiling 1 file (.yrl)

10:48:59.870 instance_id=19F47C8540C [debug] Engine build output: src/elixir_sense_parser.erl: permission denied

10:48:59.872 instance_id=19F47C8540C [debug] Engine build output: could not compile dependency :elixir_sense, "mix compile" failed. Errors may have been logged above. You can recompile this dependency with "mix deps.compile elixir_sense --force", update it with "mix deps.update elixir_sense" or clean it with "mix deps.clean elixir_sense"

10:48:59.881 instance_id=19F47C8540C [debug] sent notification server -> client window/logMessage

10:48:59.881 instance_id=19F47C8540C [error] Engine build script exited with status: 1
```

It's a matter of passing the `elixir_sense` files it needs to a `postInstallHook`

Alt-iOS:
Fixes #777 I have tested it in `x86_64 nixos` and `arm64 darwin`, simply passes some deps and grammars so that it may build `elixir_sense` properly

### DIFF
--- a/nix/expert.nix
+++ b/nix/expert.nix
@@ -2,8 +2,7 @@
   beamPackages,
   callPackages,
   lib,
-}:
-let
+}: let
   version = builtins.readFile ../version.txt;
 
   engineDeps = callPackages ../apps/engine/deps.nix {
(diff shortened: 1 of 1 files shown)


Alt-iOS:
Dug deeper and `beamPackages.buildMix` replaces the normal installPhase with Nixpkgs’ `beamModuleInstallHook`, which does not invoke `postInstall`. Consequently, the previous parser-generation override never ran. This moves generation to `postPatch`, before `beamCopySourceHook` copies the prepared dependency source into $out/src. Seemed better than the first attempt where the ElixirSense source got copied merely to modify it during postInstall

mhanberg:
I think this migh be fixed by updating the locked nixpkgs input according to https://github.com/expert-lsp/expert/pull/749/changes#r3513867439

It looks like that comment was not heeded and merged without bumping the lock. 

I have tested bumping nixpkgs and it seems to work.

Alt-iOS:
Ah, tested on darwin-arm64 and bumping does seem to fix it, I guess this would no longer be needed

PR #779, opened 2026-07-09
feat(expert): autodetect Elixir source

katafrakt:
In #760 we found a way to autodetect the Elixir source path for tests and figured it could be promoted as a feature - when there is no setting for `elixirSourcePath` we can derive it in Engine.

Only tested with `mise` installation, because I don't have any other configs at hand. But I think it degrades gracefully anyway. If the source is not detected, Expert will operate like it's not present (so as if `elixirSourcePath` is not passed).

### DIFF
--- a/apps/engine/lib/engine/code_intelligence/elixir_source.ex
+++ b/apps/engine/lib/engine/code_intelligence/elixir_source.ex
@@ -4,7 +4,7 @@ defmodule Engine.CodeIntelligence.ElixirSource do
   """
 
   # A stable file that every standard Elixir install ships under its root.
-  @sentinel Path.join("lib", "elixir/lib/kernel.ex")
+  @sentinel "lib/elixir/lib/kernel.ex"
 
   @doc """
   Returns the root of the Elixir source tree for the current installation, or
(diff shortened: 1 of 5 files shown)


apps/engine/lib/engine/code_intelligence/elixir_source.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Engine.CodeIntelligence.ElixirSource do
+  @moduledoc """
+  Autodetection of the Elixir standard library source path.
+  """
+
+  # A stable file that every standard Elixir install ships under its root.
+  @sentinel Path.join("lib", "elixir/lib/kernel.ex")
```
scohen: will this work in windows? I've always seen it
```elixir
Path.join(~w(lib elixir lib kernel.ex))
```

katafrakt: Good question. Given that the tests pass on Windows CI run, I'd assume it works, but I'll verify on actual Windows machine.

katafrakt: This is not a problem, it seems, because Elixir uses forward slashes on Windows too. This whole `Path.join` is not necessary at all here (I removed it).
However, the detection indeed does not work on Windows. I'm not sure why. Perhaps it's ElixirSense thing.

doorgan (approved):

apps/engine/lib/engine/code_intelligence/elixir_source.ex
katafrakt: I opened #792 to remember about this, but I'm going to merge it for now for other systems (I'm not sure this is an issue with autodetection or with passing Elixir source on Windows in general).

PR #780, opened 2026-07-09
ci: add check for Nix dependencies

katafrakt:
Happened to me more than once that I forgot to run `mix deps.nix` (recently on #759), so I figured we might want to have a CI check for that.

### DIFF
--- a/.github/workflows/ci-nix.yaml
+++ b/.github/workflows/ci-nix.yaml
@@ -53,15 +53,7 @@ jobs:
         uses: DeterminateSystems/magic-nix-cache-action@908b263ff629f4cc17666315b7fd3ec127c6244d # v14
 
       - name: Regenerate deps.nix
-        run: |
-          nix develop --command bash -c '
-            set -euo pipefail
-            mix local.hex --force
-            mix local.rebar --force
-            for app in expert engine forge expert_credo; do
-              (cd "apps/$app" && mix deps.get && mix deps.nix)
-            done
-          '
+        run: nix run .#update-deps
 
       - name: Ensure deps.nix is up to date
         run: |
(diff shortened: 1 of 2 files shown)


nix/expert.nix
```diff
@@ -45,7 +45,25 @@ beamPackages.mixRelease rec {
...
+    # This lives here, in deps.nix's `overrides` hook, rather than in deps.nix itself
+    # so it survives `mix deps.nix` regenerating that file.
+    overrides = _final: prev: {
+      exqlite = prev.exqlite.overrideAttrs (old: {
+        preConfigure = (old.preConfigure or "") + ''
+          export ELIXIR_MAKE_CACHE_DIR="$TMPDIR/elixir_make"
+        '';
+      });
+    };
+  };
```
katafrakt: This is pure Claude. I don't know nearly enough about Nix to tell if it's good or not, but it fixes the CI failure. Would appreciate thorough review here.

mhanberg: Once we merge #791 we can remove this part and the changes to the various deps.nix files and this PR can just be the ci change.

katafrakt: I like CI-only change much more! It's rebased and this change is removed now.

.github/workflows/ci-nix.yaml
```diff
@@ -35,3 +35,37 @@ jobs:
...
+      - name: Regenerate deps.nix
+        run: |
+          nix develop --command bash -c '
+            set -euo pipefail
+            mix local.hex --force
+            mix local.rebar --force
+            for app in expert engine forge expert_credo; do
+              (cd "apps/$app" && mix deps.get && mix deps.nix)
+            done
+          '
```
mhanberg: I believe this can just be `nix run .#update-deps`

It executes the "update-deps" "app" in the nix flake at https://github.com/expert-lsp/expert/blob/7b93e0be6eb10d9faa64aba769e854025c5f20e2/flake.nix#L34-L53

katafrakt: Nice, updated

mhanberg (approved):

PR #794, opened 2026-07-15
fix: exclude debug applications with MIX_ENV=prod

### LINKED ISSUE #786: :wx`/`:observer` in the release pull wxWidgets (GTK/X11) into Linux packages intended?
Hi! I package Expert for Homebrew. On **Linux**, the build ends up requiring **wxWidgets** and a big desktop-GUI stack. That's surprising for a headless LSP that never shows a window.

The cause is here in `apps/expert/mix.exs`:

```elixir
extra_applications: [:logger, :runtime_tools, :kernel, :wx, :observer, :telemetry],
```

`:observer` is Erlang's graphical monitor and it needs `:wx`, which links the wxWidgets GUI library. Because the release bundles the Erlang runtime, those GUI pieces ship in every build — so on Linux they drag in the whole GTK/X11 stack.

I couldn't find `:wx`/`:observer` actually used at runtime (they only appear in dev files like `.iex.exs`), so I assume they're there for local debugging.

**Question:** could they be left out of release builds, so packaged installs don't need the GUI stack? For example,

```elixir
# only add the GUI/debug apps outside of production
extra_applications: [:logger, :runtime_tools, :kernel, :telemetry] ++
                      if(Mix.env() == :prod, do: [], else: [:wx, :observer]),
```

Can you take a look this?
Thanks!

AlternateRT:
Fixes #786

### DIFF
--- a/apps/expert/mix.exs
+++ b/apps/expert/mix.exs
@@ -24,17 +24,27 @@ defmodule Expert.MixProject do
 
   def application do
     [
-      extra_applications: [
-        :logger,
-        :runtime_tools,
-        :kernel,
-        :wx,
-        :observer,
-        :telemetry,
-        :hex_core
-      ],
+      extra_applications: extra_applications() ++ debug_applications(Mix.env()),
       mod: {Expert.Application, []}
     ]
+  end
+
+  def extra_applications do
+    [
+      :logger,
+      :runtime_tools,
+      :kernel,
+      :telemetry,
+      :hex_core
+    ]
+  end
+
+  def debug_applications(:prod) do
+    []
+  end
+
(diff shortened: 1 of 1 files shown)


doorgan (approved):

AlternateRT:
@doorgan Could we backport this? I noticed you added the label for it, but nothing ever happened.

Anyways, even if it had, I don't think the bot would have succeeded in cherry-picking it - there would be a merge conflict due to the `:hex_core` dependency so it needs to be done manually. I could submit another PR targeting the v0.1 branch if that's ok.

doorgan:
@AlternateRT done!

PR #799, opened 2026-07-17
feat: use compiler tracers to index definitions

### LINKED ISSUE #356: Compiler Tracers

doorgan:
Alternative to #764 in the hopes we'll get the same benefits from tracers without paying as high of a price on big projects with large files

The idea is rather simple:
- We are already able to index most things, including references, and indexing times are not *terrible*
- We are not able to index code generated by macros, so if module A has functions defined by macros and module B calls those functions from A, it is very likely that we won't be able to index references for the B functions, nor be able to find them with workspace symbols
- Compiler tracers see the module bytecode after compilation, so we are able to extract definitions from there
- If we use the compiler tracers to complete what we index by reading source files, we get a more complete index with relatively low effort

The way this is implemented is via a new function in the existing Tracer, and a TraceBuffer(same idea as #764) that receives and processes the tracer events to not slow down compilation. The results of this are then added to the index.

This PR does not remove the locks that serialize compilation and indexing, so right now they don't run in parallel, but changes here were made assuming that they can run in parallel. If they do, then we are open to race conditions where the tracer or the index may overwrite the results of one another. To prevent, we wait for the indexer to finish if it's still running before adding the extra results, that is we try to write the results in the order source indexer -> compiler tracer.

---

The previous PR is way larger and does a lot more work than this, but ultimately I think its benefits do not outweight the costs. One particular issue is references: compiler tracers make it trivial to retrieve function, struct and module references, but they come as one event per reference. Those events don't hold as much information as we need for out index, so we need to read the source file anyways to refine the entry ranges. If we go to disk once per reference, this is easily an n^2 problem and compilation runs slow as molasses for huge files. If we want to avoid this we need a lot of ceremony to buffer references so we can perform the range analysis, but at this point this is almost the same as the source indexer, so it's not really a benefit. That is the main reason we're exploring this alternative, simpler approach.

### DIFF
--- a/apps/engine/lib/engine/compilation/tracer.ex
+++ b/apps/engine/lib/engine/compilation/tracer.ex
@@ -2,10 +2,12 @@ defmodule Engine.Compilation.Tracer do
   import Forge.EngineApi.Messages
 
   alias Engine.Build
+  alias Engine.Compilation.TraceBuffer
   alias Engine.Module.Loader
   alias Engine.Progress
 
   def trace({:on_module, module_binary, _filename}, %Macro.Env{} = env) do
+    maybe_record_module(env.file, module_binary)
     message = extract_module_updated(env.module, module_binary, env.file)
     maybe_report_progress(env.file)
     Engine.broadcast(message)
@@ -48,6 +50,13 @@ defmodule Engine.Compilation.Tracer do
       Progress.report(token, message: progress_message(file))
     end
   end
+
+  defp maybe_record_module(file, module_binary)
+       when is_binary(file) and is_binary(module_binary) do
+    TraceBuffer.record_module(file, module_binary)
+  end
+
+  defp maybe_record_module(_file, _module_binary), do: :ok
 
   defp progress_message(file) do
     relative_path_elements =
--- a/apps/engine/lib/engine/search/indexer/beams.ex
+++ b/apps/engine/lib/engine/search/indexer/beams.ex
@@ -736,9 +736,9 @@ defmodule Engine.Search.Indexer.Beams do
   end
 
   defp position(line, column, context_line) do
-    # NOTE(doorgan): We need to populate the `document_line_count` otherwise the
-    # will end up pointing to the top of the file. We are faking it here instead
-    # of actually reading the file for performance, and it seems to work
+    # NOTE(doorgan): We need to populate `document_line_count` otherwise the LSP
+    # range will end up pointing to the top of the file. We are faking it here
+    # instead of actually reading the file for performance, and it seems to work
     # well despite being a hack.
     %Position{
       line: line,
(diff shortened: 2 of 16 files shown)


apps/engine/lib/engine/search/indexer/beams.ex
```diff
@@ -693,12 +726,34 @@ defmodule Engine.Search.Indexer.Beams do
...
     Range.new(
-      %Position{line: line, character: column, starting_index: 1},
-      %Position{line: line, character: column + max(length, 1), starting_index: 1}
+      position(line, column, context_line),
+      position(line, end_column, context_line)
     )
   end
 
+  defp position(line, column, context_line) do
+    # NOTE(doorgan): We need to populate the `document_line_count` otherwise the
```
katafrakt: A word likely missing here at the end: "otherwise the [what]"

apps/expert/lib/expert/project/indexer.ex
```diff
@@ -124,22 +128,92 @@ defmodule Expert.Project.Indexer do
...
+        Logger.warning("Could not drain compiler trace index supplement: #{inspect(reason)}")
+        %{}
+    end
+  end
+
+  defp apply_trace_definitions(_project, trace_batch) when map_size(trace_batch) == 0, do: :ok
+
+  defp apply_trace_definitions(%Project{} = project, trace_batch) when is_map(trace_batch) do
+    with {:ok, all_entries} <- Search.Store.all(project, []) do
+      current_entries = Enum.filter(all_entries, &Map.has_key?(trace_batch, &1.path))
```
katafrakt: If I'm not mistaken, this loads all the entires to memory and then filters it. Perhaps we should let SQLite do that instead? It seems heavy for large projects.

apps/engine/lib/engine/compilation/tracer.ex
```diff
@@ -49,6 +51,13 @@ defmodule Engine.Compilation.Tracer do
     end
   end
 
+  defp maybe_record_module(file, module_binary)
+       when is_binary(file) and is_binary(module_binary) do
+    TraceBuffer.record_module(file, module_binary)
+  end
```
katafrakt, line 57: I wonder if for perf reasons it would not be better to parse before storing? I know it breaks the separation of concerns a bit, but storing all the binaries on a large project and clean compilation seems expensive. Any thoughts?

katafrakt (commented):
Looks good. I tested it on our project and it works fine. The compilation times look normal, so no big additional penalty. I just have few comments/quuestions.

apps/engine/lib/engine/compilation/tracer.ex
doorgan: If it's a concern with large projects we can do that, I'm a bit wary though because tracers can slow down compilation *a lot*

katafrakt: Sure, let's just keep that in mind then, if it actually shows up as a real problem in a large project. I'll keep an eye after this is merged.

PR #800, opened 2026-07-17
fix: report progress after a file is indexed

doorgan:
We were reporting progress when we *started* indexing a file, not afterwards. This would sometimes result in progress getting stuck at 100% for a while if a file was particularly slow to index.


doorgan:
Superseded by #803

PR #804, opened 2026-07-20
fix: improve performance of formatting

scohen:
When formatting code, there are a couple of potentially expensive operations:
  1. Compiling any plugins
  2. Fetching the formatter that pertains to the given file

Prior to this change, we'd use a lock to prevent formatting while the project stack might change. I talked to Jose about this, and he said that once plugins are compiled, we shouldn't need to worry about the project stack. Because of this, I tightened the lock to only happen when plugins are loaded.

I also added a cache around discovering and refetching formatters. This works by following the subdirectories that each .formatter.exs specifies, and monitoring each .formatter.exs for changes.

During my testing, I lowered formatting from 400ms to 10ms on some larger files, and completely removed any locks that would cause multi-second formatting times.

### DIFF
--- a/apps/engine/lib/engine/code_mod/format/cache.ex
+++ b/apps/engine/lib/engine/code_mod/format/cache.ex
@@ -281,13 +281,18 @@ defmodule Engine.CodeMod.Format.Cache do
 
   @spec discover_dot_formatters(Project.t()) :: State.dot_formatters()
   defp discover_dot_formatters(%Project{} = project) do
-    project
-    |> Project.root_path()
-    |> discover_dot_formatters()
+    case Project.root_path(project) do
+      nil ->
+        %{}
+
+      root_path ->
+        discover_dot_formatters_at_path(root_path, %{})
+    end
   end
 
-  @spec discover_dot_formatters(Path.t(), State.dot_formatters()) :: State.dot_formatters()
-  defp discover_dot_formatters(dir, dot_formatters \\ %{}) do
+  @spec discover_dot_formatters_at_path(Path.t(), State.dot_formatters()) ::
+          State.dot_formatters()
+  defp discover_dot_formatters_at_path(dir, dot_formatters) do
     formatter_exs = Path.join(dir, ".formatter.exs")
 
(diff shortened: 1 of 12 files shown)


scohen:
@doorgan I've been running this branch for a couple of weeks, it seems to work and has improved formatting perf fairly significantly, though it does change how formatting generates edits, which is a large change.

I think we should go with the new way that edits are generated, because even when I had a heuristic and emitted a whole document replace, emacs had a pathological case where a large set of edits would cause _it_ to calculate an edit doc on its end, and it would never finish. Changing to this strategy fixes that, since it creates edits per line, rather than per document.

apps/engine/lib/engine/code_mod/format/cache.ex
```diff
@@ -0,0 +1,343 @@
...
+    entry = Entry.new(formatter, opts, file_path)
+
+    true = :ets.insert(@entries_table, entry_row(id: entry.id, entry: entry))
+    true = :ets.insert(@paths_table, path_row(path: file_path, entry_id: entry.id))
+
+    {:ok, entry}
+  rescue
+    exception ->
+      Logger.warning(
+        "Could not resolve formatter for #{file_path}: #{Exception.message(exception)}"
```
doorgan: nit: we probably want to report the stacktrace too?

doorgan (commented):
I did some testing and it works. The new algorithm looks good and makes sense to me. The reasoning about applying line-by-line, in reverse order does as well, it's basically how I approached patching in Sourceror too.

A bunch of tests need updating though. Do you want me to update them?

apps/engine/lib/engine/code_mod/format/cache.ex
scohen: Done

scohen:
> Do you want me to update them?

Nope, I've got it.

scohen:
FWIW, a lot of those tests focused on same line edits, which won't happen any more, since we replace the entire line. I've removed those.

doorgan (approved):

doorgan (approved):

PR #806, opened 2026-07-20
chore: remove ETS backend

doorgan:
We've been testing the new SQLite backend for a while and I think it's safe to remove the old ETS one now


scohen (approved):
👋

Moosieus:
Closes https://github.com/expert-lsp/expert/issues/753?

PR #807, opened 2026-07-21
fix: do not compile mix.exs as document

### LINKED ISSUE #694: State becomes stale after minor modifications, requiring frequent manual restarts
### Environment
- **Editor:** Zed
- **Editor Version:** Zed v1.3.5 (2026-05-20)
- **OS:** macOS (arm)
- **Language Server:** Expert
- **Elixir & OTP Version:** Elixir 1.19.5-otp-28, OTP 28

### Steps to reproduce
1. Open an Elixir project in Zed.
2. Introduce a minor code change or fix an existing compile error/warning.
3. Save the file.
4. Observe that the editor diagnostics (errors/warnings) do not update, or display inline errors on lines that are now correct or blank.
5. Execute `editor: restart language server`. The diagnostics instantly clear and reflect the true state of the code.

### Expected behavior
The LSP should automatically recompile or track project changes incrementally on save, updating the diagnostics dynamically without requiring manual server restarts.

### Actual behavior
Diagnostics become completely detached from the current file state after a few minor edits, leaving the editor in a stale state.

https://github.com/zed-extensions/elixir/issues/132#issuecomment-4511606765

doorgan:
Fix #694, based on the reproduction provided in [this comment](https://github.com/expert-lsp/expert/issues/694#issuecomment-4928676147)

Editing mix.exs with parseable but invalid code could delete the active `MixProject` module, causing subsequent compilations and diagnostics to remain broken until Expert restarted. 
We were also comparing the document's *absolute* path to the literal `"mix.exs"` which would never match, as that is not an absolute path, which caused the comparison to always be `false` and the `mix.exs` file would get compiled, causing the `nil` stack crash from that issue.

I may note that this fix and the features we're relying on required me to dig a bit deep into the Mix code to understand, which I generally don't like because it means we're relying on internal behavior, but I don't currently have a better solution in mind.

### DIFF
--- a/apps/engine/lib/engine/application.ex
+++ b/apps/engine/lib/engine/application.ex
@@ -18,6 +18,7 @@ defmodule Engine.Application do
           Engine.Compilation.TraceBuffer,
           Engine.Dispatch,
           Engine.ModuleMappings,
+          {Task.Supervisor, name: Engine.TaskSupervisor},
           Engine.Build,
           Engine.ModuleStore,
           Engine.Build.CaptureServer
--- a/apps/engine/lib/engine/bootstrap.ex
+++ b/apps/engine/lib/engine/bootstrap.ex
@@ -6,6 +6,9 @@ defmodule Engine.Bootstrap do
   the project's code paths, which are then added to the code paths from the language server. At this
   point, it's safe to start the project, as we should have all the code present to compile the system.
   """
+  alias Engine.Build
+  alias Engine.Build.Isolation
+  alias Forge.Document
   alias Forge.LogFilter
   alias Forge.Project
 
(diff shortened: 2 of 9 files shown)


scohen:
will we still get diagnostics out of the mix.exs?

doorgan:
We would only get parsing error diagnostics but not diagnostics like "undefined function" which we get from the compiler.
I added a new commit that isolates the compilation of the mix.exs file only so we can retrieve the compiler diagnostics too

katafrakt:
Some comments from testing:

1. In umbrella project, I'm adding a call to undefined `foo()` function inside `def project()`. In root `mix.exs`, the diagnostics show it as an issue. In `apps/app1/mix.exs` the same code does not show any diagnostic error.

2. When I put `exit(1)` in `mix.exs` and boot Expert with this line already present, I'm getting a stacktrace in the diagnostic message - not sure if this is intended. Does not happen when I add exit with Expert already running.
Screenshot 2026-08-21 at 18 27 04

3. Compiling `mix.exs` on our large project after a save which changes nothing takes about 3 seconds, I think because of `modules_loaded_from`. Not sure if this is a problem and can be somewhat improves, just noting that,

doorgan:
@katafrakt I fixed those issues.

I had to change expert's behavior though. On main we compile the mix.exs file on type, so you get compiler diagnostics before you save the file. There is a *lot* of code that's not straightforward at all to support this, and even then it's very finicky and eventually messes up the project stack. There's also a limitation in which we always run with the project config we saw when the engine started, and never update it after changes to mix.exs.

I now copied part of the ElixirLS implementation where we always use the latest project config, but we do have the tradeoff of only being able to provide syntax diagnostics as you type, and compiler diagnostics after you save. TBH I think this is a fair tradeoff. The only other way I came up to compile the mix.exs file as you type and not run into project stack issues is to compile the mix.exs file in a separate node which is IMO a lot for just this(it's basically yet another little engine inside the engine...) and adds significant delay to diagnostics.

PR #820, opened 2026-07-28
fix: incorrect undefined protocol function diagnostic

### LINKED ISSUE #296: MFA marked as undefined with a warning when it is defined and the same MFA is suggested as a solution
On a fairly large project (170K lines of Elixir code), I've been experimenting with adding Expert in order to upgrade us to the latest Elixir versions.

After adding Expert and forcing a full compile and build of the project, deps, and indexes, I am receiving hundreds of the following error (MFAs are anonymized since this is in our production codebase):

```elixir
MyApp.MyProtocol.function/1 is undefined (module MyApp.MyProtocol is not available or is yet to be defined). Did you mean:

      * MyApp.MyProtocol.function/1
```

More information to assist in debugging:
- Elixir version 1.19.5
- Erlang version 28.3
- `MyApp.MyProtocol` is aliased at the top of the file (`alias MyApp.MyProtocol`) so the code that is causing the warning is `MyProtocol.function(xxx)`
- When I force a save within this file, it removes the warning, however if I move to a different file and then force a save in that file, the warning comes back
- I have `consolidate_protocols: Mix.env() not in [:dev, :test]` set in the mix.exs.

doorgan:
Fix #296 

I spent several weeks trying to reproduce this thinking that it was an issue with Elixir's 1.19 parallel compilation, but it turned out to be something way simpler.  The docs for `:code.load_binary` say this:

> `Filename` is only used by the code server to keep a record of from which file the object code for `Module` originates

We interpreted `Filename` to mean the source file, but it must be a .beam file: "from which file the object code from `Module` originates". Due to this confusion we were passing the .ex file path to the code server, then Elixir introduced [this commit](https://github.com/elixir-lang/elixir/commit/5f50a1f9075ccfadd0ccdc37f2328e146660b4e8#diff-5fc27e300aa733026cf10db92ae6a9a753a73e2466fa7569a28755e87443c1a8R78) in 1.19 which calls `:code.which` on that path. Elixir would then try to read the beam bytecode from that path, get nothing useful back and return `{:not_found, nil}` for that module in the parallel checker, the compiler thought the module/functions did not exist, and produced the spurious diagnostics from linked issue.

The fix is to just not pass any path there and just provide `[]`, that is a safe value for the code server meaning the bytecode does not come from a file, and the parallel checker already considers this a legal scenario and handles it appropriately.

### DIFF
--- a/apps/engine/lib/engine/compilation/tracer.ex
+++ b/apps/engine/lib/engine/compilation/tracer.ex
@@ -18,12 +18,7 @@ defmodule Engine.Compilation.Tracer do
 
   def extract_module_updated(module, module_binary, filename) do
     if !Loader.ensure_loaded?(module) do
-      erlang_filename =
-        filename
-        |> ensure_filename()
-        |> String.to_charlist()
-
-      :code.load_binary(module, erlang_filename, module_binary)
+      :code.load_binary(module, [], module_binary)
     end
 
     functions = module.__info__(:functions)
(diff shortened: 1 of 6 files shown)


katafrakt (approved):
Great find! And indeed, the message disappeared from logs in my tests.

PR #824, opened 2026-07-29
fix: stop calls to `IO` and other writes to stdout from crashing Expert.

### LINKED ISSUE #428: Correctly shutdown in case we shut down the buffer

Moosieus:
*Continued from: https://github.com/expert-lsp/expert/pull/808*

## The Problem

Expert uses stdio as the communication bus for LSP as is customary for language servers. A problem is that any writes to stdout that don't strictly follow the protocol will break it. This error case manifests as an explicit server shutdown in the best case scenario (client shuts down the server), and in the worst case, obfuscated errors like `Header must provide a Content-Length property`.

This problem persists today in Expert where the manager node makes (e)rpc calls to the project node, which redirects the project node's standard output to the manager node, which can corrupt the LSP connection. This problem is ephemeral and hard to replicate, and ~~I suspect~~ is the cause of several issues, including but not limited to https://github.com/expert-lsp/expert/issues/366#issuecomment-4349250065, https://github.com/expert-lsp/expert/issues/539, https://github.com/expert-lsp/expert/issues/744, and https://github.com/expert-lsp/expert/issues/781.

## The Solution
OTP has an undocumented but long supported `-user` flag  for providing an alternative `:user` [I/O device](https://www.erlang.org/doc/apps/stdlib/io_protocol.html) at boot. [iex notably uses this](https://github.com/elixir-lang/elixir/blob/a0b17a355a77a87884be5df4052f49b76ab8031f/bin/iex#L38), so it's about as well supported as an undocumented API can get.

We can provide our own user device that redirect nominal `:stdio` writes to `:standard_error` while giving our GenLSP adapter exclusive usage of `:standard_out`.


doorgan (commented):
How's this been working for you? I have tested it and it works as expected, the code looks good to me too. There's apparently some issues on windows though(the failing windows tests in CI) because erlang converts CRLF to LF: https://github.com/erlang/otp/issues/10258

Moosieus:
It's been working for me without issue with no more stdout leaks, marking this as ready for review.

doorgan (approved):
Thanks! I tested it and fixed the windows issues, it works well there too

PR #833, opened 2026-07-31
chore(sqlite): only store what's needed in entry_blobs table

katafrakt:
One more proposal to improve SQLite indexer a bit. Currently `entry_blobs` contains just a serialized `Entry` struct, which means a lot of duplication and table taking a lot more space than necessary. This introduces serialization to specialized entry blob shape and deserializing `Entry` from it.

On our work project:

**Before:**

```
*** Table ENTRY_BLOBS *********************************************************

Percentage of total database......................  43.6%    
Number of entries................................. 1753557   
Bytes of storage consumed......................... 805842944 
Bytes of payload.................................. 749056083   93.0% 
Bytes of metadata................................. 15598615     1.9% 
```

**After:**

```
*** Table ENTRY_BLOBS *********************************************************

Percentage of total database......................  30.7%    
Number of entries................................. 1753557   
Bytes of storage consumed......................... 461762560 
Bytes of payload.................................. 433125631   93.8% 
Bytes of metadata................................. 14172401     3.1% 
```

This is more than 40% down for `entry_blobs`, saving about 300M of disk space in the database file.

Thoughts?

### DIFF
--- a/apps/expert/lib/expert/search/store/backends/sqlite.ex
+++ b/apps/expert/lib/expert/search/store/backends/sqlite.ex
@@ -11,7 +11,7 @@ defmodule Expert.Search.Store.Backends.Sqlite do
   require Entry
   require Logger
 
-  @schema_version 3
+  @schema_version 4
   @database_file "source.index.sqlite3"
   @slow_query_threshold_ms 500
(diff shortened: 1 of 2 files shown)


doorgan (approved):
Good one! Any change that can reduce disk usage is a win, indexes get really big really fast

PR #834, opened 2026-08-03
fix(engine): resolve formatter against project mix config

kevinschweikert:
Formatter resolution now runs with the project pushed onto the Mix project stack via `Engine.Mix.with_project_config/2`, so a dependency's `.formatter.exs` can read `Mix.Project.config/0` and `:import_deps` resolves instead of silently falling back to the root `.formatter.exs`; `loadpaths` and `ensure_hex_and_rebar` are gone from that path, since `:import_deps` needs dependency paths rather than compiled dependencies. Because Mix expands `:deps_path` and `path:` dependencies against the working directory, those paths can only be read correctly inside `in_project/2`, so `Engine.Mix` records them there and `Engine.Mix.deps_paths/0` gives formatting a `:persistent_term` read — which also means the paths refresh on every build instead of freezing at first resolution, and `Engine.deps_paths/0` is deleted. Formatting takes no `Engine.Mix` lock and holds no lock across `formatter_for_file/2`; `StackMutation` is acquired once per push and once per pop, covering the `mix.exs` load and the push as a single critical section, and a project a build already pushed is borrowed rather than pushed twice.

Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/apps/engine/lib/engine.ex
+++ b/apps/engine/lib/engine.ex
@@ -109,32 +109,6 @@ defmodule Engine do
     :ok
   end
 
-  def deps_paths do
-    deps_paths(get_project())
-  end
-
-  defp deps_paths(%Project{kind: :mix}) do
-    case :persistent_term.get({__MODULE__, :deps_paths}, :error) do
-      :error ->
-        {:ok, deps_paths} =
-          Engine.Mix.in_project(fn _ ->
-            Engine.Mix.ensure_hex_and_rebar()
-            Mix.Task.run("loadpaths")
-            Mix.Project.deps_paths()
-          end)
-
-        :persistent_term.put({__MODULE__, :deps_paths}, deps_paths)
-        deps_paths
-
-      deps_paths ->
-        deps_paths
-    end
-  end
-
-  defp deps_paths(%Project{}) do
-    []
-  end
-
   def with_lock(lock_type, func) do
     :global.trans({lock_type, self()}, func, [Node.self()])
   end
--- a/apps/engine/lib/engine/code_mod/format/resolver.ex
+++ b/apps/engine/lib/engine/code_mod/format/resolver.ex
@@ -191,7 +191,6 @@ defmodule Engine.CodeMod.Format.Resolver do
 
   defp mix_formatter_from_task(%Project{} = project, file_path) do
     root_path = Project.root_path(project)
-    deps_paths = Engine.deps_paths()
 
     task_module =
       if Features.formatter_has_plugin_loader?() do
@@ -200,14 +199,20 @@ defmodule Engine.CodeMod.Format.Resolver do
         Mix.Tasks.Future.Format
       end
 
-    formatter_and_opts =
-      task_module.formatter_for_file(file_path,
-        root: root_path,
-        deps_paths: deps_paths,
-        plugin_loader: fn plugins -> Enum.filter(plugins, &Code.ensure_loaded?/1) end
-      )
-
-    {:ok, formatter_and_opts}
+    # Resolving :import_deps evaluates each dependency's .formatter.exs, and some
+    # of those read Mix.Project.config/0, so the project has to be on the Mix
+    # stack, the same as it is under mix format.
+    with {:error, reason} <-
+           Engine.Mix.with_project_config(project, fn _ ->
+             task_module.formatter_for_file(file_path,
+               root: root_path,
+               deps_paths: Engine.Mix.deps_paths(),
+               plugin_loader: fn plugins -> Enum.filter(plugins, &Code.ensure_loaded?/1) end
+             )
+           end) do
+      Logger.error("Cannot find formatter for #{file_path}: #{inspect(reason)}")
+      :error
+    end
   rescue
     ex ->
       Logger.error("Cannot find formatter due to: #{inspect(ex)}")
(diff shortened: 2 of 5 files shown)


apps/engine/lib/engine/code_mod/format/resolver.ex
```diff
@@ -200,14 +200,23 @@ defmodule Engine.CodeMod.Format.Resolver do
...
-        root: root_path,
-        deps_paths: deps_paths,
-        plugin_loader: fn plugins -> Enum.filter(plugins, &Code.ensure_loaded?/1) end
-      )
-
-    {:ok, formatter_and_opts}
+    # Resolve with the project pushed: `:import_deps` evaluates each dependency's
+    # `.formatter.exs`, some of which read `Mix.Project.config/0`. Without the
+    # project pushed that config is empty and resolution raises.
+    case Engine.Mix.in_project(project, fn _ ->
```
doorgan: This module was created in #804 explicitly to avoid locks during formatting, as it causes serious performance problems on large projects. A solution to this problem must try to avoid it.

apps/engine/lib/engine.ex
```diff
@@ -135,6 +135,27 @@ defmodule Engine do
     []
   end
 
+  # deps_paths is only used to resolve formatter `:import_deps`, which needs
+  # dependency paths, not compiled deps. `loadpaths` checks that deps are
+  # compiled and raises otherwise, so run it best-effort and fall back to the
```
doorgan: What version of elixir are you reproducing the issue on?
Looking at the mix source it seems no raise happens, but we get a default configuration back, it seems this is where the lack of locals_without_parens comes from

apps/engine/lib/engine.ex
```diff
@@ -135,6 +135,27 @@ defmodule Engine do
...
+  # paths alone when it fails.
+  defp load_deps_paths do
+    case Engine.Mix.in_project(fn _ ->
```
doorgan: Again, we need to avoid locks for the formatter

apps/engine/test/engine/code_mod/format_test.exs
```diff
@@ -108,6 +108,37 @@ defmodule Engine.CodeMod.FormatTest do
...
+
+    # locals_without_parens is only produced when the project is on the Mix
+    # stack, mirroring dep .formatter.exs files that read Mix.Project.config/0.
+    File.write!(Path.join(root, ".formatter.exs"), """
+    locals = if Mix.Project.config()[:app], do: [my_dsl: 1], else: []
+
+    [inputs: ["lib/**/*.{ex,exs}"], locals_without_parens: locals]
+    """)
+
+    File.write!(Path.join(lib_dir, "format.ex"), "my_dsl :foo\n")
```
doorgan, line 155: There's no `import_deps` or dependencies whatsoever here or in the new tests, so I'm not sure it's testing what the PR says

It does seem to exercise a different failure scenario

doorgan (changes requested):
Hi, thanks for the PR!

It does seem to fix a few things so that's good, but it should not be using the Mix locks. There's more context in #804 and maybe we can include this in the Resolver moduledoc, but the tl;dr is that Mix locks makes formatting take forever on large files, as it makes the formatter and the compiler share a lock and block each other.

Also please remove the co-authored by Claude from the commits, I appreciate the disclosure and it's the right thing to do, but *you* are the commit author not the tool you used. We can keep the disclosure in the PR description.

Thank you!

apps/engine/lib/engine.ex
kevinschweikert: Elixir 1.20.2, OTP 29

kevinschweikert:
@doorgan thanks for the timely response and review. I'll have to be honest. I just wanted to fix the issue that `error_tracker` uses the mix project in it's formatter config and this broke the formatting in our project. I had some Claude usage left today and thought i provide a fix and not just open an issue. I'll push something tomorrow but disclaimer that i have not a deep understanding of the code base. If that is a problem, so that you  don't have to review the work of an agent, i'm fine with closing this PR and opening an issue for someone else to implement a better solution. Anyways, thank you so much for your work on Expert!

doorgan:
@kevinschweikert no worries! The issues is not the AI usage btw, this PR already shows two issues that I'm able to reproduce thanks to you, If we can fix it without using locks, and we have tests exercising the failing paths, then this PR is good to go

apps/engine/test/engine/code_mod/format_test.exs
kevinschweikert: i've rewritten the rest, to exercise the exact issue we had with `error_tracker`

kevinschweikert:
@doorgan i got it down to lock only for the `StackMutation`. I tested the build locally and it fixes our issues and i can't find any regressions

doorgan:
@kevinschweikert I've been testing this and found a race condition where the compiler could clear the project stack and mess up with the solution here. I pushed a commit that fixes it and gets rids of locks altogether, let me know if it's working for you too!

kevinschweikert:
> @kevinschweikert I've been testing this and found a race condition where the compiler could clear the project stack and mess up with the solution here. I pushed a commit that fixes it and gets rids of locks altogether, let me know if it's working for you too!

@doorgan the build works for me! no more issues with the parenthesis and everything else still works. Thank you!

doorgan (approved):
Thank you! 💜

PR #837, opened 2026-08-06
chore(deps): bump burrito to 1.6 and zig to 0.16.0

AlternateRT:
This should allow anyone using Xcode 26.4 or higher to build a burrito release.


warmwaffles:
This is also effecting AUR builds. https://aur.archlinux.org/packages/expert-git

AlternateRT:
@warmwaffles Silly question, but why not build a plain build (using `just release` or `just install --prefix="/dir"`) instead? That would eliminate the Zig dependency.

warmwaffles:
That is not a silly question. I tried that route before but ran into issues packaging it for the AUR. On Jan 6th, 2026 I had to revert it back to using burrito. I can try not using burrito again, perhaps the issue is no longer present.

warmwaffles:
Can confirm now, I was able to drop `burrito` on the AUR side. Now the timing of it all, is that its down for maintenance so I can't push the fix up yet.

doorgan (approved):

PR #852, opened 2026-08-18
fix(expert): force unicode encoding on IO

katafrakt:
After coming back from vacations and updating Expert, it greeted me with a persistent error crashing it on start:

```
11:43:38.078 instance_id=1A0144148B1 [error] GenServer XPExpert.Buffer terminating
** (XPJason.DecodeError) unexpected end of input at position 136839
    (xp_jason 1.4.5) lib/jason.ex:92: XPJason.decode!/2
    (xp_gen_lsp 0.11.3) lib/gen_lsp/buffer.ex:87: anonymous fn/3 in XPGenLSP.Buffer.handle_cast/2
    (xp_telemetry 1.4.2) src/telemetry.erl:359: :xp_telemetry.span/3
    (xp_gen_lsp 0.11.3) lib/gen_lsp/buffer.ex:85: XPGenLSP.Buffer.handle_cast/2
    (stdlib 7.3.0.1) gen_server.erl:2460: :gen_server.try_handle_cast/3
    (stdlib 7.3.0.1) gen_server.erl:2418: :gen_server.handle_msg/3
    (stdlib 7.3.0.1) proc_lib.erl:333: :proc_lib.init_p_do_apply/3
Last message: {:"$gen_cast", {:incoming, "{\"jsonrpc\":\"2.0\",\"method\":\"textDocument/didOpen\",\"params\":{\"textDocument\":{\"uri\":\"file:///Users/pawel.sw/jump/Jump/api/lib/jump_web/live/meetings_live.ex\",\"languageId\":\"elixir\",\"version\":1,\"text\":\"[truncated]]" <> ...}}
State: %{awaiting_response: %{1 => {#PID<0.197.0>, #Reference<0.3052531354.1266941954.132065>}}, comm_data: %{}, comm: XPExpert.Stdio.Adapter, lsp: #PID<0.197.0>}
```

This was probably because of some unicode after the changes made in #824. Not without some difficulties I forced Claude Code and Codex to get me a reproduction test which failed with:

```
1) test preserves multibyte characters split across stdin reads (Expert.Stdio.UserTest)
     test/expert/stdio/user_test.exs:97
     Assertion with == failed
     code:  assert byte_size(round_tripped) == byte_size(payload)
     left:  10001
     right: 5001
     stacktrace:
       test/expert/stdio/user_test.exs:103: (test)
```

Then it's a matter of changing default encoding in the VM args of the release. After these changes Expert starts correctly.

### DIFF
--- a/apps/expert/rel/vm.args.eex
+++ b/apps/expert/rel/vm.args.eex
@@ -13,3 +13,6 @@
 
 # Alternative user_drv
 -user Elixir.XPExpert.Stdio.User
+
+# LSP messages are UTF-8 regardless of the host locale
+-kernel standard_io_encoding unicode
--- a/apps/expert/test/expert/stdio/user_test.exs
+++ b/apps/expert/test/expert/stdio/user_test.exs
@@ -93,13 +93,22 @@ defmodule Expert.Stdio.UserTest do
     refute merged =~ "STILL_RUNNING"
   end
 
+  @tag :tmp_dir
+  test "preserves multibyte characters split across stdin reads", %{tmp_dir: tmp_dir} do
+    # A 4-byte character offset by one ASCII byte straddles the driver's 1024-byte reads.
+    payload = "a" <> String.duplicate("😀", 1_250)
+
+    round_tripped = round_trip(payload, tmp_dir)
+
+    assert byte_size(round_tripped) == byte_size(payload)
+    assert round_tripped == payload
+  end
+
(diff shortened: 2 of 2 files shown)


doorgan (approved):
Good catch, I tested that PR in both macos and windows with nvim/vscode and didn't find this problem, I wonder what's different in your case

In any case the change looks good, thanks!

katafrakt:
The problem happened only in one project, so I guess there was some non-ASCII character somewhere, maybe in a filename even.

Issue #853, opened 2026-08-19
Go to definition to function defined by macro with name ending in `?` or `!` does not work


katafrakt:
This is a tracking issue for a bug found by a coworker. #759 added support for going to definition to functions defined by a macro. But turns out is only works for functions with "regular" names (for example `MyApp.Repo.one/1`) but not for the ones ending in `?` or `!` (`MyApp.Repo.exists?/1`)

I opened PR to ElixirSense already, but I think current HEAD of it does not pass tests in Expert, so even after the merge we'd have to figure something out (fork and cherry-pick for a time being, perhaps?).

https://github.com/elixir-lsp/elixir_sense/pull/344

doorgan:
We have the compiler tracer indexer on `main` which should be able to index macro defined functions, maybe this can be fixed on the expert side too?

katafrakt:
I'll check that

katafrakt:
#856 fixed it for functions defined by macros in the project codebase, but apparently not for dependencies (like `Repo.exists?/1`). Not sure if it should or if it's just something for wha we need ElixirSense fallback.

doorgan:
I'm getting completions from them. Maybe you need to do a cold reindex? If so we should bump the index version to force that.
<img width="1219" height="117" alt="Image" src="https://github.com/user-attachments/assets/6d12d4ef-1d71-4b1e-bdb5-2d180c8b3a4c" />

katafrakt:
Yes, I get completions too. I think they are indexed, but it's something with go to definition itself. So far I have spent few hours on it, but haven't been able to pinpoint the reason.

doorgan:
I get go to definition too, maybe it's some project specific bug?

https://github.com/user-attachments/assets/dad27d9a-b66b-43a8-986d-80d8d920e4cc

There it takes me to where the function is defined, which is the call site for the macro that injects that function. If we'd want that to take us to `Ecto.Repo` in the deps folder we'd need to add some Ecto specific handler or do it as part of #825

katafrakt:
@doorgan but when you go to definition on `AshHq.Repo.all`, it will actually go to `def all`, right? This is the gap I'm trying to close.

doorgan:
@katafrakt it does, and looking into it yeah it's the elixir sense fallback that does not recognize the ?/! functions, your fix to it should fix it for us
If we want to fix it within expert we'd need to do the same as elixir sense and do some extra analysis on generated functions, but the upstream patch should do

[...]

PR #856, opened 2026-08-21
fix: index macro-generated definitions from compiler traces

katafrakt:
This is an attempt to solve #853 on the Expert side, leveraging compiler traces. In my tests it fixes the issue described in the linked ticket, however it works for go-to-definition but not for references (probably needs adjustments for reference queries; but I think it's a material for a different PR).

### DIFF
--- a/apps/engine/lib/engine/search/indexer/beams.ex
+++ b/apps/engine/lib/engine/search/indexer/beams.ex
@@ -37,6 +37,23 @@ defmodule Engine.Search.Indexer.Beams do
     end
   end
 
+  def extract_definitions_from_binary(beam, source_path, module, compiler_definitions)
+      when is_binary(beam) and is_binary(source_path) and is_atom(module) and
+             is_list(compiler_definitions) do
+    compiler_entries =
+      remote_context_definition_entries(
+        Forge.Path.native(source_path),
+        module,
+        compiler_definitions
+      )
+
+    case extract_definitions_from_binary(beam, source_path) do
+      {:ok, entries} -> {:ok, entries ++ compiler_entries}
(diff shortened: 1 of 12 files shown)


apps/engine/lib/engine/search/indexer/beams.ex
```diff
@@ -368,29 +398,114 @@ defmodule Engine.Search.Indexer.Beams do
...
+      root_block: Block.root(),
+      source_path: source_path
+    }
+
+    compiler_definitions
+    |> Enum.flat_map(&foreign_context_definition_entry(&1, context))
+    |> Enum.uniq_by(&{&1.subject, &1.type})
   end
 
+  defp foreign_context_definition_entry(
```
doorgan: If I'm understanding this correctly, foreign here means "remote"?

doorgan (approved):
Good one, thanks! Agreed that adjustments to references can be made in a different PR if needed

apps/engine/lib/engine/search/indexer/beams.ex
katafrakt: It means that the definition is in a different module than the one currently indexed. "Foreign" is not a great name. Do you think "remote" is better? To me it slightly suggests network, but maybe not justly.

doorgan: Yeah, within erlang/elixir, a remote function is one defined in a different module, like `Foo.bar()` is a remote function call. Not the best name I agree but it's consistent with them and with the rest of the expert codebase

katafrakt: ok, I'll change to remote because it's better than foreign

PR #857, opened 2026-08-26
fix: load dev-only dependencies during analysis

### LINKED ISSUE #533: "Module X is not loaded and cannot be found" errors displayed inline
We have a lot of `Module such-and-such is not loaded and cannot be found` errors in the editor (Cursor, in my case) that marks lines with red squiggly underlines. They appear to be random and total luck of the draw.

Similarly, we get a lot of compiler warnings that are false negatives, resulting in yellow squiggly underlines. Sometimes they go away when we delete the `.expert` folder and sometimes they don't. The compiler doesn't give warnings when compiling the project in the terminal, so there's a discrepancy in how the LSP behaves.

Expert LSP version: 0.3.1
Platform: MacOS Tahoe 26.0

Editor:
```
Version: 2.5.26
VSCode Version: 1.105.1
Commit: 7d96c2a03bb088ad367615e9da1a3fe20fbbc6a0
Date: 2026-02-26T04:57:56.825Z
Build Type: Stable
Release Track: Default
Electron: 39.4.0
Chromium: 142.0.7444.265
Node.js: 22.22.0
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.0.0
```

mvanhorn:
Extend the existing initial dependency-preparation path in `Engine.Build.Project` so Expert prepares the union of dependencies active in the development and test environments inside Expert's isolated build workspace. Expert runs project analysis with `Mix.env(:test)` so it can compile test helpers and test-only code.

A temporary Mix project references a local path dependency with `only: :dev`; after Expert's initial build, project code and an `.exs` document can resolve that module without a missing-module diagnostic; The same project references a different local path dependency with `only: :test`; the normal test-environment compile still resolves it and test-helper analysis remains available.

Fixes #533

### DIFF
--- a/apps/engine/lib/engine/build/project.ex
+++ b/apps/engine/lib/engine/build/project.ex
@@ -129,26 +129,70 @@ defmodule Engine.Build.Project do
   end
 
   defp prepare_for_project_build(token) do
+    fetch_dependencies? = prepare_build_tools(token)
+    prepare_dev_dependencies(token, fetch_dependencies?)
+    prepare_dependencies(token, fetch_dependencies?: fetch_dependencies?)
+  end
+
+  defp prepare_build_tools(token) do
     if Internet.connected_to_internet?() do
       Progress.report(token, message: "mix local.hex")
       Mix.Task.run("local.hex", ~w(--force --if-missing))
 
       Progress.report(token, message: "mix local.rebar")
       Mix.Task.run("local.rebar", ~w(--force --if-missing))
-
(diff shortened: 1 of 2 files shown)


doorgan (commented):
Thanks for the PR!

I'm not sure this is a safe direction though. My main concern is that the two calls to `Mix.Task.run(:loadconfig)` merges the dev and test configurations, which I can see causing problems. There's likely other subtle issues stemming from this or by compiling code in both envs(different envs definining modules with the same name is one that comes to mind)

PR #866, opened 2026-08-28
perf: avoid forced warm-start compiles

### LINKED ISSUE #863: Do we still need to `mix compile --force` on startup?
Currently on project open we are doing `mix compile --force` every time, even if no change was made after the last session. On large projects it is quite heavy tax to pay, especially given that we don't compile concurrently and use compiler traces.

I wonder if this forced compilation is still necessary after all the changes with traces and reading BEAM files, but I'm not really sure why it was there in the first place (it seems to be present since at least 2023, so not a new thing). Anyone can share insight on why it was needed?

sophiathedev:
## Summary

- choose the startup compile mode from the persisted search index status
- separate first-compile dependency preparation from the Mix `--force` option
- preserve forced compilation after an engine-node crash deletes build artifacts
- cover cold, warm, forwarded-mode, and repeated-build behavior

## Testing

- `mix test --warnings-as-errors` (engine: 1,162 passed; expert: 844 passed)
- `mix format --check-formatted`
- `mix credo`
- `mix dialyzer --force-check`

Closes #863

### DIFF
--- a/apps/engine/lib/engine/build/project.ex
+++ b/apps/engine/lib/engine/build/project.ex
@@ -9,7 +9,7 @@ defmodule Engine.Build.Project do
 
   require Logger
 
-  def compile(%Project{kind: :mix} = project, initial?) do
+  def compile(%Project{kind: :mix} = project, initial?, force?) do
     Engine.Mix.in_project(fn _ ->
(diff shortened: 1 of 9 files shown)


doorgan (approved):
Great work, thank you!

PR #874, opened 2026-09-02
fix: record module binary if module metata read fails

scohen:
It's possible that the module finishes compiling in the time between the time it's reported in the compile tracer and when the trace buffer handle it. If this happens, a function clause error is raised, which will crash the server.

This fix catches the function clause error, and uses the module's binary to get the module's information.

### DIFF
--- a/apps/engine/lib/engine/compilation/trace_buffer.ex
+++ b/apps/engine/lib/engine/compilation/trace_buffer.ex
@@ -66,6 +66,10 @@ defmodule Engine.Compilation.TraceBuffer do
       )
 
     {:noreply, state}
+  rescue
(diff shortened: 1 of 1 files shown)


apps/engine/lib/engine/compilation/trace_buffer.ex
doorgan, line 56: I didn't catch this before, but shouldn't we be reading the `binary` here instead of using `Module`? The debug chunks in it should contain the definitions and would not be tied to whether the module is compiled or not

doorgan (approved):
The fallback works. I'll address my above comment in another PR, it's not introduced by this one.

PR #878, opened 2026-09-02
fix: clear stale diagnostics as you type

doorgan:
Fixes diagnostics not clearing until you save the file

Say you have this:

```elixir
def foo(bar) do
  {:ok, bar}
end
```
Then change `bar` to `baz` without saving, you will get an `unused variable baz` warning. Fixing it will remove the diagnostic.
If you change it to `baz`, get the diagnostic, save the document, then fix it back to `bar`, the `unused variable` diagnostic will not go away until you save the file.

This is caused by project diagnostics not being cleared when you type. This change makes such that dirty files clear the project diagnostics.

This also means that if project compilation produces a diagnostic that file compilation cannot produce, a dirty document will not show them. I have not yet found a way to reliably display both, keeping only the valid diagnostics, as we need to compile the project anyways to get the right diagnostics anyways.

### DIFF
--- a/apps/expert/lib/expert/project/diagnostics.ex
+++ b/apps/expert/lib/expert/project/diagnostics.ex
@@ -64,6 +64,8 @@ defmodule Expert.Project.Diagnostics do
         file_diagnostics(uri: uri, build_number: build_number, diagnostics: diagnostics),
         %State{} = state
       ) do
+    state = State.clear_stale_project_diagnostics(state, uri)
+
     state =
       case diagnostics do
         [] ->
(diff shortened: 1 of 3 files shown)


katafrakt (approved):
Works as advertised 👍 

> This also means that if project compilation produces a diagnostic that file compilation cannot produce, a dirty document will not show them. I have not yet found a way to reliably display both, keeping only the valid diagnostics, as we need to compile the project anyways to get the right diagnostics anyways.

Perhaps we could have the distinction ins `State` for file-level diagnostics and project-level diagnostics, and only clean the former on typing and the latter on recompile. But maybe current limitation is ok.

PR #881, opened 2026-09-04
feat(expert): fold anonymous functions

sophiathedev:
## Summary

- Add folding ranges for multi-line anonymous functions.
- Use parser closing metadata so the final `end` line remains visible.
- Cover simple, nested, multi-clause, and single-line anonymous functions.

## Validation

- `mix test test/expert/provider/handlers/code_folding_test.exs --warnings-as-errors` (25 passed)
- `mix test --warnings-as-errors` (849 passed, 12 excluded)
- `mix format --check-formatted`
- `mix credo --strict`
- `mix dialyzer`

Refs #146


katafrakt (approved):
this works well, thank you!

PR #883, opened 2026-09-05
perf(expert): fold code in one AST pass

sophiathedev:
code folding walks the same AST twice, once for blocks and once for strings. this collects both in one pass, keeping separate accumulators so the ranges and their order don't change. includes a regression test with nested blocks, strings and comments, plus a benchmark script.

### benchmark

reran locally with Benchee 1.5.1 on an Apple M3 Pro, Elixir 1.20.3 / OTP 29.0.5. before is `2afc03ce`, after is `945d6b6e`; the Benchee conversion used for this run is now committed in `8ce34be3` (handler unchanged).

2s warmup and 5s measurement per scenario, parallelism 1. timings include parsing and folding, but exclude document creation and LSP transport. both implementations receive the same prebuilt inputs, and exact ranges/order are checked before timing.

| input | before median | after median |
| --- | ---: | ---: |
| small handler | 0.849 ms | 0.570 ms |
| SQLite backend | 5.49 ms | 6.31 ms |
| Enum fixture | 20.10 ms | 21.81 ms |

this run had substantial background load (about 50% CPU busy before starting), with relative standard deviations around 26–79%. results are mixed and too noisy to claim a reliable speedup; a quiet-machine rerun is still needed. these replace the earlier custom-harness numbers rather than mixing the two measurement methods.

to rerun, from `apps/expert`:

```sh
mix compile --warnings-as-errors
elixir -pa '_build/dev/lib/*/ebin' benchmarks/code_folding.exs 2afc03ce
```

### validation

rechecked the handler, regression test and benchmark diff. the latest local run passed:

- `ERL_FLAGS='-start_epmd false -epmd_module Elixir.Forge.EPMD' mix test test/expert/provider/handlers/code_folding_test.exs --warnings-as-errors` — 26 tests
- the Benchee run above, including output/order parity on all three inputs
- `git diff --check`

the earlier full Expert suite passed with 851 tests, 12 excluded; it was not rerun for this benchmark-only change.

### DIFF
--- a/apps/expert/benchmarks/code_folding.exs
+++ b/apps/expert/benchmarks/code_folding.exs
@@ -1,0 +1,79 @@
+# run from apps/expert: mix run --no-start benchmarks/code_folding.exs [baseline_ref]
+# includes parsing and folding, excludes document creation and transport.
+# use an idle machine for wall-time comparisons.
+
+alias Expert.Document.Context
+alias Expert.Provider.Handlers.CodeFolding
+alias Forge.Document
+alias GenLSP.Requests.TextDocumentFoldingRange
+alias GenLSP.Structures.FoldingRangeParams
+alias GenLSP.Structures.TextDocumentIdentifier
+
+root = Path.expand("../../..", __DIR__)
+baseline = List.first(System.argv()) || "main"
+handler_path = "apps/expert/lib/expert/provider/handlers/code_folding.ex"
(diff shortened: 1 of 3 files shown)


katafrakt (approved):
Looks like a good change to me. Not sure if we need the benchmark file committed to the repo though - and if we do, shouldn't we use Benchee?

PR #884, opened 2026-09-05
fix: don't reverse text document params

doorgan:
I have not been able to reproduce this myself, but every now and then someone reports a weird diagnostic where a part of it is reversed, like this screenshot:

Screenshot 2026-09-05 at 5 27 28 PM

Notice how it says `fed/2` is undefined, which would mean `def/2` instead, and that function does exist and is in scope. This is usually sporadic and users report that "it fixes by itself".

With a synthetic reproduction test I was able to verify that we are indeed sometimes reversing the `textDocument/didChange` params if it contains multiple sequential edits, due to a missing protocol implementation for `Forge.Protocol.Convertible` for that structure, which made it default to the List implementation and be put through an `Enum.reduce_while` that reversed it.

My guess is that if you type `def` and make a short pause before continuing typing, the client will send `d`, `e` and `f` as multiple sequential changes instead of a single `def` addition, that sequence gets accidentally reversed, and we get those spurious diagnostics with mangled names, so this is what this PR reproduces and fixes.

### DIFF
--- a/apps/expert/lib/convertibles/gen_lsp.structures.did_change_text_document_params.ex
+++ b/apps/expert/lib/convertibles/gen_lsp.structures.did_change_text_document_params.ex
@@ -1,0 +1,11 @@
+defimpl Forge.Protocol.Convertible, for: GenLSP.Structures.DidChangeTextDocumentParams do
+  alias GenLSP.Structures.DidChangeTextDocumentParams
+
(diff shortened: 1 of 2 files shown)


katafrakt (approved):
I have never encountered this, but the change makes sense

PR #889, opened 2026-09-08
fix(engine): keep ModuleStore alive across builds

benwilson512:
We have been attempting to improve our project compilation performance, and in particular how that interacts with Expert. One weird issue we ran into when testing rapid "save" calls on a file is that this could result in a crash in Expert.

# Cause / Resolution

Engine.ModuleStore exited :normal after every project_compiled event. As a permanent child of Engine.Supervisor that is one restart per build, so the fourth build to finish within five seconds exceeded the default restart intensity and shut the engine application down; the server then stopped reading stdin once its main process was restarted. Four quick saves that produce no-op builds are enough to trigger it since the edit window dropped to 100 ms (#872). Stay alive and keep rebuilding the store after each compile.

Full disclosure: The test was written with the assistance of Fable, I'm pretty sure the stubs make sense, but if it doesn't seem right feel free to just tell me to go away and do it properly :)

### DIFF
--- a/apps/engine/lib/engine/module_store.ex
+++ b/apps/engine/lib/engine/module_store.ex
@@ -24,6 +24,10 @@ defmodule Engine.ModuleStore do
       {:done, token}
     end)
 
-    {:stop, :normal, state}
+    # This is a permanent child of Engine.Supervisor. Exiting here, even with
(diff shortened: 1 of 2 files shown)


doorgan (approved):
Good catch, thank you for the fix!

PR #896, opened 2026-09-13
fix: update nixpkgs, switch to beamPackages.overrideScope

rkallos:
Hello, and thank you for Expert!

I ran across the following error when trying to update my nix-darwin configuration (which imports Expert as a flake):

```
error: 'beamPackages.extend' has been replaced by 'beamPackages.overrideScope'
```

After a bit of digging, I learned this helpful message was added in [this nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/556282), following a [first PR](https://github.com/NixOS/nixpkgs/pull/554005) that changes beamPackages in ways I don't quite understand aside from `.extend` not existing anymore.

This PR (1) updates the nixpkgs flake to a more recent date, and (2) replaces `.extend` in flake.nix with `.overrideScope`.

I confirmed that Expert builds locally via the flake, and using my clone of expert as a flake for my nix-darwin configuration succeeds.


rkallos:
Trying to fix the failure in CI expanded the scope of this branch, and I'm now playing with the idea of removing nix-beam-flakes. The same `.extend` deprecation affects that flake as well, making it unusable with a more recent nixpkgs. Further, it seems that nix-beam-flakes was mainly used for the flake's default devShell. With the installation of language servers disabled, it seems the only thing nix-beam-flakes was providing was erlang and elixir (of course), and enabling shell history.

mhanberg:
Don't remove the beam flakes. I have updated that project to fix the issue so you can just bump that input.

PR #897, opened 2026-09-15
feat(expert): fold multiline bitstrings

sophiathedev:
adds folding for multiline `<< ... >>` literals and patterns, covering the bitstring item in #146.

reuses the existing line/closing metadata collector for anonymous functions, with no extra AST traversal. the closing `>>` line stays visible, matching existing block folding. single-line, empty and unclosed bitstrings are skipped; sigil contents don't produce extra bitstring folds.

includes six regression tests for literals, nested/typed segments, patterns, skipped inputs and sigil contents. three tests failed before the fix; all 31 folding tests pass after it.

validation on macOS, Elixir 1.20.3 / OTP 29.0.5, from `apps/expert`:

- `ERL_FLAGS='-start_epmd false -epmd_module Elixir.Forge.EPMD' mix test --warnings-as-errors` — 857 passed, 12 excluded
- `mix credo --strict`
- `mix dialyzer`
- `mix format --check-formatted lib/expert/provider/handlers/code_folding.ex test/expert/provider/handlers/code_folding_test.exs`
- `git diff --check`

based directly on main, independent of #883. both touch the same folding handler, so a small rebase may be needed depending on merge order. other unchecked items in #146 are out of scope.


doorgan (approved):
Thank you!

PR #898, opened 2026-09-15
fix: skip expanding non-grouped aliases

sophiathedev:
addresses the already-expanded alias bug reported in #265. selecting `alias Foo.Bar` could offer "Expand aliases" and rewrite it to `alias Foo.Foo.Bar`; an alias with `as:` could also raise during execution.

skip non-grouped alias declarations and their children in the shared eligibility check, so both listing and executing the refactor reject them. grouped and nested aliases still expand normally.

adds regression coverage for whole declarations and selected module names, including `as:`, `__MODULE__` and Erlang aliases, plus a grouped alias following a plain alias. the new regression test failed before the fix; all 11 ExpandAliases tests pass after it. the remaining audit items in #265 are out of scope.

validation on macOS, Elixir 1.20.3 / OTP 29.0.5, from `apps/forge`:

- `ERL_FLAGS='-start_epmd false -epmd_module Elixir.Forge.EPMD' mix test --warnings-as-errors` — 3,028 passed (3,013 tests, 9 doctests, 6 properties)
- `mix credo --strict`
- `mix format --check-formatted lib/forge/refactor/alias/expand_aliases.ex test/forge/refactor/alias/expand_aliases_test.exs`
- `git diff --check`


doorgan (approved):
Thank you!

